PR #15, opened 2017-09-29
Added Copy + Clone traits to values & types

### LINKED ISSUE #14: Types & Values should be Copy + Clone
Rust does this for [their Value types](https://github.com/rust-lang/rust/blob/1b38776c1f68c6fd47c1b2f7b7974efc7dd64901/src/librustc_trans/value.rs#L16) and LLVM suggests Types [are singletons](https://llvm.org/doxygen/classllvm_1_1Type.html#details). Although Values don't seem to be singletons, this should still be safe to do.

6A:
Also added tests related to them. Fixes #14.

### DIFF
--- a/src/types/array_type.rs
+++ b/src/types/array_type.rs
@@ -8,7 +8,7 @@ use types::traits::AsTypeRef;
 use types::{Type, BasicType, PointerType, FunctionType};
 use values::{BasicValue, ArrayValue, PointerValue, IntValue};
 
-#[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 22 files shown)


6A:
Alright, I fixed it. Thanks for the heads-up.

6A:
Just realized that the macro-generated enums aren't cloneable / copyable. I'll take care of this tomorrow, as it makes things quite hard when dealing with any type.

TheDan64:
Probably worth adding to the super traits as well

Issue #2, opened 2017-06-29
Examples in Documention


TheDan64:
Probably https://docs.rs. Will add a checklist here to keep track of progress.

6A:
I'm currently working on the Kaleidoscope tutorial using Inkwell, and I think it would make a great self-documenting example. Any chance this would get added to the repo, once it fully works?

TheDan64:
Definitely! I was actually thinking the same thing. I just cant decided if it should be a test or if it should live as its own binary in an /examples directory. The former has the advantage of being run in every build, so we can easily ensure it compiles every time. But the latter has the advantage of being more easily noticeable when someone if browsing the repo or root directory.

For reference, the example in the readme is also duplicated as a test but I think it's small enough that it isn't a big deal. A Kaleidoscope example would be much larger, for sure. I will add Kaleidoscope to the checklist in the original post.

6A:
I think living in the /examples directory is definitely a plus. Furthermore, `cargo test` automatically compiles examples without running them (or at least can be configured to do so), which removes this problem.

I'm actually almost done with the first three chapters of the tutorial: the basic lexer, parser, and compiler are more-or-less working.

TheDan64:
Oh neat, I didn't know that. Let's do that then. You can put it in /examples/kaleidoscope

Michael-F-Bryan:
Are there any plans on getting Travis to upload docs for `master` to GitHub Pages after every commit? It'd probably make hacking on the repository or importing `inkwell` directly from `git` a lot easier.

As an example, [this] is what we've got set up for `mdbook`.

[this]: https://github.com/rust-lang-nursery/mdBook/blob/b2ad669c619849fb664e5015486382c91e3048f5/.travis.yml#L64-L65

TheDan64:
No, I've never even thought about doing this as I am not really familiar with how github pages work.

That said, it does sound like an excellent idea, so thanks! I'll add it to the above checklist and try and prioritize it.

So if I understand correctly, you're basically committing the docs from `cargo doc` to the `gh-pages` branch? And GH pages will automatically host that?

Could you also clarify how it would make importing inkwell from git easier?

Michael-F-Bryan:
> So if I understand correctly, you're basically committing the docs from cargo doc to the gh-pages branch? And GH pages will automatically host that?

[...]

PR #19, opened 2017-10-02
Added Kaleidoscope example

6A:
Added the [Kaleidoscope](https://llvm.org/docs/tutorial) example to the project. Additionally, `write_bitcode_to_file` now compiles on Windows (although it will always fail and return `false`).

*Note: Yes, the first three commits have the wrong author, but after an hour of trying to fix this, I just give up on the fame.*

### DIFF
--- a/examples/kaleidoscope.rs
+++ b/examples/kaleidoscope.rs
@@ -1,16 +1,29 @@
+#![feature(box_syntax)]
+//! This is an example of the [Kaleidoscope tutorial](https://llvm.org/docs/tutorial/)
+//! made in Rust, using Inkwell.
+//! Currently, all features up to the [4th chapter](https://llvm.org/docs/tutorial/LangImpl04.html)
+//! are available.
+
 extern crate inkwell;
 
 use std::collections::HashMap;
-use std::io;
+use std::io::{self, Write};
 use std::iter::Peekable;
 use std::str::Chars;
 use std::ops::DerefMut;
 
+use inkwell::FloatPredicate;
 use inkwell::builder::Builder;
-use inkwell::types::FloatType;
-use inkwell::values::{BasicValue, FloatValue, FunctionValue};
+use inkwell::context::Context;
+use inkwell::module::Module;
+use inkwell::passes::PassManager;
+use inkwell::targets::{InitializationConfig, Target};
+use inkwell::types::BasicType;
+use inkwell::values::{BasicValue, FloatValue, FunctionValue, PhiValue};
 
 use Token::*;
+
+const ANONYMOUS_FUNCTION_NAME: &str = "anonymous";
 
 
 // ======================================================================================
@@ -18,15 +31,18 @@ use Token::*;
 // ======================================================================================
 
 /// Represents a primitive syntax token.
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub enum Token {
     EOF,
     Comment,
     LParen,
     RParen,
     Comma,
-    Def,
-    Extern,
+    Def, Extern,
+    If, Then, Else,
+    For, In,
+    Unary, Binary,
+
     Ident(String),
     Number(f64),
     Op(char)
@@ -127,13 +143,10 @@ impl<'a> Lexer<'a> {
             ch @ '.' | ch @ '0' ... '9' => {
                 // Parse number literal
                 loop {
-                    let ch = *chars.peek();
-
-                    if ch.is_none() {
-                        return Ok(Token::EOF);
-                    }
-
-                    let ch = ch.unwrap();
+                    let ch = match chars.peek() {
+                        Some(ch) => *ch,
+                        None => { return Ok(Token::EOF); }
+                    };
 
                     // Parse float.
                     if ch != '.' && !ch.is_digit(16) {
@@ -150,13 +163,10 @@ impl<'a> Lexer<'a> {
             'a' ... 'z' | 'A' ... 'Z' | '_' => {
                 // Parse identifier                
                 loop {
-                    let ch = chars.peek();
-
-                    if ch.is_none() {
-                        break;
-                    }
-
-                    let ch = *ch.unwrap();
+                    let ch = match chars.peek() {
+                        Some(ch) => *ch,
+                        None => { return Ok(Token::EOF); }
+                    };
 
                     // A word-like identifier only contains underscores and alphanumeric characters.
                     if ch != '_' && !ch.is_alphanumeric() {
@@ -170,6 +180,14 @@ impl<'a> Lexer<'a> {
                 match &src[start..pos] {
                     "def" => Ok(Token::Def),
                     "extern" => Ok(Token::Extern),
+                    "if" => Ok(Token::If),
+                    "then" => Ok(Token::Then),
+                    "else" => Ok(Token::Else),
+                    "for" => Ok(Token::For),
+                    "in" => Ok(Token::In),
+                    "unary" => Ok(Token::Unary),
+                    "binary" => Ok(Token::Binary),
+
                     ident => Ok(Token::Ident(ident.to_string()))
                 }
             },
@@ -210,35 +228,61 @@ impl<'a> Iterator for Lexer<'a> {
 pub enum Expr {
     Number(f64),
     Variable(String),
-    Binary(char, Box<Expr>, Box<Expr>),
-    Call(String, Vec<Expr>)
+
+    Binary { 
+        op: char,
+        left: Box<Expr>,
+        right: Box<Expr>
+    },
+
+    Call {
+        fn_name: String,
+        args: Vec<Expr>
+    },
+
+    Conditional {
+        cond: Box<Expr>,
+        consequence: Box<Expr>,
+        alternative: Box<Expr>
+    },
+
+    For {
+        var_name: String,
+        start: Box<Expr>,
+        end: Box<Expr>,
+        step: Option<Box<Expr>>,
+        body: Box<Expr>
+    }
 }
 
 /// Defines the prototype (name and parameters) of a function.
 #[derive(Debug)]
 pub struct Prototype {
     pub name: String,
-    pub args: Vec<String>
+    pub args: Vec<String>,
+    pub is_op: bool,
+    pub prec: usize
 }
 
 /// Defines a user-defined function.
 #[derive(Debug)]
 pub struct Function {
     pub prototype: Prototype,
-    pub body: Expr
+    pub body: Expr,
+    pub is_anon: bool
 }
 
 /// Represents the `Expr` parser.
-pub struct Parser {
+pub struct Parser<'a> {
     tokens: Vec<Token>,
     pos: usize,
-    prec: HashMap<&'static str, i32>
+    prec: &'a mut HashMap<char, i32>
 }
 
-impl Parser {
+impl<'a> Parser<'a> {
     /// Creates a new parser, given an input `str` and a `HashMap` binding
     /// an operator and its precedence in binary expressions.
-    pub fn new(input: String, op_precedence: HashMap<&'static str, i32>) -> Parser {
+    pub fn new(input: String, op_precedence: &'a mut HashMap<char, i32>) -> Self {
         let mut lexer = Lexer::new(input.as_str());
         let tokens = lexer.by_ref().collect();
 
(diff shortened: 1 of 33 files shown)


6A:
Oh, by the way, this should lead to the update of #2.

6A:
Want me to add tests to Kaleidoscope? In any case, I'll add some improvements, and after this I'm planning to work on the documentation. I also have some remarks on some structures and enums

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1290 @@
...
+                    &Some(ref step) => self.compile_expr(&step)?,
+                    &None => self.context.f64_type().const_float(1.0)
+                };
+
+                let end_cond = self.compile_expr(&end)?;
+
+                let curr_var = self.builder.build_load(&start_alloca, var_name);
+                let next_var = self.builder.build_float_add(&curr_var.as_float_value(), &step, "nextvar");
+
+                self.builder.build_store(&unsafe { std::mem::transmute(next_var) }, &start_alloca);
```
TheDan64: Why is a float being used as a pointer? Did you mean `build_store(&start_alloca, &next_var)`?

6A:
Yeah, in theory `transmute` should never be used, but I let it up the time to fix some issues I found with Inkwell. Furthermore everything is in a single file, which is, to me, a plus. It is a little less overwhelming to see a single file with 1300 LoC than 5 smaller files. But this is your project, and thus your choice. I'll reply to your other comments as soon as I have the time.

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1290 @@
...
+                        &Expr::Variable(ref var_name) => var_name,
+                        _ => {
+                            return Err("Expected variable as left-hand operator of assignement.");
+                        }
+                    };
+
+                    let var_val = self.compile_expr(&right)?;
+                    let var = self.variables.get(var_name.as_str()).ok_or("")?;
+
+                    Ok(unsafe { std::mem::transmute(self.builder.build_store(&var, &var_val)) })
```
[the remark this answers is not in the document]
6A: The method returns a `FloatValue`, but `build_store` returns a `PointerValue`. I'm casting the latter to the former here, since the structure is actually a pointer to a float.

examples/kaleidoscope.rs
6A: Good catch, thanks!

6A:
My last commit should address all your concerns. The only `transmute` call left is necessary in order to invoke the compiled function.

TheDan64:
I got to try it out and I like it a lot!

Some thoughts:
* \> arrow has a space to the left of it that seems out of place
* Hitting enter at the prompt panics with index out of bounds, maybe this should just be a no-op that prints a newline and the prompt again?
* LLVM IR seems to always output, and 2x when the cli arg is passed in
* Even if you don't want to split up the file, it should still go in its own directory, along with a README.md describing the basic syntax (stuff like variable declarations, functions, etc)

src/execution_engine.rs
```diff
@@ -50,6 +50,12 @@ impl ExecutionEngine {
         }
     }
 
+    pub fn add_global_mapping(&mut self, global: &AnyValue, addr: *const ()) {
```
TheDan64: I think it might make sense to remove this for now since it's not used and is likely to change anyway. Long term it may end up looking something like this:
```rust
struct MemoryAddress(usize);

pub fn add_global_mapping(&(mut?) self, global: &GlobalValue, addr: &MemoryAddress) {}
```

For example, the execution engine outputs an address which can easily be moved or copied out of the EE's scope, so I think that will end up returning a non-Copy `&MemoryAddress` so that the underlying value is kept in scope but is still transmutable to a function.

What do you think?

src/module.rs
```diff
@@ -264,14 +263,24 @@ impl Module {
...
         // REVIEW: as_raw_fd docs suggest it only works in *nix
         // Also, should_close should maybe be hardcoded to true?
         unsafe {
             LLVMWriteBitcodeToFD(self.module, file.as_raw_fd(), should_close as i32, unbuffered as i32) == 0
         }
     }
 
+    #[cfg(windows)]
+    #[allow(unused_variables)]
+    pub fn write_bitcode_to_file(&self, file: &File, should_close: bool, unbuffered: bool) -> bool {
```
TheDan64: Sweet! Thanks! I hope we'll be able to get this working on windows eventually, and if not we'll probably just remove the method completely from windows.

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+
+                Ok(Token::Comment)
+            },
+
+            ch @ '.' | ch @ '0' ... '9' => {
+                // Parse number literal
+                loop {
+                    let ch = match chars.peek() {
+                        Some(ch) => *ch,
+                        None => { return Ok(Token::EOF); }
```
TheDan64: Nit, I believe you can just do `None => return Ok(Token::EOF),`

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+                
+                Ok(Token::Number(src[start..pos].parse().unwrap()))
+            },
+
+            'a' ... 'z' | 'A' ... 'Z' | '_' => {
+                // Parse identifier                
+                loop {
+                    let ch = match chars.peek() {
+                        Some(ch) => *ch,
+                        None => { return Ok(Token::EOF); }
```
TheDan64: No need for parenthesis or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+
+                (id, false, 0)
+            },
+
+            Binary => {
+                self.advance()?;
+
+                let op = match self.curr() {
+                    Op(ch) => ch,
+                    _ => { return Err("Expected operator in custom operator declaration."); }
```
TheDan64: No need for parenthesis or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+
+                (name, true, prec)
+            },
+
+            Unary => {
+                self.advance()?;
+
+                let op = match self.curr() {
+                    Op(ch) => ch,
+                    _ => { return Err("Expected operator in custom operator declaration."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+                let mut name = String::from("unary");
+
+                name.push(op);
+
+                self.advance()?;
+
+                (name, true, 0)
+            },
+
+            _ => { return Err("Expected identifier in prototype declaration.") }
```
TheDan64: No need for parenthesis or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+        };
+
+        match self.curr() {
+            LParen => (),
+            _ => { return Err("Expected '(' character in prototype declaration.") }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+            self.advance();
+            return Ok(Prototype { name: id, args: vec![], is_op: is_operator, prec: precedence });
+        }
+
+        let mut args = vec![];
+
+        loop {
+            match self.curr() {
+                Ident(name) => args.push(name),
+                _ => { return Err("Expected identifier in parameter declaration.") }
```
TheDan64: No need for brackets

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+
+            match self.curr() {
+                RParen => {
+                    self.advance();
+                    break;
+                },
+                Comma => {
+                    self.advance();   
+                },
+                _ => { return Err("Expected ',' or ')' character in prototype declaration.") }
```
TheDan64: No need for brackets

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+            },
+            _ => Err("Expected number literal.")
+        }
+    }
+
+    /// Parses an expression enclosed in parenthesis.
+    fn parse_paren_expr(&mut self) -> Result<Expr, &'static str> {
+        match self.current()? {
+            LParen => (),
+            _ => { return Err("Expected '(' character at start of parenthesized expression.") }
```
TheDan64: No need for brackets

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+        }
+
+        self.advance()?;
+
+        let expr = self.parse_expr()?;
+
+        match self.current()? {
+            RParen => (),
+            _ => { return Err("Expected ')' character at end of parenthesized expression.") }
```
TheDan64: No need for brackets

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+        self.advance();
+
+        Ok(expr)
+    }
+
+    /// Parses an expression that starts with an identifier (either a variable or a function call).
+    fn parse_id_expr(&mut self) -> Result<Expr, &'static str> {
+        let id = match self.curr() {
+            Ident(id) => id,
+            _ => { return Err("Expected identifier."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+
+                let mut args = vec![];
+                
+                loop {
+                    args.push(self.parse_expr()?);
+
+                    match self.current()? {
+                        Comma => (),
+                        RParen => break,
+                        _ => { return Err("Expected ',' character in function call."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+    }
+
+    /// Parses an unary expression.
+    fn parse_unary_expr(&mut self) -> Result<Expr, &'static str> {
+        let op = match self.current()? {
+            Op(ch) => {
+                self.advance()?;
+                ch
+            },
+            _ => { return self.parse_primary(); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+        loop {
+            let curr_prec = self.get_tok_precedence();
+
+            if curr_prec < prec || self.at_end() {
+                return Ok(left);
+            }
+
+            let op = match self.curr() {
+                Op(op) => op,
+                _ => { return Err("Invalid operator."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+    fn parse_conditional_expr(&mut self) -> Result<Expr, &'static str> {
+        // eat 'if' token
+        self.advance()?;
+
+        let cond = self.parse_expr()?;
+
+        // eat 'then' token
+        match self.current() {
+            Ok(Then) => { self.advance()? },
+            _ => { return Err("Expected 'then' keyword."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+        }
+
+        let then = self.parse_expr()?;
+
+        // eat 'else' token
+        match self.current() {
+            Ok(Else) => { self.advance()? },
+            _ => { return Err("Expected 'else' keyword."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+    }
+
+    /// Parses a loop for..in.. expression.
+    fn parse_for_expr(&mut self) -> Result<Expr, &'static str> {
+        // eat 'for' token
+        self.advance()?;
+
+        let name = match self.curr() {
+            Ident(n) => n,
+            _ => { return Err("Expected identifier in for loop."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+        };
+
+        // eat identifier
+        self.advance()?;
+
+        // eat '=' token
+        match self.curr() {
+            Op('=') => { self.advance()?; },
+            _ => { return Err("Expected '=' character in for loop."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+        }
+
+        let start = self.parse_expr()?;
+
+        // eat ',' token
+        match self.current()? {
+            Comma => { self.advance()?; },
+            _ => { return Err("Expected ',' character in for loop."); }
```
TheDan64: No need for brackets or semicolon

examples/kaleidoscope.rs
```diff
@@ -0,0 +1,1331 @@
...
+                Some(self.parse_expr()?)
+            },
+
+            _ => None
+        };
+
+        // eat 'in' token
+        match self.current()? {
+            In => { self.advance()?; },
+            _ => { return Err("Expected 'in' keyword in for loop."); }
```
TheDan64: No need for brackets or semicolon

TheDan64:
Some stylistic comments, but this PR is starting to look pretty good. Did you try running clippy on it by any chance?

6A:
Nope, I'll get to it right away. Also thanks for your stylistic comments, but it wasn't worth the trouble to comment every single mistake... And I just realized I let up something I had used for testing purposes... Oops.

TheDan64:
No trouble at all. I'll leave it up to you as to whether you want the space or not. Maybe consider something like Python's approach:
```python
> 1 + 1
2
```

6A:
I instead went with a different approach and prefixed all input with `?>`, all debug messages with `->`, all errors with `!>` and all output with `=>`. I also think everything should be good to go. Some examples have been added, as well as a README.

TheDan64:
Cool!

Issue #13, opened 2017-09-25
Linker error for doc tests on Travis-CI


TheDan64:
Had to mark certain doc tests as `no_run` because they were throwing a linking error on travis (possibly due to LLVM?) I could not reproduce on either of my local machines.

6A:
Doesn't happen on my machine either.

6A:
The logs show that the error comes from a `SIGKILL` signal, so one can guess that Travis is purposely killing the process. Maybe limiting the number of concurrent tests (using [`cargo test -- --test-threads 1`](https://users.rust-lang.org/t/how-are-cargo-tests-run-serially-or-in-parallel/1749/10)) might fix things?

6A:
## Some tests
I used the `time -v` command to see if there was a big difference between the tests that crash, and the ones that don't. In every case, the project needed to be built beforehand.

### With the [`ExecutionEngine::add_global_mapping`](https://github.com/6A/inkwell/blob/630647e02eb14cd69639456cac540c6667766896/src/execution_engine.rs#L57-L92) tests
```
        Command being timed: "cargo test"
	User time (seconds): 87.06
	System time (seconds): 87.98
	Percent of CPU this job got: 359%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:48.66
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 357852
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 2864079
	Voluntary context switches: 0
	Involuntary context switches: 0
	Swaps: 0
	File system inputs: 0
	File system outputs: 0
	Socket messages sent: 0
	Socket messages received: 0
	Signals delivered: 0
	Page size (bytes): 4096
	Exit status: 0
```

### Without
```
        Command being timed: "cargo test"
	User time (seconds): 85.96
	System time (seconds): 83.15
	Percent of CPU this job got: 328%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:51.52
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 361876
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
[...]

PR #20, opened 2017-10-04
Improvements to documentation

6A:
Modified some signatures for safety / readability, and added some doc comments.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -125,6 +125,19 @@ impl FloatPredicate {
     }
 }
 
+
+/// Defines the optimization level used to compile a `Module`.
+/// 
+/// # Remarks
+/// See also: http://llvm.org/doxygen/CodeGen_8h_source.html
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum OptimizationLevel {
+    None       = 0,
+    Less       = 1,
+    Default    = 2,
+    Aggressive = 3
+}
+
 // REVIEW: Maybe this belongs in some sort of prelude?
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum GlobalVisibility {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -130,12 +130,20 @@ impl FloatPredicate {
 /// 
 /// # Remarks
 /// See also: http://llvm.org/doxygen/CodeGen_8h_source.html
+#[repr(u32)]
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub enum OptimizationLevel {
     None       = 0,
     Less       = 1,
     Default    = 2,
     Aggressive = 3
+}
+
+impl Default for OptimizationLevel {
+    /// Returns the default value for `OptimizationLevel`, namely `OptimizationLevel::Default`.
+    fn default() -> Self {
+        OptimizationLevel::Default
+    }
 }
 
 // REVIEW: Maybe this belongs in some sort of prelude?
--- a/src/module.rs
+++ b/src/module.rs
@@ -11,12 +11,13 @@ use std::mem::{forget, uninitialized, zeroed};
 use std::path::Path;
 use std::slice::from_raw_parts;
 
+use OptimizationLevel;
 use context::{Context, ContextRef};
 use data_layout::DataLayout;
 use execution_engine::ExecutionEngine;
 use memory_buffer::MemoryBuffer;
 use types::{AsTypeRef, BasicType, FunctionType, BasicTypeEnum};
-use values::{AsValueRef, BasicValue, FunctionValue, GlobalValue, MetadataValue};
+use values::{AsValueRef, FunctionValue, GlobalValue, MetadataValue};
 
 // REVIEW: Maybe this should go into it's own module?
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
@@ -38,18 +39,6 @@ pub enum Linkage {
     PrivateLinkage,
     WeakAnyLinkage,
     WeakODRLinkage,
-}
-
-/// Defines the optimization level used to compile a `Module`.
-/// 
-/// # Remarks
-/// See also: http://llvm.org/doxygen/CodeGen_8h_source.html
-#[derive(Debug, PartialEq, Eq, Copy, Clone)]
-pub enum OptimizationLevel {
-    None       = 0,
-    Less       = 1,
-    Default    = 2,
-    Aggressive = 3
 }
 
 /// Defines the address space in which a global will be inserted.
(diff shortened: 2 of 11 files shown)


TheDan64:
CI is failing because of #13. For some reason you get linker errors when trying to run certain subsets of llvm via doctests. I'm guessing rustdoc doesn't know how to link it correctly? I'm not sure which doctests are specifically causing the problem but either way the workaround is to add `no_run` to the doc comment's code snippets, like so: https://github.com/TheDan64/inkwell/blob/cdc1560f4c247698684204d563b8a88b3fded2d3/src/builder.rs#L509

src/module.rs
```diff
@@ -40,6 +40,31 @@ pub enum Linkage {
     WeakODRLinkage,
 }
 
+/// Defines the optimization level used to compile a `Module`.
+/// 
+/// # Remarks
+/// See also: http://llvm.org/doxygen/CodeGen_8h_source.html
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum OptimizationLevel {
```
TheDan64: This enum already exists here under target.rs: https://github.com/TheDan64/inkwell/blob/cdc1560f4c247698684204d563b8a88b3fded2d3/src/targets.rs#L17-L21

Assigning values is definitely a good idea, but I think `Option<OptimizationLevel>` is a bit more rusty. It's easy enough to use internally:  `optimization_level_opt.unwrap_or(0) as i32` and I feel like it's easier for API users, since they can just toss in a `None` if they don't actually know what it does, and are just experimenting for example. Otherwise they have to import the enum and look up the variants.

Any thoughts on this? I'm alright with renaming `CodeGenOptLevel` to `OptimizationLevel` if you think that's a better name. I'm not quite sure where this should live though now that it's needed in two places

src/module.rs
```diff
@@ -40,6 +40,31 @@ pub enum Linkage {
...
+    Default    = 2,
+    Aggressive = 3
+}
+
+/// Defines the address space in which a global will be inserted.
+/// 
+/// # Remarks
+/// See also: http://llvm.org/doxygen/NVPTXBaseInfo_8h_source.html
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum AddressSpace {
```
TheDan64: I didn't know AddressSpace was an enum, good catch!

src/module.rs
```diff
@@ -146,6 +192,25 @@ impl Module {
...
+    /// use inkwell::module::Module;
+    /// 
+    /// let global_context = Context::get_global();
+    /// let global_module = Module::create("my_global_module");
+    /// 
+    /// assert_eq!(global_module.get_context(), global_context);
+    /// 
+    /// let local_context = Context::create();
+    /// let local_module = local_context.create_module("my_module");
+    /// let local_context = ContextRef::new(local_context);
```
TheDan64: This made me realize, I don't think I meant `ContextRef::new()` to be public, as it's only really useful when returned by Inkwell. Mind fixing this? It should probably be `pub(crate)`. If you want to compare contexts here, you should be able to dereference the `local_module`'s `ContextRef` to a `Context` and compare that.

src/module.rs
```diff
@@ -289,7 +367,17 @@ impl Module {
...
+    /// that describes whether or not it is.
+    /// 
+    /// * `print` - Whether or not a message describing the error should be printed
+    ///   to stderr, in case of failure.
+    /// 
+    /// # Remarks
+    /// See also: http://llvm.org/doxygen/Analysis_2Analysis_8cpp_source.html
     pub fn verify(&self, print: bool) -> bool {
+        // REVIEW <6A>: Maybe, instead of printing the module to stderr with `print`,
+        // we should return Result<(), String> that contains an error string?
```
TheDan64: Yeah I think There's another variant `LLVMVerifierFailureAction::LLVMAbortProcessAction` too, which prints to stderr and aborts the process. Not sure if we want to support that. But `Result<(), String>` does seem reasonable.

I also haven't been able to decide if we should return `&CStr`s because they point to already allocated memory or if we should reallocate to a `String` and dispose the original. A bigger design question for another time I suppose.

src/module.rs
6A: Oh, I didn't see, sorry. As you wish; it looks like most members in Inkwell are more verbose than in LLVM (for example: `build_float_add` vs `CreateFAdd`), which is why I spelled `OptimizationLevel` completely.

Is the fact that it is more "rusty" good? If so, that's actually what I used.  
In any case, choose what you find is best. To me, mapping enum values is better since it's a feature, and it allows to put all the logic in the definition of the enum, rather than on in the functions that use it / in a trait.

TheDan64: For names I've mostly been snake_casing llvm-sys names, lol. Keeping the logic in one place is a good point. Just trying to think about how the API will be used. Hmm.

src/module.rs
6A: I've been thinking about `&CStr`'s as well; in any case the user probably won't use them without `as_str` or `to_string_lossy` for example, and since `&str` is used everywhere else, it is rather strange; but it has indeed a performance cost.

In any case, I think aborting should be left to the user.

src/module.rs
6A: Alright, doing it now. I thought it was strange as well.

src/module.rs
6A: Oh, and since `OptimizationLevel` is an `u32`, it's faster to pass it by value (via copying) rather than by reference. I'm pushing the new changes now, and using my solution, at least for now.

6A:
Damn, such as shame Travis is doing this; I was actually using those tests.

TheDan64:
The doctests will still compile fine at least, but wont run

6A:
Oh, good

src/lib.rs
```diff
@@ -125,6 +125,19 @@ impl FloatPredicate {
     }
 }
 
+
+/// Defines the optimization level used to compile a `Module`.
+/// 
+/// # Remarks
+/// See also: http://llvm.org/doxygen/CodeGen_8h_source.html
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum OptimizationLevel {
```
TheDan64: I think you can do something like `#[repr(u32)]` so that you don't have to cast the enum as a u32?

src/targets.rs
```diff
@@ -541,15 +525,15 @@ impl Target {
         }
     }
 
-    pub fn create_target_machine(&self, triple: &str, cpu: &str, features: &str, level: Option<CodeGenOptLevel>, reloc_mode: RelocMode, code_model: CodeModel) -> Option<TargetMachine> {
+    pub fn create_target_machine(&self, triple: &str, cpu: &str, features: &str, level: Option<OptimizationLevel>, reloc_mode: RelocMode, code_model: CodeModel) -> Option<TargetMachine> {
```
TheDan64: We should probably use `OptimizationLevel` with None value or `Option<OptimizationLevel>`, but not both?

6A: Well, `None` is a member of  `OptimizationLevel` in LLVM, so I'd like to keep it in the enum; and I think `Option<OptimizationLevel>` means something like "either choose your own opt level, or we'll choose the default one for you", but it does seem redundant. Want me to change this to `OptimizationLevel`, and maybe impl `Default` for it?

TheDan64: It's times like this that make me wish rust had optional arguments haha. Yeah I guess `OptimizationLevel` with a default impl is fine

6A: Alright, I'll do it this way. Btw, I sent you some messages on the Gitter.

PR #21, opened 2017-10-11
Adding ExecutionEngine::add_global_mapping

6A:
It works on my machine (and some tests have been made related to it), so it should be good. I'll add `no_run` if Travis fails to build this.

### DIFF
--- a/examples/kaleidoscope/main.rs
+++ b/examples/kaleidoscope/main.rs
@@ -1193,16 +1193,15 @@ macro_rules! print_flush {
 // However, extern functions such as cos(x) and sin(x) can be imported without any problem.
 // Other lines related to this program can be found further down.
 
-// pub extern "C" fn putchard(x: f64) -> f64 {
-//     print_flush!("{}", x as u8 as char);
-//     x
-// }
+pub extern fn putchard(x: f64) -> f64 {
+    print_flush!("{}", x as u8 as char);
+    x
+}
 
-// pub extern "C" fn printd(x: f64) -> f64 {
-//     println!("Fn called");
-//     println!("{}", x);
-//     x
-// }
+pub extern fn printd(x: f64) -> f64 {
+    println!("{}", x);
+    x
+}
 
 /// Entry point of the program; acts as a REPL.
 pub fn main() {
@@ -1222,9 +1221,6 @@ pub fn main() {
 
     Target::initialize_native(&InitializationConfig::default()).expect("Failed to initialize native target.");
 
-    // add_symbol("printd", &printd, printd as *const ());
-    // add_symbol("putchard", &putchard, putchard as *const ());
-    
     let context = Context::create();
     let module = context.create_module("repl");
     let builder = context.create_builder();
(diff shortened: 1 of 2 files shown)


src/execution_engine.rs
```diff
@@ -51,11 +51,50 @@ impl ExecutionEngine {
...
 
-    // pub fn add_global_mapping(&mut self, global: &AnyValue, addr: *const ()) {
-    //     unsafe {
-    //         LLVMAddGlobalMapping(self.execution_engine, global.as_value_ref(), addr as *mut ::libc::c_void)
-    //     }
-    // }
+    /// Maps the specified value to an address.
+    ///
+    /// # Example
+    /// ```
```
TheDan64: Likely going to need no_run

src/execution_engine.rs
```diff
@@ -51,11 +51,50 @@ impl ExecutionEngine {
...
+    /// builder.build_return(Some(&retv));
+    ///
+    /// let mut ee = module.create_jit_execution_engine(OptimizationLevel::None).unwrap();
+    /// ee.add_global_mapping(&extf, sumf as *mut _);
+    ///
+    /// let result = unsafe { ee.run_function(&f, &[]) }.as_float(&ft);
+    ///
+    /// assert_eq!(result, 128.);
+    /// ```
+    pub fn add_global_mapping(&mut self, value: &AnyValue, addr: *const ()) {
```
TheDan64: I still think addr should be a `usize`, since it's safe to cast to a usize but not a `*const ()`?

6A:
Huh, gonna have to fix this. Do note that the [beta build](https://travis-ci.org/TheDan64/inkwell/jobs/286703862) actually passed, whereas the [stable](https://travis-ci.org/TheDan64/inkwell/jobs/286703861) and [nightly](https://travis-ci.org/TheDan64/inkwell/jobs/286703863) ones didn't.

This kind of "random" bug adds up to what I previously said on #13.

TheDan64:
Yeah, it's basically pure chance as to whether or not travis decides rustdoc is taking too long and kills it

src/execution_engine.rs
6A: Alright, switched to `usize`.

6A:
Once again a fail in something completely different that I haven't even touched...

TheDan64:
I restarted the job and it passed that time.

examples/kaleidoscope/main.rs
```diff
@@ -1193,16 +1193,15 @@ macro_rules! print_flush {
 // However, extern functions such as cos(x) and sin(x) can be imported without any problem.
 // Other lines related to this program can be found further down.
 
-// pub extern "C" fn putchard(x: f64) -> f64 {
-//     print_flush!("{}", x as u8 as char);
-//     x
-// }
+pub extern fn putchard(x: f64) -> f64 {
```
TheDan64: These two extern functions are no longer used.. did you mean to remove them?

6A: Ugh, I tried to use them from the Kaleidoscope REPL, but didn't succeed, and forgot to comment them once again.

PR #22, opened 2017-10-25
Made type and value enums clone- and copy-able.

### LINKED ISSUE #14: Types & Values should be Copy + Clone

6A:
## Description
Added `#[derive(Clone, Copy)]` to `AnyValueEnum`, `AggregateValueEnum`, `BasicValueEnum`, `BasicMetadataValueEnum`, `BasicTypeEnum` and `AnyTypeEnum`.

## Related Issue
Fixes #14.

## Motivation and Context
Allows values and types (which are opaque pointers) to be copied and cloned at will.

## How Has This Been Tested?
Existing tests have been run without a problem. The change is implemented by the Rust compiler, and shouldn't cause any issue.

## Types of Changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

### Required

- [X] My code follows the code style of this project
- [X] I have added or updated documentation (and doc tests) to any new or changed functions or types
- [ ] I have added tests to cover my changes
- [X] All new and existing tests passed

### Desired
- [ ] I have ran clippy and updated portions of code pertaining to my changes (Shouldn't be needed, since I added two keywords. Also Clippy doesn't compile on my machine).


TheDan64:
While we're at it, we should probably add Copy + Clone to the super traits?

6A:
Mhm, such as the following snippet?

```rust
pub trait AnyType: AsTypeRef + Debug + Clone + Copy { }
```

TheDan64:
Yup, I think that should suffice for both macros

6A:
Welp, that's not gonna be possible...  
Adding `Clone` or `Copy` requires the passed object to be `Sized`, which also means that any function accepting `AnyType` or `AnyValue` (for example) will no longer compile.

There are two solutions:
- Use generic functions (`fn do_something<V: BasicValue>(value: &V)` instead of `fn do_something(value: &BasicValue)`), which means that all existing function will be generated multiple times in the produced binary for a single purpose;
- Do not implement `Copy` or `Clone`.

Your thoughts?

TheDan64:
Generics are preferable to trait objects in almost all situations due to it being a compile time cost. The one case I can think of where trait objects are required is when we have input of mixed types (ie function parameters for one). We definitely can't get rid of those. So I guess we should just leave it out.

Maybe later we can add a Trait <-> Enum mapping which allows us to convert between them, which would allow you to clone. But for now this should suffice. Please be sure to add `Debug` at the very least.

src/values/traits.rs
```diff
@@ -22,6 +22,30 @@ macro_rules! trait_value_set {
     );
 }
 
+/// Represents an aggregate value, built on top of other values.
+pub trait AggregateValue: AnyValue {
```
TheDan64: I think this should be `pub trait AggregateValue: BasicValue`? Since `AggregateValue` is a subset of `BasicValue`?

src/types/traits.rs
```diff
@@ -14,13 +14,27 @@ pub trait AsTypeRef {
...
         $(
             impl $trait_name for $args {}
         )*
     );
 }
 
+/// Represents any LLVM type.
+pub trait AnyType: AsTypeRef + Debug {
+    /// Returns an `AnyTypeEnum` that represents the current type.
+    fn as_any_type_enum(&self) -> AnyTypeEnum {
```
TheDan64: It's interesting to note that these basically seem like associated types

6A: They kind of do, although not in a generic way

src/values/traits.rs
6A: Oh yep, good catch. I'll fix this as soon as I can. No other problem to fix here?

TheDan64: I'll give it one final look before I merge, but I did not see anything else

TheDan64:
Thanks for the PR!

PR #25, opened 2017-10-26
AddressSpace everywhere (fixes #24)

### LINKED ISSUE #24: AddressSpace is not used everywhere
Some methods accept `u32` as arguments instead of `AddressSpace`. An integer does not represent well what the target address space is, which is why the `AddressSpace` enum should be used instead.

(PR incoming)

6A:
## Description
Use `AddressSpace` instead of `u32` everywhere.

## Related Issue
#24.

## How This Has Been Tested
All existing tests have been run again.

## Breaking Changes
Many methods can no longer be called with an integer, and now need an enum. The methods should be clearer and safer to use, although some calls may have to be changed.

## Checklist
- [X] I have read the [Contributing Guide](../blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -61,6 +61,33 @@ pub fn shutdown_llvm() {
     }
 }
 
+/// Defines the address space in which a global will be inserted.
+///
+/// # Remarks
+/// See also: http://llvm.org/doxygen/NVPTXBaseInfo_8h_source.html
+#[repr(u32)]
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum AddressSpace {
+    Generic = 0,
+    Global  = 1,
+    Shared  = 3,
+    Const   = 4,
+    Local   = 5
+}
+
+impl From<u32> for AddressSpace {
+    fn from(val: u32) -> Self {
+        match val {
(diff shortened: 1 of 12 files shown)


6A:
Note: I also took the liberty of moving `AddressSpace` to `lib.rs`, since it is used in many different types, and does not really belong in `module.rs`.

TheDan64:
So, that contributing guide link tries to PR a branch :joy: 

re: Moving `AddressSpace`. Makes sense. I think eventually we'll better organize lib.rs and split it out into more well defined modules. For now, it makes a good place to keep the misfit functions and types.

TheDan64:
Also, going forward, would you mind rebasing onto master instead of merging master into your branch? It tends to add a lot of unnecessary merge commits. I'll be sure to work that into the contributing guide. Feel free to dispute it though if you have a good reason for doing so.

6A:
Huh, to be perfectly honest I didn't know there was an alternative to merging (and thus adding a commit "merge branch 'master'"), so thanks for actually telling me this. I'll try to see if I can avoid the extra commit next time I submit a PR.

TheDan64:
No problem!

Issue #1, opened 2017-06-29
Should support multiple LLVM-sys versions


TheDan64:
Probably by feature toggle, and conditional compilation. Either default to latest or require a feature to be selected (if possible)

- [x] v3.6 (Initial support, usable)
- [x] v3.6 (Version specific features)
- [x] v3.7 (Initial support, usable)
- [x] v3.7 (Version specific features)
- [x] v3.8 (Initial support, usable)
- [x] v3.8 (Version specific features)
- [x] v3.9 (Initial support, usable)
- [ ] v3.9 (Version specific features)
- [x] v4.0 (Initial support, usable)
- [ ] v4.0 (Version specific features)
- [x] v5.0 (Initial support, usable)
- [ ] v5.0 (Version specific features)
- [x] v6.0 (Initial support, usable)
- [ ] v6.0 (Version specific features)
- [x] v7.0 (Initial support, usable)
- [ ] v7.0 (Version specific features)
- [x] v8.0 (Initial support, usable)
- [ ] v8.0 (Version specific features)
- [x] v9.0 (Initial support, usable)
- [ ] v9.0 (Version specific features)
- [x] v10.0 (Initial support, usable)
- [ ] v10.0 (Version specific features)

TheDan64:
Worth noting that a particular version will only be considered supported if all functions *intended* to be supported are supported. That is, there are some functions that may be deemed too redundant to be useful.

For example "is_int()" is useful in LLVM because of their hyper abstraction, but in our IntType/IntValue it would never be useful and will be skipped because that piece of information is ingrained into the type system.

Or, if a function is deprecated from v3.6 onward (and likely has a replacement function), it may very well be skipped as well.

That said, nearly all LLVM functions should be supported, except in similarly outlying cases.

6A:
Cargo apparently does not support dependencies to be specified depending on the current feature set (see https://github.com/rust-lang/cargo/issues/4576). An alternative would be to force the user to:
1. Override inkwell's default `llvm_sys` version using the [[replace] section](http://doc.crates.io/manifest.html#the-replace-section).
2. Specify the feature associated with this version using the [[dependencies.inkwell] section](http://doc.crates.io/manifest.html#the-replace-section) and the `features` field.

Needless to say, this solution is very inconvenient, as it requires the user to make multiple changes, **and** may lead to conflicts.

TheDan64:
[...]

PR #31, opened 2018-01-06
Support for multiple LLVM versions, starting with 3.6

### LINKED ISSUE #3: Setup Travis CI for Linux
- [x] Initial Setup
- [ ] Multiple LLVM version setup to support #1

TheDan64:
## Description

* Updated .travis.yml to build one job per supported LLVM version
* Updated .travis.yml to share deployment procedures across jobs via anchors
* Added support for LLVM 3.6

## Related Issue

Updates #1 to officially support LLVM 3.6. Also fixes #3 

## How This Has Been Tested

Travis-CI have been updated to run the full test suite in both LLVM 3.6 & LLVM 3.7. I've also run 3.6 tests locally with success.

## Option\<Breaking Changes\>

There is a breaking change which requires an LLVM version feature flag to be selected. If not provided, it will throw a compile time error pointing to the README.md.

## Checklist

- [x] I have read the [Contributing Guide](CONTRIBUTING.md)


TheDan64:
Still needs a bit of additional work to throw a compiler error when no llvm version feature is specified and to update the README.md to document how to specify a version.

6A:
Wouldn't [compile_error!](https://doc.rust-lang.org/std/macro.compile_error.html) be the perfect match for this?

TheDan64:
Yes! I knew it existed, but forgot what it was called. Thanks!

TheDan64:
In case anyone is following this PR; the last remaining piece here is being able to override the llvm-sys version, which I have yet to have been able to figure out how to do successfully. If it's not possible, this will probably have to go on hold and I'll try and get cargo to support the changes if possible and focus on some other part of Inkwell in the mean time.

6A:
According to the rustc maintainers, right now Cargo cannot do this at all. Considering the few hacks we've tried, and that none of them worked, I think it's safe to say it cannot work yet unless some changes are made to Cargo itself.

Thus, I think we can only work on one version right now... Which raises the question, shall we focus on one version, or blindly add features to other versions as well? If we only focus on one version, then which one?

TheDan64:
Yeah - So I think we can work on more than one, but only have one as a default for Inkwell. I'm leaning towards merging this PR mostly as is (minus/modifying the documentation changes) so that we can keep testing against other LLVM versions in travis and keep working towards supporting other versions as far as this code base is concerned.

As far as the active inkwell version, I'm thinking about slowly updating it to the latest LLVM version(IE decide 3.7 is pretty good as is, and switch to a compilable initial implementation of 3.8, keeping the feature flag stuff for previous versions), since more people will probably want a newer version. This will allow us to have forward progress. It might be a breaking change for anyone currently using inkwell.. though probably not many people are using it yet, and it's better to do so early on.

If someone really needs to use an older version, they can fork or clone a local copy and modify the llvm-sys version and feature flag to the older version, and it should work (will document it). Far from ideal, but I will try and get some support from cargo even if I have to implement it myself haha

TheDan64:
One more option might be to explore if build scripts might be able to help somehow: https://doc.rust-lang.org/cargo/reference/build-scripts.html I don't have any concrete ideas yet, but it might be worth exploring if we could modify inkwell's Cargo.toml that way...

TheDan64:
Like, maybe we could require the user to pass in an environment variable "INKWELL_LLVM_VERSION='3.7'" or so, and the build script would do all of the modifications to the llvm-sys version and feature flags? Glancing through the above docs makes it sound like it might be possible

6A:
Indeed, build scripts can probably automate ways to achieve our goal, but it might not always be very "graceful." For example, we might be able to edit the dependency in `Cargo.toml` using `build.rs`, and trigger a rebuild thanks to the `rerun-if-changed` parameter.

TheDan64:
build script is starting to look like a no go either. I was able to get the compiler to set the required feature flag pretty easily, but I don't think I can change the required llvm-sys version since dependencies seem to be acquired before the build step. sigh.

TheDan64:
Random idea I had today that I'll log here so I don't forget: The approach we can take is to branch off master, with the only change being the llvm-sys version and default feature flag, and keep them otherwise rebased on master, which has cfgs for different versions.

This would mean we can't release an inkwell version to crates.io when we are ready to do so that includes multiple version support, though. I suppose we can have the crates.io version be hardcoded to the latest version of LLVM(-sys) and anyone desiring an older version of LLVM would be required to point at the appropriate GH branch.

As always, not an ideal approach, but this one would actually work, I think.

6A:
That sounds great to me, and that might be the only way to go for now, so this isn't a big problem.  
I don't have an example right now, but I know some widely used crates also suggest using the GitHub repo source instead of an hardcoded version on crates.io.

PR #38, opened 2018-04-14
Support LLVM 3.8

TheDan64:
## Description

* Add 3.8 build to travis
* Updated travis doc build to 3.8
* Added LLVM 3.8 support

## Related Issue

Updates #1 to support LLVM 3.8

## How This Has Been Tested

Tests are now passing with minor updates. A Linking error still needs to be fixed, so the travis build will fail.

## Option\<Breaking Changes\>

No breaking changes

## Checklist

- [x] I have read the [Contributing Guide](CONTRIBUTING.md)

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,2 @@
+[build]
+rustflags = ["-lffi"]
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,7 +94,7 @@ matrix:
 
 env:
   global:
-    - RUSTFLAGS="-C link-dead-code -C target-cpu=native"
+    - RUSTFLAGS="-C link-dead-code -C target-cpu=native -l ffi"
 
 after_success: |
   wget https://github.com/SimonKagstrom/kcov/archive/master.tar.gz &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,6 +55,17 @@ matrix:
           packages:
             - *BASE_PACKAGES
             - llvm-3.7-dev
+    - env:
+        - LLVM_VERSION="3.8"
+      <<: *BASE
+      addons:
+        apt:
+          sources:
+            - *BASE_SOURCES
+            - llvm-toolchain-precise-3.8
+          packages:
+            - *BASE_PACKAGES
+            - llvm-3.8-dev
     - deploy: # Documentation build; Only latest supported LLVM version for now
         provider: pages
         skip-cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ matrix:
         on:
           branch: master
       script:
-        - cargo doc --no-default-features --features llvm3-7
+        - cargo doc --no-default-features --features llvm3-8
         - echo '<meta http-equiv="refresh" content="1; url=inkwell/index.html">' > target/doc/index.html
       after_success:
       rust: nightly
@@ -84,11 +84,13 @@ matrix:
           sources:
             - *BASE_SOURCES
             # - llvm-toolchain-precise-3.6
-            - llvm-toolchain-precise-3.7
+            # - llvm-toolchain-precise-3.7
+            - llvm-toolchain-precise-3.8
           packages:
             - *BASE_PACKAGES
             # - llvm-3.6-dev
-            - llvm-3.7-dev
+            # - llvm-3.7-dev
+            - llvm-3.8-dev
 
 env:
   global:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,12 +14,13 @@ categories = ["development-tools::ffi"]
 default = []
 llvm3-6 = []
 llvm3-7 = []
+llvm3-8 = []
 
 [dependencies]
 either = "1.4.0"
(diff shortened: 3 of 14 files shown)


Michael-F-Bryan:
I'm not sure, but I believe updating `llvm-sys` from 37 to 38 broke the `llvm3-7` feature. I just updated my `Cargo.lock` from using e3a6f5457c6d8cfa355b54b9053510f74f6088b5 to 99f3e0a091e8598bee79d69f4ff7607449b46685 and now my `calc` fails with a compilation error.

```
   Compiling calc v0.1.0 (file:///home/michael/Documents/calc)
   Compiling inkwell v0.1.0 (https://github.com/TheDan64/inkwell#99f3e0a0)
error[E0432]: unresolved import `llvm_sys::core::LLVMEnablePrettyStackTrace`
  --> /home/michael/.cargo/git/checkouts/inkwell-946411d814d2c9f8/99f3e0a/src/lib.rs:39:9
   |
39 |     use llvm_sys::core::LLVMEnablePrettyStackTrace;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `LLVMEnablePrettyStackTrace` in `core`

error[E0432]: unresolved import `llvm_sys::core::LLVMResetFatalErrorHandler`
  --> /home/michael/.cargo/git/checkouts/inkwell-946411d814d2c9f8/99f3e0a/src/lib.rs:99:9
   |
99 |     use llvm_sys::core::LLVMResetFatalErrorHandler;
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `LLVMResetFatalErrorHandler` in `core`

error: aborting due to 2 previous errors

For more information about this error, try `rustc --explain E0432`.
error: Could not compile `inkwell`.
```

Changing to use the `llvm3-8` feature fixes the build though.

TheDan64:
99f3e0a091e8598bee79d69f4ff7607449b46685 is only on master at the moment, and master defaults to the latest supported llvm-sys. You should instead be pointed at the `llvm3-7` branch like so:
```toml
[dependencies]
inkwell = { git = "https://github.com/TheDan64/inkwell", branch = "llvm3-7" }
```

Michael-F-Bryan:
> 99f3e0a is only on master at the moment, and master defaults to the latest supported llvm-sys

Ah okay that makes sense, although it's probably a bug that we can't compile `master` for the different versions by using different feature flags (i.e. `git checkout master && cargo test --features llvm3-8 && cargo test --features llvm3-7` doesn't compile). If `master` will only work with `llvm3-8` then isn't it redundant to have the feature flags in the first place?

> Cargo doesn't support our use-case for multiple parallel versions very well today.

I think the recommended approach would be for different versions to be in different crates. Then we would conditionally enable the `llvm-sys-3-7` or `llvm-sys-3-8` crates using normal feature flags (assuming the relevant `llvm-sys` crates existed).

TheDan64:
I've reordered your questions in a way that I think makes it flow a little better:
> If master will only work with llvm3-8 then isn't it redundant to have the feature flags in the first place?

Well, no. Feature flags separate and keep track of llvm-sys version differences. Different versions add, remove, and modify paths to functions and types. The way llvm-sys handles LLVM differences is to have different branches with all changes applied per branch and this has a huge maintenance cost which is probably n branches * m changes per branch (IE if a bugfix needs to be applied across multiple versions, it has to be manually applied to those version branches which each have their own unique commit history).

Instead, for Inkwell, this would simply be a change to master and then the version branch rebases 1 version commit on master which 99.99% of the time is conflict free - so much easier to maintain in one place. This is roughly m changes to master + n branch rebases that all share the majority of their commit history. Master is the base branch that all others stem from. It's only a coincidence/byproduct that it has the latest version on it; because it isn't meant to be used directly. (The readme states what branches should be used - and actually I need to update it since specifying a feature isn't required nor is it useful since the branch applies the feature for you)

> it's probably a bug that we can't compile master for the different versions by using different feature flags (i.e. git checkout master && cargo test --features llvm3-8 && cargo test --features llvm3-7 doesn't compile).

I agree in theory, but in practice I see this as more of a Cargo deficiency because it doesn't support parallel versions at all. We're just making the best of what we've got. I agree that idealistically, `master` should work with any version, but that's exactly why we've reached this branch versioning issue in the first place: Cargo doesn't let us assign (llvm-sys) versions based on feature flags, which would make this much simpler. Using stuff like the `[replace]` section of `Cargo.toml` didn't really work out. See #1 for some of the online discussions I've had regarding how to approach this before we came to the conclusion of doing it this way.

> I think the recommended approach would be for different versions to be in different crates. Then we would conditionally enable the llvm-sys-3-7 or llvm-sys-3-8 crates using normal feature flags (assuming the relevant llvm-sys crates existed).

I'm not sure I understand. Do you mean separate version modules? Or separate crates as in part of a workspace? Either way, it sounds like a maintenance nightmare keeping common code in sync across versions (even if versions `x` and `y` both share a function, they might have different `llvm-sys` paths to that function) meaning you can't share a wrapper for that function across version module/sub-crates.

PR #47, opened 2018-06-22
Support vector types in math ops, fixes #45

### LINKED ISSUE #45: Support vector types in math operations
**Is your feature request related to a problem? Please describe.**
Yep -- in LLVM IR it is entirely valid to use mathematical operations on vector values, provided the two vectors have the same lengths and element type (and the element type is numeric, of course). However, all of Inkwell's math builder functions take and return `IntValue` or `FloatValue` values, meaning vectors can't easily be passed in without converting to a `BasicValueEnum` then back or using transmute.

**Describe the solution you'd like**
I'm not too sure what a solution would look like - a simple but messy solution might be to add functions like `build_float_vec_add` and `build_int_vec_add`, but you do lose the benefit of the typed parameters (although there's probably no way around that until #8 is resolved...)

I'm happy to make a pull request with the solution outlined above, although it's probably worth some discussion on a better alternative first?

cpdt:
This PR introduces the ability for math functions to take and return vector values in a type-safe way, as discussed in #45. Please note that I'm pretty new to Rust in general, so there are possibly some things here that can be done better :)

## Description

The PR primarily introduces two new traits, `FloatMathValue` and `IntMathValue`, which are implemented by `FloatValue`/`VectorValue` and `IntValue`/`VectorValue` respectively. Applicable methods on the `Builder` have been adjusted to take and return a type that implements these traits - meaning, for example, if you call `build_float_add` with a vector as the LHS, you must pass a vector for the RHS and you will get a vector back. Since there's no sub-type support yet we can't ensure the passed vectors are actually composed of integers or floats, however I assume this will be a part of 0.2.0.

There are some more complicated statements that require both a value and a target type, such as `build_float_to_signed_int` (which takes a floating point and new integer type, and returns an integer). LLVM allows us to pass in a vector as the value here, but also requires in that case that we pass in a vector type - in order to support this, the PR adds `FloatMathType` and `IntMathType` traits and associated types to navigate between value -> type -> converted type -> value. This means we can ensure the target type is valid (i.e is a vector when the input value is a vector, or an int type otherwise in the `build_float_to_signed_int` case), at the cost of some long type paths in `builder.rs`.

**Note:** this PR doesn't implement vector support for functions like `build_ptr_to_int`, although [LLVM does support vectors](https://llvm.org/docs/LangRef.html#ptrtoint-to-instruction) for these functions, since I didn't want to complicate the PR too much. I can add these in if it makes sense though.

## Breaking Changes

As per one of the requested changes, all math functions now take ownership of their input types/values instead of taking references.

## How This Has Been Tested

Due to some weird `llvm-sys` linking issues on my computer, I haven't actually been able to run the test suite (so we'll see if the CI passes :)). However, this PR is entirely type changes and shouldn't affect anything during runtime. Both the tests, the example, and my own personal project compile fine without any changes.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -5,8 +5,8 @@ use llvm_sys::{LLVMTypeKind, LLVMAtomicOrdering};
 
 use {IntPredicate, FloatPredicate};
 use basic_block::BasicBlock;
-use values::{AggregateValue, AsValueRef, BasicValue, BasicValueEnum, PhiValue, FunctionValue, IntValue, PointerValue, VectorValue, InstructionValue, GlobalValue, IntMathValue, FloatMathValue, InstructionOpcode};
-use types::{AsTypeRef, BasicType, PointerType, IntType, IntMathType, FloatMathType};
+use values::{AggregateValue, AsValueRef, BasicValue, BasicValueEnum, PhiValue, FunctionValue, IntValue, PointerValue, VectorValue, InstructionValue, GlobalValue, IntMathValue, FloatMathValue, PointerMathValue, InstructionOpcode};
+use types::{AsTypeRef, BasicType, PointerType, IntMathType, FloatMathType, PointerMathType};
 
 use std::ffi::CString;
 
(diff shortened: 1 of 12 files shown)


src/builder.rs
```diff
@@ -254,305 +254,305 @@ impl Builder {
     // TODO: Possibly make this generic over sign via struct metadata or subtypes
     // SubType: <I: IntSubType>(&self, lhs: &IntValue<I>, rhs: &IntValue<I>, name: &str) -> IntValue<I> {
     //     if I::sign() == Unsigned { LLVMBuildUDiv() } else { LLVMBuildSDiv() }
-    pub fn build_int_unsigned_div(&self, lhs: &IntValue, rhs: &IntValue, name: &str) -> IntValue {
+    pub fn build_int_unsigned_div<T: IntMathValue>(&self, lhs: &T, rhs: &T, name: &str) -> T {
```
TheDan64: The references here are artifacts of a time when `*Value`s and `*Type`s weren't `Copy`. Since you're touching this code, I would greatly appreciate if you could change this and the other methods you modified to take ownership of lhs/rhs trait types.

IE, the signature should be `..., lhs: T, rhs: T, ...`

Thanks!

src/builder.rs
```diff
@@ -794,25 +794,25 @@ impl Builder {
...
             LLVMBuildICmp(self.builder, op.as_llvm_predicate(), lhs.as_value_ref(), rhs.as_value_ref(), c_string.as_ptr())
         };
 
-        IntValue::new(value)
+        T::new(value)
     }
 
     // SubType: <F>(&self, op, lhs: &FloatValue<F>, rhs: &FloatValue<F>, name) -> IntValue<bool> { ?
-    pub fn build_float_compare(&self, op: FloatPredicate, lhs: &FloatValue, rhs: &FloatValue, name: &str) -> IntValue {
+    pub fn build_float_compare<T: FloatMathValue>(&self, op: FloatPredicate, lhs: &T, rhs: &T, name: &str) -> <<T::BaseType as FloatMathType>::ConvType as IntMathType>::ValueType {
```
TheDan64: I think the issue with `build_int_compare` and `build_float_compare` returning `T` and `<<T::BaseType as FloatMathType>::ConvType as IntMathType>::ValueType` (man is that ugly) respectively is that the return type is actually `IntValue<bool>` or `VecValue<IntValue<bool>>`.

This will be rendered incorrect once subtypes are implemented. IE input of `IntValue<32bit>` will return `IntValue<32bit>` when it should be `IntValue<bool>`

I think the way we might need to work around this with subtypes is to add more associated types to go from `IntValue<X>` -> `IntValue<bool>` & `VecValue<IntValue<X>>` -> `Vec<IntValue<bool>>` (and likewise for floats) somehow

...so I think this is fine as is, unless you can think of a good way to do that here. I would greatly appreciate adding another comment under the subtype comment either describing the above or pasting the URL to this comment

cpdt: Yeah, there's possibly a similar problem with functions like `build_int_cast`, if the `Type` types are also going to have subtypes.

There's currently [an RFC being worked on](https://github.com/rust-lang/rust/issues/44265) for generic associated types, which _I think_ would allow us to do something like `<<T::BaseType as FloatMathType>::ConvType<bool> as IntMathType>::ValueType` (notice how `ConvType` has a generic parameter with the type of integer to 'return') - a similar thing would also work for `BaseType` for the cast functions. But yeah, it's pretty ugly.

src/builder.rs
cpdt: Is it worth going through and changing all methods to take ownership, or would you rather I just changed the ones I touched?

TheDan64: Just the ones you have touched - I have to look at the remainder on a case by case basis.

cpdt:
@TheDan64 Yeah, for sure. I was hoping to include tests with the original PR but wasn't actually able to get LLVM to link (Windows ugh). I've managed to get that working in the meantime (since I needed it for the project I've started working on anyway) so I should be able to get those tests in soon :)

cpdt:
@TheDan64 Tests done :)

Issue #74, opened 2019-04-08
Extern functions are not detected by Kaleidoscope JIT


OlegTheCat:
**Describe the Bug**
[As stated in the comment](https://github.com/TheDan64/inkwell/blob/b541566a0ac18a0f52bef54596d00c740c8b0597/examples/kaleidoscope/main.rs#L1186-L1192) extern functions defined in Rust code are not detected by Kaleidoscope JIT.

**To Reproduce**
1. Run Kaleidoscope REPL.
2. Type `extern printd(x)`
3. Type `printd(10)`
4. Function returns `NaN`

**Expected Behavior**
`printd` function should have returned passed value and printed it.

**LLVM Version (please complete the following information):**
 - LLVM Version: 7.0.1
 - Inkwell Branch Used: llvm7-0

**Desktop (please complete the following information):**
 - OS: OS X

**Additional Context**
After a bit of investigation, it seems that there are a couple of problems here:

1. `rustc` will mangle the names of those function unless `#[no_mangle]` attribute is applied.
2. Unless those functions are used in the Rust code, the compiler will remove them. To avoid this they should be called somewhere or put into a static array with `#[used]` attribute like [here](https://github.com/rust-lang/rust/pull/39987).
3. Currently, [body that returns `NaN` is set for `extern` functions](https://github.com/TheDan64/inkwell/blob/b541566a0ac18a0f52bef54596d00c740c8b0597/examples/kaleidoscope/main.rs#L519-L523). This causes that functions that are defined in Rust are being overridden by those that return `NaN` (strangely, but functions like `sin` and `cos` are not being overridden). It seems that if we got `extern` definition, we should not attach any body to it.

I kinda [fixed](https://github.com/OlegTheCat/inkwell/commit/5c1ce9e0979354f34d0ea7df22c9bf51cf888110) these issues in my fork, but it will segfault if `extern` function, that is not actually defined, is called. I'm not sure what is desired behavior here. Let me know what do you think about this.

TheDan64:
@71 As the original author of this code, any thoughts on this?

71:
I remember struggling to make extern functions work, before giving up after trying several different things. I do remember trying to add `#[no_mangle]`, but am not sure I took the necessary steps to ensure the functions were not removed by the Rust compiler. I'm glad you managed to make it work!

[...]

PR #76, opened 2019-04-15
Fix detection of extern functions defined in Rust

OlegTheCat:
## Description

This fixes the ability of Kaleidoscope JIT to detect functions defined in Rust.

## Related Issue

#74 

## How This Has Been Tested

External functions have been tested in Kaleidoscope REPL:

```
?> extern printd(x)

?> printd(10)
10
=> 10

?> extern putchard(x)

?> putchard(33)
!=> 33
```

## Checklist

### DIFF
--- a/examples/kaleidoscope/main.rs
+++ b/examples/kaleidoscope/main.rs
@@ -286,11 +286,11 @@ pub struct Prototype {
     pub prec: usize
 }
 
-/// Defines a user-defined function.
+/// Defines a user-defined or external function.
 #[derive(Debug)]
 pub struct Function {
     pub prototype: Prototype,
-    pub body: Expr,
+    pub body: Option<Expr>,
     pub is_anon: bool
 }
 
@@ -502,7 +502,7 @@ impl<'a> Parser<'a> {
         // Return new function
         Ok(Function {
             prototype: proto,
-            body: body,
+            body: Some(body),
             is_anon: false
         })
     }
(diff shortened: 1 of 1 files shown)


OlegTheCat:
As discussed in #74, if one tries to execute a function that is not actually defined, it will lead to a segfault:

```
?> extern non_existing(x)

?> non_existing(1)
cargo: warning: cargo crashed, killed by SIGSEGV.
cargo itself has not crashed, but for transparency,
it will now kill itself (without dumping core) with the same signal


warnings can be silenced by the --no-warnings (-n) option
Segmentation fault: 11
```

From @71 [comment](https://github.com/TheDan64/inkwell/issues/74#issuecomment-480912770) it seemed to me that this is acceptable behavior, but I'm not sure that I've got everything right. Let's discuss it here if needed.

TheDan64:
I think the best thing to do would be to see what the C version of Kaleidoscope does - if it also segfaults on unknown extern functions then this is acceptable (and likely a limitation of the LLVM JIT), and if not then there's likely something we need to fix

OlegTheCat:
Yeah, makes sense. I'll try to investigate this thing.

OlegTheCat:
So, the C++ version seems not to segfault, but the issue is that it uses different API for JITting (which is old and deprecated right not, btw). I'm currently struggling to map it to the new API through inkwell <-> llvm-sys <-> llvm-c <-> llvm-c++ chain. It seems I'll need more time to investigate this.

Issue #79, opened 2019-04-19
Support setting linkage for GlobalValues


hansihe:
LLVM supports setting linkage on `GlobalValue`s with `GlobalValue::setLinkage`. I want this to be possible through inkwell as well.

I have considered using the C API directly for doing this, but I thought I would open an issue here first.

If it is wanted, I can make a PR implementing it.

As an aside, thanks for the project! It has saved me a lot of time.

TheDan64:
You're welcome, thanks for the kind words!

If the C api allows for it then we should absolutely support it. Please be sure to write a test or two!

TheDan64:
Fixed by #81

Issue #80, opened 2019-04-29
Segfault in test suite


hansihe:
Running the test suite, always in a test after `test_attributes::test_enum_attribute_kinds`, will sometimes result in a segfault.

On my machine it happens every 20 runs or so.

I can not reproduce this when running the tests single threaded.

I can try to provide more info if this is not reproducible for anyone else.

**LLVM Version (please complete the following information):**
 - LLVM Version: 7.0
 - Inkwell Branch Used: master with llvm7.0 feature

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04

**Additional Context**
```
hansihe@hansihe-XPS-13-9370:~/proj/rust/inkwell$ cargo test --features llvm7-0
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running target/debug/deps/inkwell-24beb4bfd67943ff

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/all-3b42dad18507485c

running 95 tests
test test_attributes::test_enum_attribute_kinds ... ok
error: process didn't exit successfully: `/home/hansihe/proj/rust/inkwell/target/debug/deps/all-3b42dad18507485c` (signal: 11, SIGSEGV: invalid memory reference)
```

TheDan64:
I've been struggling to reproduce/isolate this for a long time now, so thanks for this lead!

TheDan64:
Definitely seems to be a test in the `test_attributes` file; but I haven't been able to nail a specific test down yet. GDB seems to be of little help as it appears to be a Heisenbug. I might have to disable tests one by one until the SF stops appearing

nlewycky:
I've been running `cargo test --features=llvm7-0 -- --test-threads=1` to reduce the Heisenbugginess of the tests and it seems to help for me.

TheDan64:
This may have been fixed by recent changes to how we access the global context

TheDan64:
Closing as I haven't been able to reproduce since the latest changes

PR #81, opened 2019-04-29
Add ability to set linkage on GlobalValues

hansihe:
## Description

Add ability to set linkage on GlobalValues.

Added basic tests as well, if you want some more tests, tell me and I will add them.

## Related Issue
https://github.com/TheDan64/inkwell/issues/79

## How This Has Been Tested
I believe this should be a very localized change. There are some very basic tests.

I have tested this with LLVM 7.0 only. I presume CI will run tests for everything else.


hansihe:
Okey, so CI seems to hit the same issue as in https://github.com/TheDan64/inkwell/issues/80.

I can fully reproduce that issue in master before my PR.

TheDan64:
I restarted those builds, so that they end up passing. I'll take a look at that segfault you mentioned when I have some time (unless you'd like to look into fixing it) - thanks for filing it!

PR #84, opened 2019-06-19
bugfix: use CString when passing path across FFI

### LINKED ISSUE #83: Bitcode parse error (string ffi problem?)
**Describe the Bug**
`Module::parse_bitcode_from_path()` errors with `No such file or directory` even when the path does exist.  My reproduction case follows closely the example from the code comments on `parse_bitcode_from_path()`:

```rust
use inkwell::module::Module;
use std::path::Path;

fn main() {
    let path = Path::new("<path to existing, valid bitcode file>");
    if path.exists() {
        println!("Path definitely exists");
    } else {
        panic!("Path does not exist");
    }
    let _llvm_mod = Module::parse_bitcode_from_path(&path).expect("Failed to parse module");
}
```

The output I get from this example is
```
Path definitely exists
thread 'main' panicked at 'Failed to parse module: "No such file or directory"'
```

I was able to solve this problem on my machine with the one-line patch at https://github.com/TheDan64/inkwell/commit/781731b6b661be7a0164cafe131e2663f7759746.  My theory is that the existing code in `MemoryBuffer::create_from_file()` fails to null-terminate the filepath string before passing it to LLVM (simply using `to_str()` and `as_ptr()`), resulting in LLVM seeing a much longer filepath string (whatever data is in memory until the next null byte).  I'm happy to submit the patch linked above as a PR, but (a) I'm fairly new to Rust, and have no idea whether that patch is the "correct" fix; and (b) at any rate, the contribution guidelines here request an issue before PR.  Let me know if you want me to submit the PR.  It's also possible I'm completely misunderstanding something here or doing something wrong.

**Expected Behavior**
`Module::parse_bitcode_from_path()` should not error when given a path to an existing, valid bitcode file. Also, the example from the code comments on that function should pass.

 - LLVM Version: 7.0.1
 - Inkwell Branch Used: `llvm7-0`

 - OS: macOS 10.14.4

cdisselkoen:
## Description

One-line fix for #83

## Related Issue

Fixes #83

## How This Has Been Tested

The reproduction example in #83 throws an error without this patch, and does not throw an error after this patch has been applied (to the `llvm7-0` branch).  `cargo test` still passes on the `llvm7-0` branch after this patch has been applied.  I have not tested branches other than the `llvm7-0` branch, including `master` (although the patch applies cleanly), and have not tested on platforms other than macOS.

## Checklist

### DIFF
--- a/src/memory_buffer.rs
+++ b/src/memory_buffer.rs
@@ -26,7 +26,7 @@ impl MemoryBuffer {
     }
 
     pub fn create_from_file(path: &Path) -> Result<Self, LLVMString> {
-        let path = path.to_str().expect("Did not find a valid Unicode path string");
+        let path = CString::new(path.to_str().expect("Did not find a valid Unicode path string")).expect("Failed to convert to CString");
         let mut memory_buffer = ptr::null_mut();
         let mut err_string = unsafe { zeroed() };
 


TheDan64:
Thanks for fixing this!

PR #85, opened 2019-06-20
Derive Hash on a bunch of Value types

cdisselkoen:
## Description

This is a very small PR, so I hope you don't mind I submitted it without making an issue first (although I can certainly go back and make an issue if you want).  This PR simply adds `#[derive(Hash)]` to `Value`, its subtypes (e.g. `IntValue`), and the value-related enums (e.g. `BasicValueEnum`).  The primary motivation is so that downstream users can use any of these types as keys in `HashMap`s.  No functionality should be affected.

## How This Has Been Tested

I tested this on the `llvm7-0` branch and all tests still pass.

## Checklist

### DIFF
--- a/src/values/array_value.rs
+++ b/src/values/array_value.rs
@@ -9,7 +9,7 @@ use crate::types::ArrayType;
 use crate::values::traits::AsValueRef;
 use crate::values::{Value, InstructionValue, MetadataValue};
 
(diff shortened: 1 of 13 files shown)


TheDan64:
Thanks for the PR - this seems like a good idea. There was a spurious error in travis, so I've restarted the build. But I don't expect there to be any problems.

PR #87, opened 2019-06-20
add bindings for LLVMGet{I,F}CmpPredicate

### LINKED ISSUE #86: Any way to query the predicate of an `ICmp` or `FCmp`?
I was wondering if there is any way, given an `InstructionValue` with opcode `ICmp` or `FCmp`, to get the "predicate" of the instruction.  I.e., for the LLVM instruction `%3 = icmp slt i32 %0, %1`, I'm looking for how to get the `slt`.  My grepping through the `inkwell` codebase didn't turn up any uses of the LLVM C APIs `LLVMGetICmpPredicate()` or `LLVMGetFCmpPredicate()`.  If it's just that those bindings were missed, I'm happy to add.

cdisselkoen:
## Description

First shot at adding bindings for `LLVMGetICmpPredicate` and `LLVMGetFCmpPredicate`, functions in `llvm-sys` which currently aren't exposed.

Slightly unsure about the correct API here.  I have both of the bindings checking that the `InstructionValue` is indeed the correct operation (`ICmp` or `FCmp` respectively) before making the call to get the predicate; this will involve an extra unnecessary call into LLVM in the case where the caller does know that the opcode is correct.  A simpler binding would not perform this check (and therefore not have to return `Option` either).  I'm unsure what the behavior of LLVM is when calling, e.g., `LLVMGetICmpPredicate` on a non-`ICmp` instruction.

## Related Issue

Fixes #86

## How This Has Been Tested

New tests were added (well, ok, existing tests were extended) to cover both bindings.  All tests are passing when this patch is applied to the `llvm7-0` branch.  Not tested for other branches, or on platforms other than macOS.

## Checklist

### DIFF
--- a/tests/all/test_instruction_values.rs
+++ b/tests/all/test_instruction_values.rs
@@ -1,6 +1,6 @@
 extern crate inkwell;
 
-use self::inkwell::AddressSpace;
+use self::inkwell::{AddressSpace, IntPredicate, FloatPredicate};
 use self::inkwell::context::Context;
 use self::inkwell::values::{BasicValue, InstructionOpcode::*};
 
@@ -201,7 +201,7 @@ fn test_instructions() {
     let i64_type = context.i64_type();
     let f32_type = context.f32_type();
     let f32_ptr_type = f32_type.ptr_type(AddressSpace::Generic);
-    let fn_type = void_type.fn_type(&[f32_ptr_type.into()], false);
+    let fn_type = void_type.fn_type(&[f32_ptr_type.into(), f32_type.into()], false);
 
     let function = module.add_function("free_f32", fn_type, None);
     let basic_block = context.append_basic_block(&function, "entry");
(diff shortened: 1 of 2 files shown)


tests/all/test_instruction_values.rs
```diff
@@ -201,28 +201,38 @@ fn test_instructions() {
...
     let return_instruction = builder.build_return(None);
 
     assert_eq!(store_instruction.get_opcode(), Store);
     assert_eq!(ptr_val.as_instruction().unwrap().get_opcode(), PtrToInt);
     assert_eq!(ptr.as_instruction().unwrap().get_opcode(), IntToPtr);
+    assert_eq!(icmp.as_instruction().unwrap().get_opcode(), ICmp);
+    assert_eq!(icmp.as_instruction().unwrap().get_icmp_predicate().unwrap(), IntPredicate::EQ);
+    assert_eq!(f32_sum.as_instruction().unwrap().get_opcode(), FAdd);
+    assert_eq!(fcmp.as_instruction().unwrap().get_opcode(), FCmp);
+    assert_eq!(fcmp.as_instruction().unwrap().get_fcmp_predicate().unwrap(), FloatPredicate::OEQ);
```
TheDan64: This looks pretty good; but you should also check that using the wrong cmp call return None. IE get_icmp_predicate on a float cmp (or even some other random instruction) return None

cdisselkoen: Sure, very reasonable. Added some more checks in the test.

PR #88, opened 2019-06-23
Derive Clone, Copy, and Hash for BasicBlock

cdisselkoen:
## Description

Like many other structs in this crate, `BasicBlock` is under-the-hood just a wrapper around an LLVM ref; however, unlike most similar structs, it doesn't implement `Clone` or `Copy`.  I added derives for `Clone` and `Copy` (and while I was at it, `Hash`).  I'm presuming that this won't be contentious, unless I'm missing something about how `BasicBlock` behaves.

## Related Issue

None; can make one if discussion is needed.

## How This Has Been Tested

All tests still pass after this patch has been applied to the `llvm7-0` branch.  Not tested with other branches, or on platforms other than macOS.

## Checklist

### DIFF
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -19,7 +19,7 @@ use std::rc::Rc;
 /// A well formed `BasicBlock` is a list of non terminating instructions followed by a single terminating
 /// instruction. `BasicBlock`s are allowed to be malformed prior to running validation because it may be useful
 /// when constructing or modifying a program.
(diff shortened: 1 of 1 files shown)


TheDan64:
The change itself looks fine, but I'm trying to remember if there was a reason why we didn't make `BasicBlock`s `Clone` and `Copy`..

TheDan64:
Thanks for the PR!

PR #89, opened 2019-06-28
Add support for LLVM 8

bitwalker:
## Description

There are two main commits here:

- First, as per the PR title, this adds support for LLVM 8, or more specifically, adds the requisite feature and updates to `llvm-sys` version `80.0`
- Second, a refactoring of the `llvm_versions` procedural macro that adds support for single versions in addition to ranges, and also provides a cleaner path for future extensibility, as well as better error reporting. One such future improvement may be to change the syntax from `X.Y => X.Y` to `X.Y..X.Y` and `X.Y..=X.Y` to allow for both inclusive and exclusive ranges, whereas right now only inclusive ranges are possible.

## Related Issue

There isn't one, but presumably LLVM 8 support is desired, and I was making the changes on my fork anyway for use in my own project, so why not share :)

## How This Has Been Tested

So far I've only made sure that the code compiles and that the test suite passes successfully. I'll be working with the new version heavily in the coming days though, so any bugs that surface I will certainly be implementing fixes for and upstreaming here. For now though, I'm not aware of any breakages.

NOTE: I'm compiling on macOS Mojave, using llvm-sys 80.1.0, against a copy of LLVM that I have built (using the release80 branch). I have not yet tested on other architectures or operating systems. I also compiled using both stable and nighly. toolchains.

The actual changes required were pretty minimal from what I've seen so far, one set of soft deprecations (`LLVM{Has,Set}UnnamedAddr` => `LLVM{Get,Set}UnnamedAddress`), some compiler warnings, and one new addition, the `LLVMFNeg` opcode was added to the `LLVMOpcode` enum.

## Option\<Breaking Changes\>

None that I'm aware of

## Checklist

### DIFF
--- a/internal_macros/build.rs
+++ b/internal_macros/build.rs
@@ -1,0 +1,24 @@
+extern crate cargo_toml;
+
+use cargo_toml::Manifest;
+
+use std::path::Path;
+
+fn main() {
+    let manifest_path = "../Cargo.toml";
+    let manifest = Manifest::from_path(Path::new(manifest_path))
+        .expect("Unable to load Cargo manifest!");
+
+    let features = manifest.features
+        .keys()
+        .filter(|feature| feature.as_str().starts_with("llvm"))
+        .cloned()
+        .collect::<Vec<_>>();
+
+    let features_csv = features.as_slice().join(",");
+
+    // Exports a comma-separated feature list in the environment during compilation
+    // Can be fetched with `env!("INKWELL_FEATURES")`
+    println!("cargo:rustc-env=INKWELL_FEATURES={}", features_csv);
+    println!("cargo:rerun-if-changed={}", manifest_path);
+}
--- a/internal_macros/src/lib.rs
+++ b/internal_macros/src/lib.rs
@@ -8,11 +8,13 @@ extern crate syn;
 
 use proc_macro::TokenStream;
 use quote::quote;
-use syn::{Arm, Expr, ExprLit, Item, Lit, punctuated::Pair, Pat, PatLit, parse_macro_input};
+use syn::parse_macro_input;
+use syn::parse::{Parse, ParseStream, Result, Error};
+use syn::{Token, LitFloat, Ident, Item};
 
 // We could include_str! inkwell's Cargo.toml and extract these features
 // so that they don't need to also be managed here...
-const ALL_FEATURE_VERSIONS: [&str; 8] = [
+const ALL_FEATURE_VERSIONS: [&str; 9] = [
     "llvm3-6",
     "llvm3-7",
     "llvm3-8",
@@ -21,25 +23,41 @@ const ALL_FEATURE_VERSIONS: [&str; 8] = [
     "llvm5-0",
     "llvm6-0",
     "llvm7-0",
+    "llvm8-0",
 ];
 
 fn panic_with_usage() -> ! {
-    panic!("llvm_versions must take the form: X.Y => (X.Y || latest)");
+    panic!("llvm_versions must take the form: X.Y or X.Y => (X.Y || latest)");
 }
 
-fn get_latest_feature() -> String {
-    ALL_FEATURE_VERSIONS[ALL_FEATURE_VERSIONS.len() - 1].into()
+fn get_latest_feature_index() -> usize {
+    ALL_FEATURE_VERSIONS.len() - 1
 }
 
-fn get_feature_slice(start_feature: &str, end_feature: &str) -> Option<&'static [&'static str]> {
-    let start_index = ALL_FEATURE_VERSIONS.iter().position(|&str| str == start_feature)?;
-    let end_index = ALL_FEATURE_VERSIONS.iter().position(|&str| str == end_feature)?;
-
-    if end_index < start_index {
-        return None;
+fn get_feature_slice(vt: VersionType) -> Option<&'static [&'static str]> {
+    match vt {
+        VersionType::Specific(version) => {
+            let feature = f64_to_feature_string(version);
+            let index = ALL_FEATURE_VERSIONS.iter().position(|&s| s == feature)?;
+            Some(&ALL_FEATURE_VERSIONS[index..index])
+        }
+        VersionType::RangeToLatest(version) => {
+            let latest = get_latest_feature_index();
+            let feature = f64_to_feature_string(version);
+            let index = ALL_FEATURE_VERSIONS.iter().position(|&s| s == feature)?;
+            Some(&ALL_FEATURE_VERSIONS[index..=latest])
+        }
+        VersionType::Range(start, end) => {
+            let start_feature = f64_to_feature_string(start);
+            let end_feature = f64_to_feature_string(end);
+            let start_index = ALL_FEATURE_VERSIONS.iter().position(|&s| s == start_feature)?;
+            let end_index = ALL_FEATURE_VERSIONS.iter().position(|&s| s == end_feature)?;
+            if end_index < start_index {
+                return None;
+            }
+            Some(&ALL_FEATURE_VERSIONS[start_index..=end_index])
+        }
     }
-
-    Some(&ALL_FEATURE_VERSIONS[start_index..=end_index])
 }
 
 fn f64_to_feature_string(float: f64) -> String {
@@ -48,29 +66,52 @@ fn f64_to_feature_string(float: f64) -> String {
     format!("llvm{}-{}", int, (float * 10.) % 10.)
 }
 
+enum VersionType {
+    Specific(f64),
+    RangeToLatest(f64),
+    Range(f64, f64),
+}
+impl Parse for VersionType {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let lookahead = input.lookahead1();
+        if lookahead.peek(LitFloat) {
+            let from = input.parse::<LitFloat>().unwrap().value();
+            if input.is_empty() {
+                return Ok(VersionType::Specific(from));
+            }
+            let lookahead = input.lookahead1();
+            if lookahead.peek(Token![=>]) {
+                let _: Token![=>] = input.parse().unwrap();
+                let lookahead = input.lookahead1();
+                if lookahead.peek(Ident) {
+                    let to = input.parse::<Ident>().unwrap();
+                    if to.to_string() == "latest" {
+                        Ok(VersionType::RangeToLatest(from))
+                    } else {
+                        Err(Error::new(to.span(), "expected `latest`"))
+                    }
+                } else if lookahead.peek(LitFloat) {
+                    let to = input.parse::<LitFloat>().unwrap().value();
+                    Ok(VersionType::Range(from, to))
+                } else {
+                    Err(lookahead.error())
+                }
+            } else {
+                Err(lookahead.error())
+            }
+        } else {
+            Err(lookahead.error())
+        }
+    }
+}
+
 #[proc_macro_attribute]
 pub fn llvm_versions(attribute_args: TokenStream, attributee: TokenStream) -> TokenStream {
-    let arm = parse_macro_input!(attribute_args as Arm);
+    let vt = parse_macro_input!(attribute_args as VersionType);
 
-    if arm.pats.len() != 1 {
-        panic_with_usage();
-    }
+    let features = get_feature_slice(vt)
+        .unwrap_or_else(|| panic_with_usage());
 
-    let start_feature = match arm.pats.first().unwrap() {
-        Pair::End(Pat::Lit(PatLit { expr })) => match **expr {
-            Expr::Lit(ExprLit { lit: Lit::Float(ref literal_float), .. }) => f64_to_feature_string(literal_float.value()),
-            _ => panic_with_usage(),
-        },
-        _ => panic_with_usage(),
-    };
-    let end_feature = match *arm.body {
-        Expr::Lit(ExprLit { lit: Lit::Float(literal_float), .. }) => Some(f64_to_feature_string(literal_float.value())),
-        // We could assert the path is just "latest" but this seems like a lot of extra work...
-        Expr::Path(_) => None,
-        _ => panic_with_usage(),
-    }.unwrap_or_else(get_latest_feature);
-
-    let features = get_feature_slice(&start_feature, &end_feature).unwrap_or_else(|| panic_with_usage());
     let attributee: Item = syn::parse(attributee).expect("Could not parse attributed item");
 
     let q = quote! {
--- a/internal_macros/src/lib.rs
+++ b/internal_macros/src/lib.rs
@@ -18,19 +18,19 @@ use syn::{Token, LitFloat, Ident, Item};
 
 const FEATURES_ENV: &'static str = env!("INKWELL_FEATURES");
 
-// Fetches a vector of feature version strings, e.g. llvm8-0
+/// Fetches a vector of feature version strings, e.g. llvm8-0
 fn get_feature_versions() -> Vec<&'static str> {
     FEATURES_ENV
         .split(',')
         .collect()
 }
 
-// Gets the index of the feature version that represents `latest`
+/// Gets the index of the feature version that represents `latest`
 fn get_latest_feature_index(features: &[&str]) -> usize {
     features.len() - 1
 }
 
-// Gets the index of the feature version that matches the input string
+/// Gets the index of the feature version that matches the input string
 fn get_feature_index(features: &[&str], feature: String, span: Span) -> Result<usize> {
     let feat = feature.as_str();
     match features.iter().position(|&s| s == feat) {
@@ -39,7 +39,7 @@ fn get_feature_index(features: &[&str], feature: String, span: Span) -> Result<u
     }
 }
 
-// Gets a vector of feature versions represented by the given VersionType
+/// Gets a vector of feature versions represented by the given VersionType
 fn get_features(vt: VersionType) -> Result<Vec<&'static str>> {
     let features = get_feature_versions();
     let latest = get_latest_feature_index(&features);
@@ -94,8 +94,8 @@ fn get_features(vt: VersionType) -> Result<Vec<&'static str>> {
     }
 }
 
-// Converts a version number as a float to its feature version 
-// string form (e.g. 8.0 => llvm8-0)
+/// Converts a version number as a float to its feature version 
+/// string form (e.g. 8.0 ..= llvm8-0)
 fn f64_to_feature_string(float: f64) -> String {
     let int = float as u64;
 
--- a/internal_macros/src/lib.rs
+++ b/internal_macros/src/lib.rs
@@ -3,97 +3,160 @@
 //! Here be dragons 🐉
 
 extern crate proc_macro;
+extern crate proc_macro2;
 extern crate quote;
 extern crate syn;
 
+use std::iter::IntoIterator;
+
 use proc_macro::TokenStream;
+use proc_macro2::Span;
 use quote::quote;
 use syn::parse_macro_input;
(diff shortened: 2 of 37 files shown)


TheDan64:
Thanks for submitting this PR!

Unfortunately I'm traveling for the next week and a half or so, so I won't be able to get to reviewing this immediately. However I look forward to doing so upon my return!

Best,
Dan

bitwalker:
No problem @TheDan64! In the meantime, I'll make sure the Travis build passes.

I also went ahead and finished the refactor of `llvm_versions`, namely implementing support for both inclusive and exclusive ranges, moving away from the non-standard range syntax (`7.0 => latest`) to normal Rust range syntax (`7.0..=latest` for inclusive ranges and `7.0..8.0` for exclusive ranges). This should be much nicer to use in the future, and the errors it produces now are a lot better.

bitwalker:
Oh, and I forgot to mention, but now the `llvm_versions` macro doesn't have the features hardcoded anymore, it reads them from the root `Cargo.toml` manifest, which should ease maintenance on that front.

TheDan64:
Seems like there's a missing import in the older llvm versions? See travis

bitwalker:
@TheDan64 Ah, missed the last build failure there - seems like `LLVMCreateTargetDataLayout` is only supported after 3.9, I've updated the imports and function definition accordingly.

TheDan64:
The documentation build isn't working, I think you need to update .travis.yml to use LLVM 8 for doc builds (since the new default version is 8)

bitwalker:
Yeah looks like that's the issue (and I notice that I was also missing a build env specifically for LLVM 8, so I added that too). I'll have to wait on the build to know for sure though.

I also have some additional fixes staged up for a few other things I've found as I'm going along, but I'll open another PR for those things when this one is wrapped up.

internal_macros/build.rs
```diff
@@ -0,0 +1,24 @@
...
+        .keys()
+        .filter(|feature| feature.as_str().starts_with("llvm"))
+        .cloned()
+        .collect::<Vec<_>>();
+
+    let features_csv = features.as_slice().join(",");
+
+    // Exports a comma-separated feature list in the environment during compilation
+    // Can be fetched with `env!("INKWELL_FEATURES")`
+    println!("cargo:rustc-env=INKWELL_FEATURES={}", features_csv);
```
TheDan64: Pretty cool idea to parse the Cargo.toml!

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
-];
-
-fn panic_with_usage() -> ! {
-    panic!("llvm_versions must take the form: X.Y => (X.Y || latest)");
-}
+use syn::parse_macro_input;
+use syn::parse::{Parse, ParseStream, Result, Error};
+use syn::{Token, LitFloat, Ident, Item};
+
+const FEATURES_ENV: &'static str = env!("INKWELL_FEATURES");
```
TheDan64: Hmm since this all runs at compile time I'm a little confused about performance implications of const vs static here.

Normally const means it's evaluated once at compile time and inserted at all call sites... but we're already in compile time... so what happens? 😄

I wonder if a global variable is better here...
Maybe ideally `get_feature_versions` would be a const function, but those aren't stable yet.
Maybe I'm thinking about it too much. Just don't want a compile time bloat

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
     }
+}
 
-    Some(&ALL_FEATURE_VERSIONS[start_index..=end_index])
+// Gets a vector of feature versions represented by the given VersionType
+fn get_features(vt: VersionType) -> Result<Vec<&'static str>> {
+    let features = get_feature_versions();
+    let latest = get_latest_feature_index(&features);
+    match vt {
+        VersionType::Specific(version, span) => {
```
TheDan64: This `VersionType` enum is a great idea, thanks!

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
-}
+use syn::parse_macro_input;
+use syn::parse::{Parse, ParseStream, Result, Error};
+use syn::{Token, LitFloat, Ident, Item};
+
+const FEATURES_ENV: &'static str = env!("INKWELL_FEATURES");
 
-fn get_latest_feature() -> String {
-    ALL_FEATURE_VERSIONS[ALL_FEATURE_VERSIONS.len() - 1].into()
+// Fetches a vector of feature version strings, e.g. llvm8-0
```
TheDan64:
```suggestion
/// Fetches a vector of feature version strings, e.g. llvm8-0
```

Doc string?

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
+fn get_feature_versions() -> Vec<&'static str> {
+    FEATURES_ENV
+        .split(',')
+        .collect()
 }
 
-fn get_feature_slice(start_feature: &str, end_feature: &str) -> Option<&'static [&'static str]> {
-    let start_index = ALL_FEATURE_VERSIONS.iter().position(|&str| str == start_feature)?;
-    let end_index = ALL_FEATURE_VERSIONS.iter().position(|&str| str == end_feature)?;
+// Gets the index of the feature version that represents `latest`
```
TheDan64:
```suggestion
/// Gets the index of the feature version that represents `latest`
```
Doc string?

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
+fn get_latest_feature_index(features: &[&str]) -> usize {
+    features.len() - 1
+}
 
-    if end_index < start_index {
-        return None;
+// Gets the index of the feature version that matches the input string
```
TheDan64:
```suggestion
/// Gets the index of the feature version that matches the input string
```
Doc string?

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
+fn get_feature_index(features: &[&str], feature: String, span: Span) -> Result<usize> {
+    let feat = feature.as_str();
+    match features.iter().position(|&s| s == feat) {
+        None => Err(Error::new(span, format!("Invalid feature version: {}, not defined", feature))),
+        Some(index) => Ok(index)
     }
+}
 
-    Some(&ALL_FEATURE_VERSIONS[start_index..=end_index])
+// Gets a vector of feature versions represented by the given VersionType
```
TheDan64:
```suggestion
/// Gets a vector of feature versions represented by the given VersionType
```
Doc string?

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
+                Err(Error::new(end_span, message))
+            } else {
+                Ok(features[start_index..end_index].to_vec())
+            }
+        }
+    }
 }
 
+// Converts a version number as a float to its feature version 
+// string form (e.g. 8.0 => llvm8-0)
```
TheDan64:
```suggestion
/// string form (e.g. 8.0 ..= llvm8-0)
```

TheDan64:
Not sure why the docs build is still failing... your docs changes look right. But seems to be something with syn?

internal_macros/src/lib.rs
```diff
@@ -3,74 +3,194 @@
...
+                let message = format!("Invalid version range: {} must be greater than {}", start, end);
+                Err(Error::new(end_span, message))
+            } else {
+                Ok(features[start_index..end_index].to_vec())
+            }
+        }
+    }
 }
 
+// Converts a version number as a float to its feature version
```
TheDan64:
```suggestion
/// Converts a version number as a float to its feature version 
```

src/execution_engine.rs
```diff
@@ -318,7 +318,7 @@ impl ExecutionEngine {
 
         // LLVMGetFunctionAddress segfaults in llvm 5.0 -> 7.0 when fn_name doesn't exist. This is a workaround
```
TheDan64:
```suggestion
        // LLVMGetFunctionAddress segfaults in llvm 5.0 -> 8.0 when fn_name doesn't exist. This is a workaround
```

src/passes.rs
```diff
@@ -1040,16 +1046,25 @@ impl<T: PassManagerSubType> PassManager<T> {
         }
     }
 
-    #[llvm_versions(7.0 => latest)]
+    #[llvm_versions(7.0)]
```
TheDan64: Was this removed in 8.0?

src/values/instruction_value.rs
```diff
@@ -38,6 +38,8 @@ pub enum InstructionOpcode {
     CleanupRet,
     ExtractElement,
     ExtractValue,
+    #[cfg(any(feature = "llvm8-0"))]
+    FNeg,
```
TheDan64: Does `llvm_versions` not yet work on enum variants? :( Wish proc macros would keep advancing
Although it's verbose, I would prefer this to be `#[cfg(not(any(feature = "llvm3-6", ...., "llvm7-0")))]` so that it's future proofed. Same for the ones later in this file.

tests/all/test_values.rs
```diff
@@ -321,10 +321,10 @@ fn test_verify_fn() {
 
     let function = module.add_function("fn", fn_type, None);
 
-    #[cfg(not(any(feature = "llvm3-9", feature = "llvm4-0", feature = "llvm5-0", feature = "llvm6-0", feature = "llvm7-0")))]
+    #[cfg(not(any(feature = "llvm3-9", feature = "llvm4-0", feature = "llvm5-0", feature = "llvm6-0", feature = "llvm7-0", feature = "llvm8-0")))]
     assert!(!function.verify(false));
     // REVIEW: Why does 3.9 -> 7.0 return true here? LLVM bug? Bugfix?
```
TheDan64:
```suggestion
     // REVIEW: Why does 3.9 -> 8.0 return true here? LLVM bug? Bugfix?
```

src/passes.rs
bitwalker: It was moved apparently, from the `llvm_sys::transforms::scalar` module to the `llvm_sys::transforms::aggressive_instcombine` module.

src/values/instruction_value.rs
bitwalker: I think there was a reason it didn't work, but I can't recall if I actually tried to make it work or not. I'll see if that's doable, otherwise I'll make the change you've described

internal_macros/src/lib.rs
bitwalker: I would be very surprised if there was overhead using a const here, especially since it has one usage in `get_feature_versions`, so at worst it would get folded in to that function. In general though, it may be possible for it to be less ideal than a global static, but its going to depend heavily on what compiler settings are used (i.e. is it optimizing for speed or size) and how large the values are, since I suspect LLVM doesn't like to replace usages of constants with the constant value itself when that value is larger than a few words. The only way to know for sure though is to do some side-by-side comparisons.

As for why I chose const vs static here, it was just a semantic choice - the value is constant from the point of view of the module definition (i.e. when it is built, the value is fed in from the environment and it never changes). A static would also work just fine, and I think for our purposes there is likely no meaningful difference between the two, but until I know for sure, I will defer to your preference, just let me know how you'd like it to be set up :)

bitwalker:
Looking at the last failed build, the docs successfully built, but there were spurious SIGSEGV failures in the tests. These seem to happen in different LLVM versions randomly in Travis, and I can't tell if it is the Travis environment, or something in the tests, but I'm pretty confident that those failures precede my changes. Unfortunately without being able to step through the code in a debugger, we're basically SOL on figuring out why those tests fail occasionally :(.

It may be worth looking into some property-based testing at some point as a way to fuzz the library a bit - then just attach a debugger with a breakpoint on panics and step backwards if a SIGSEGV is hit. The only alternative seems to be doing the same thing, but running the same test suite over and over again until it fails.

nlewycky:
Before going to fuzzing, please make sure to run the tests against an asserts-enabled build of LLVM. The binary releases from releases.llvm.org have assertions disabled. LLVM tends to use asserts to detect inputs before it does things that would cause undefined behaviour. The trouble with undefined behaviour in C++ code is that it could do anything, including work in the fuzzer but crash in production.

I've been debugging the crashes and fixing them on my own branch https://github.com/wasmerio/inkwell/tree/debug which has commits that I think are clean enough to merge (yet collectively cause the tests to fail most-or-all of the time unfortunately) and a second branch with untriaged cleanups https://github.com/wasmerio/inkwell/tree/debug2 . Removing code from the tests that makes LLVM misbehave is easy, what I'd like input on is how to make the APIs safe by design. (I don't have any reason to think that this will be difficult, I'm simply not familiar with rust and rust API design in particular. I'm a C++ LLVM developer.)

bitwalker:
@nlewycky Definitely agree, especially on enforcing invariants in the API. 

That said, I think that goal is confounded a bit by the (at least current) requirement to be backwards compatible all the way to 3.6. APIs have to correctly enforce invariants for all versions, which means carefully replicating the rules from the original C++, an enormous effort. 

Personally I think the project could do more in this area if it only had to support the most recent stable version, or perhaps the latest two releases, but beyond that it becomes extremely difficult to maintain (IMO).

TheDan64:
> I've been debugging the crashes and fixing them on my own branch https://github.com/wasmerio/inkwell/tree/debug which has commits that I think are clean enough to merge (yet collectively cause the tests to fail most-or-all of the time unfortunately) and a second branch with untriaged cleanups https://github.com/wasmerio/inkwell/tree/debug2 . Removing code from the tests that makes LLVM misbehave is easy, what I'd like input on is how to make the APIs safe by design. (I don't have any reason to think that this will be difficult, I'm simply not familiar with rust and rust API design in particular. I'm a C++ LLVM developer.)

Thanks for all of this work! I started looking at your two PRs but I'd like to merge this one before #96 due to the size and complexity of both.

> Personally I think the project could do more in this area if it only had to support the most recent stable version, or perhaps the latest two releases, but beyond that it becomes extremely difficult to maintain (IMO).

Not sure I totally agree. Sure it's a bit more of a hassle but it's not like LLVM 3.6 and 8.0 are entirely different beasts. And things like [llvmenv](https://github.com/termoshtt/llvmenv/) make it a lot easier to switch between versions. Plus travis builds. That said, I'm not against coming to the point where we're like "9.0 just came out and we should drop the two oldest versions, 3.6 & 3.7" for example. I just don't want to regularly pull versions out from underneath users' feet just because it's a little inconvenient

src/passes.rs
TheDan64: Ohh I see.

src/values/instruction_value.rs
TheDan64: Any luck with this?

bitwalker: @TheDan64 Sorry about the delay, had my hands full the past couple days, I'll try and get to this tonight or in the morning!

TheDan64: No problem!

bitwalker: @TheDan64 I've pushed a commit which addresses this by extending the places `llvm_versions` can be used, and adds a new macro `llvm_enum` which I think may come in handy, in certainly cleaned up the implementation of `InstructionOpcode` nicely. Mostly that part was motivated by the problem of not being able to decorate expressions with attributes (on stable anyway), as well as not being able to do the same for match arms, even when they do nothing but generate `#[cfg(..)]` attributes in their place.

When that stuff makes it into stable, then the need for `#[llvm_versioned_item]` goes away (which isn't used yet, but I suspect it either could be, or will be eventually). The `llvm_enum` attribute still has the benefit of automating the boilerplate for conversions between LLVM enums and Inkwell enums like in the case of `InstructionOpcode`, but not sure how common that really is in practice, so perhaps it could be removed eventually as well.

Anyway, hope that makes maintaining a few of these things a bit less annoying :)

TheDan64: `llvm_enum` looks really neat! This is pretty awesome!

> not sure how common that really is in practice, so perhaps it could be removed eventually as well.

This is actually really common, we already have a `macro_rules` macro for this already (though I think your version is better) so maybe we should replace that with this?

IE see https://github.com/TheDan64/inkwell/blob/e1854817e5b8814ad7776a1839c57392596f74e0/src/comdat.rs#L8-L22 but there are a bunch more.

One important aspect that should work in yours (it might already since you track attrs) is applying doc comments like `enum_rename` does

bitwalker: I I think we can extend the macro to cover the usages of `enum_rename`, may be better to do that in its own PR though. 

Otherwise, the macro preserves all attributes IIRC, but I will confirm that when I'm at my laptop here in a bit.

bitwalker: Confirmed, doc comments on the enum variants are preserved (as are any other attributes I expect). Let me know if there is anything else outstanding!

TheDan64:
Thanks for the PR!

bitwalker:
No problem! Working on the newly released 9.0 now haha

TheDan64:
9 is release already??

bitwalker:
Haha yeah I couldn't believe it either, just a few days ago I think, `release90`. Has some key WebAssembly improvements, so definitely glad to see it!

TheDan64:
Doesn't look like there are any llvm-sys bindings for it yet

bitwalker:
Yeah I have to start there first

TheDan64:
@bitwalker When I try and build inkwell master locally I'm getting:
```rust
   Compiling inkwell_internal_macros v0.1.0 (/d/LinuxData/repos/inkwell/internal_macros)
error[E0599]: no method named `is_ident` found for type `syn::Path` in the current scope
   --> internal_macros/src/lib.rs:236:23
    |
236 |         if !attr.path.is_ident("llvm_versions") {
    |                       ^^^^^^^^

error[E0599]: no method named `is_ident` found for type `syn::Path` in the current scope
   --> internal_macros/src/lib.rs:364:40
    |
364 |         attrs.retain(|attr| !attr.path.is_ident("llvm_variant"));
    |                                        ^^^^^^^^

error[E0599]: no method named `is_ident` found for type `syn::Path` in the current scope
   --> internal_macros/src/lib.rs:376:40
    |
376 |         attrs.retain(|attr| !attr.path.is_ident("llvm_variant"));
    |                                        ^^^^^^^^

error[E0599]: no method named `is_ident` found for type `syn::Path` in the current scope
   --> internal_macros/src/lib.rs:425:72
    |
425 |         if let Some(attr) = variant.attrs.iter().find(|attr| attr.path.is_ident("llvm_variant")) {
    |                                                                        ^^^^^^^^

error[E0599]: no method named `is_ident` found for type `syn::Path` in the current scope
   --> internal_macros/src/lib.rs:435:64
    |
435 |                         variant.attrs.retain(|attr| !attr.path.is_ident("llvm_variant"));
    |                                                                ^^^^^^^^

error: aborting due to 5 previous errors

For more information about this error, try `rustc --explain E0599`.
error: Could not compile `inkwell_internal_macros`.
```

Any idea why? Seems to work fine in CI

TheDan64:
What's also confusing is that the docs https://docs.rs/syn/0.15.40/syn/struct.Path.html#method.is_ident seem to suggest that syn needs the `"parsing"` feature enabled, but `parsing` is enabled by default...

TheDan64:
Doing `cargo clean` didn't help; but re-cloning the repo fixed it 🤷‍♂

bitwalker:
@TheDan64 Sorry for the delay, haven't been active much today. Definitely sounds like some kind of build artifact got stuck, maybe something with incremental compilation? Not sure what would have caused that otherwise, since as you say CI seems to work fine. I haven't hit that myself, but I'll keep an eye out and if it does I'll see if I can figure out the cause.

PR #92, opened 2019-07-02
Add support for llvm 8.0.

nlewycky:
## Description

Add support for llvm 8.0:
 * Add new FNeg instruction.
 * LLVMAddAggressiveInstCombinerPass moved to llvm_sys::transforms::aggressive_instcombine.

## How This Has Been Tested

See issues I've filed recently. The tests fail, but they were already failing. I tested this by commenting out the bad parts and running cargo test under valgrind.

## Checklist

- [x] I have read the [Contributing Guide](https://github.com/TheDan64/inkwell/blob/master/.github/CONTRIBUTING.md)

Issue #73, opened 2019-04-06
Test case failed when we use a llvm which enable LLVM_ENABLE_ASSERTIONS option


longwusha:
**Describe the Bug**
Test case failed when we use a llvm which enable LLVM_ENABLE_ASSERTIONS option.

```shell
cargo test --all --no-default-features --features "llvm7-0" --verbose
       Fresh lazy_static v1.3.0
       Fresh ucd-util v0.1.3
       Fresh semver-parser v0.7.0
       Fresh unicode-xid v0.1.0
       Fresh utf8-ranges v1.0.2
       Fresh cc v1.0.34
       Fresh unicode-xid v0.0.4
       Fresh quote v0.3.15
       Fresh either v1.5.2
       Fresh thread_local v0.3.6
       Fresh regex-syntax v0.6.6
       Fresh semver v0.9.0
       Fresh synom v0.11.3
       Fresh memchr v2.2.0
       Fresh proc-macro2 v0.4.27
       Fresh libc v0.2.51
       Fresh syn v0.11.11
       Fresh aho-corasick v0.7.3
       Fresh quote v0.6.11
       Fresh enum-methods v0.0.8
       Fresh regex v1.1.5
       Fresh syn v0.15.30
       Fresh inkwell_internal_macros v0.1.0 (/work/project/inkwell/internal_macros)
       Fresh llvm-sys v70.1.0
       Fresh inkwell v0.1.0 (/work/project/inkwell)
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `/work/project/inkwell/target/debug/deps/inkwell-710be66657a29f6e`

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running `/work/project/inkwell/target/debug/deps/all-feb08d9805fc64a7`

running 91 tests
all-feb08d9805fc64a7: /work/tool/llvm/source/70/llvm-7.0.1/lib/IR/AttributeImpl.h:129: llvm::IntAttributeImpl::IntAttributeImpl(llvm::Attribute::AttrKind, uint64_t): Assertion `(Kind == Attribute::Alignment || Kind == Attribute::StackAlignment || Kind == Attribute::Dereferenceable || Kind == Attribute::DereferenceableOrNull || Kind == Attribute::AllocSize) && "Wrong kind for int attribute!"' failed.
[...]

PR #96, opened 2019-07-03
fix misuse of LLVM API found through inkwell tests

nlewycky:
Running inkwell's tests with against an asserts-enabled expensive-checks-enabled, address-sanitizer undefined-behaviour-sanitizer LLVM found some problems. The ones I have fixes for are collected in this branch, the more complicated ones I filed (or will file) bugs for.

This is not yet enough to make inkwell tests run cleanly on my test system, this contains only the parts I think are clean enough to be committed. Most of the changes are to the tests rather than deeper changes that make inkwell safe. I want to work on that too, but I'm going to start with just getting to green (tests passing).

## Description

I've marked the changes where I only fixed the test but where inkwell still allows a consumer to trigger the problem.

 * const_null() is removed. See \<Breaking Changes\>.
 * the element type of a vector may only be an integer, floating point, or pointer.  [test only]
 * void type can not be used as an aggregate in a pointer, structure, array, etc.  [test only]
 * LLVMConstIntGet[SZ]ExtValue are convenience functions to access the value of your constant integer when it is 64 bits or less. Return None for larger integers.
 * test_enum_attributes() is removed, it was testing creation and use of custom kinds of enum attributes. As far as I can tell, LLVM does not intend to allow library users to create their own kinds of attributes, and it sometimes asserts on it.  [test only]
 * don't build a bitcast from f32 to i64. LLVM will detect that this bitcast can't be valid and assert. (In general, LLVM tries hard to assert as early as possible with the intention that this will help developers find bugs.)   [test only]
 * 12 is not a valid alignment as it is not a power of 2. Replaced with 16.  [test only]
 * fix 8-bit literals in right shift example in doc string vs. rust nightly.

## Related Issue

Related to issues #73 and #80.

This is part of what I mentioned was coming in PR #92.

## How This Has Been Tested

`cargo +nightly test --features=llvm7-0` but also I did a build of LLVM 7.0.1 with ASan and UBSan and added RUSTFLAGS="-Z sanitizer=address" to the test run.

## Option\<Breaking Changes\>

I deleted the const_null() function. It does exactly what const_zero() does except when it exhibits undefined behaviour.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -834,14 +834,14 @@ impl Builder {
     /// #![allow(overflowing_literals)]
     ///
     /// // Logical Right Shift
-    /// assert_eq!(0b1100_0000 >> 2, 0b0011_0000);
-    /// assert_eq!(0b0000_0010 >> 1, 0b0000_0001);
-    /// assert_eq!(0b0000_1100 >> 2, 0b0000_0011);
+    /// assert_eq!(0b1100_0000u8 >> 2, 0b0011_0000);
+    /// assert_eq!(0b0000_0010u8 >> 1, 0b0000_0001);
+    /// assert_eq!(0b0000_1100u8 >> 2, 0b0000_0011);
     ///
     /// // Sign Extended Right Shift
     /// assert_eq!(0b0100_0000i8 >> 2, 0b0001_0000);
-    /// assert_eq!(0b1110_0000i8 >> 1, 0b1111_0000);
-    /// assert_eq!(0b1100_0000i8 >> 2, 0b1111_0000);
+    /// assert_eq!(0b1110_0000u8 as i8 >> 1, 0b1111_0000u8 as i8);
+    /// assert_eq!(0b1100_0000u8 as i8 >> 2, 0b1111_0000u8 as i8);
     /// ```
     ///
     /// In Rust, functions that could do this for 8bit values look like:
--- a/src/values/int_value.rs
+++ b/src/values/int_value.rs
@@ -420,6 +420,9 @@ impl IntValue {
         if !self.is_const() {
             return None;
         }
+        if self.get_type().get_bit_width() > 64 {
+            return None;
+        }
 
         unsafe {
             Some(LLVMConstIntGetZExtValue(self.as_value_ref()))
@@ -442,6 +445,9 @@ impl IntValue {
     pub fn get_sign_extended_constant(&self) -> Option<i64> {
         // Garbage values are produced on non constant values
         if !self.is_const() {
+            return None;
+        }
+        if self.get_type().get_bit_width() > 64 {
             return None;
         }
 
--- a/tests/all/test_builder.rs
+++ b/tests/all/test_builder.rs
@@ -661,6 +661,7 @@ fn test_bitcast() {
     let void_type = context.void_type();
     let f32_type = context.f32_type();
     let i32_type = context.i32_type();
+    let f64_type = context.f64_type();
     let i64_type = context.i64_type();
     let i32_ptr_type = i32_type.ptr_type(AddressSpace::Generic);
     let i64_ptr_type = i64_type.ptr_type(AddressSpace::Generic);
@@ -670,6 +671,7 @@ fn test_bitcast() {
         f32_type.into(),
         i32_vec_type.into(),
         i32_ptr_type.into(),
+        f64_type.into(),
     ];
     let fn_type = void_type.fn_type(&arg_types, false);
     let fn_value = module.add_function("bc", fn_type, None);
@@ -679,6 +681,7 @@ fn test_bitcast() {
     let f32_arg = fn_value.get_nth_param(1).unwrap();
     let i32_vec_arg = fn_value.get_nth_param(2).unwrap();
     let i32_ptr_arg = fn_value.get_nth_param(3).unwrap();
+    let f64_arg = fn_value.get_nth_param(4).unwrap();
 
     builder.position_at_end(&entry);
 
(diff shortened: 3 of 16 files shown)


TheDan64:
Could you explain this some more:
> I deleted the const_null() function. It does exactly what const_zero() does except when it exhibits undefined behavior.

Why/when is it UB? Why isn't `const_zero`?

TheDan64:
> don't build a bitcast from f32 to i64. LLVM will detect that this bitcast can't be valid and assert. (In general, LLVM tries hard to assert as early as possible with the intention that this will help developers find bugs.) [test only]

I wonder if the C api has a way for us to compare sizes ourselves...

> 12 is not a valid alignment as it is not a power of 2. Replaced with 16. [test only]

We should probably assert that `set_param_alignment_attribute`'s arg is a power of two (ie `num::count_ones() == 1`?)

src/values/int_value.rs
```diff
@@ -444,6 +447,9 @@ impl IntValue {
         if !self.is_const() {
             return None;
         }
+        if self.get_type().get_bit_width() > 64 {
```
TheDan64: > LLVMConstIntGet[SZ]ExtValue are convenience functions to access the value of your constant integer when it is 64 bits or less. Return None for larger integers.

LLVM supports 128bit ints as well as custom bit width ints. Why wouldn't this work for > 64?

src/builder.rs
```diff
@@ -834,14 +834,14 @@ impl Builder {
...
+    /// assert_eq!(0b1100_0000u8 >> 2, 0b0011_0000);
+    /// assert_eq!(0b0000_0010u8 >> 1, 0b0000_0001);
+    /// assert_eq!(0b0000_1100u8 >> 2, 0b0000_0011);
     ///
     /// // Sign Extended Right Shift
     /// assert_eq!(0b0100_0000i8 >> 2, 0b0001_0000);
-    /// assert_eq!(0b1110_0000i8 >> 1, 0b1111_0000);
-    /// assert_eq!(0b1100_0000i8 >> 2, 0b1111_0000);
+    /// assert_eq!(0b1110_0000u8 as i8 >> 1, 0b1111_0000u8 as i8);
+    /// assert_eq!(0b1100_0000u8 as i8 >> 2, 0b1111_0000u8 as i8);
```
TheDan64: Hmm, how come the cast is necessary?

tests/all/test_builder.rs
```diff
@@ -695,7 +698,7 @@ fn test_bitcast() {
     let first_iv = cast.as_instruction_value().unwrap();
 
     builder.position_before(&first_iv);
-    builder.build_bitcast(f32_arg, i64_type, "f32toi64");
+    builder.build_bitcast(f64_arg, i64_type, "f64toi64");
 
-    assert!(module.verify().is_err());
+    assert!(module.verify().is_ok());
```
TheDan64: I think the point of this was to test that LLVM will infact declare the module invalid if you try to bitcast `f32` to `i64`. So long as it isn't UB, what's the issue with this?

src/values/int_value.rs
nlewycky: > > LLVMConstIntGet[SZ]ExtValue are convenience functions to access the value of your constant integer when it is 64 bits or less. Return None for larger integers.
> 
> LLVM supports 128bit ints as well as custom bit width ints. Why wouldn't this work for > 64?

On common platforms the largest type available in C and C++ is a 64-bit type, so LLVM provided a convenience function that returns a `uin64_t`. (This is akin to the fact that llvm supports 256-bit integers but rust only has a `u128`.) If the value is larger, LLVMConstIntGet[SZ]ExtValue will saturate. In C++ you can access the larger value with ConstantInt::getValue(), but there doesn't appear to be any way at all to get it through the C API.

src/builder.rs
nlewycky: Looks like a change in rust nightly?

tests/all/test_builder.rs
nlewycky: An asserts build of LLVM will assert and terminate the program.

It's true that creating the bitcast instruction won't trigger UB internally to LLVM, but there is no way to ever convert such a bitcast into code that in valid. The type of a value is immutable assigned only upon creation, so neither the target type of the bitcast nor the type of its operand can change. LLVM has a similar assertion for RAUW between different types too.

nlewycky:
> Could you explain this some more:
> 
> > I deleted the const_null() function. It does exactly what const_zero() does except when it exhibits undefined behavior.
> 
> Why/when is it UB? Why isn't `const_zero`?

It casts the Type* it's given to a PointerType*. If the target memory is not a PointerType, that's UB on the spot. In an asserts build of llvm, it's a checked cast with an assertion failure. In non-asserts build of llvm, it's just an unchecked cast to an object of the wrong type, then it starts dereferencing members that don't exist on PointerType.

> > don't build a bitcast from f32 to i64. LLVM will detect that this bitcast can't be valid and assert. (In general, LLVM tries hard to assert as early as possible with the intention that this will help developers find bugs.) [test only]
> 
> I wonder if the C api has a way for us to compare sizes ourselves...
> 
> > 12 is not a valid alignment as it is not a power of 2. Replaced with 16. [test only]
> 
> We should probably assert that `set_param_alignment_attribute`'s arg is a power of two (ie `num::count_ones() == 1`?) We should also mention this in the method's docs.

There's LLVMGetIntTypeWidth but only if the type is an IntegerType, LLVMGetVectorSize for VectorType, there's LLVMSizeOf which produces a Value that can be used at runtime (but we could check if it's a constant int and pull its value?). Ideally LLVM's CastInst::castIsValid would be exposed through the C API.

Issue #98, opened 2019-07-08
int_type.const_int_from_string radix can not be an arbitrary number


nlewycky:
**Describe the Bug**
LLVM will assert if the radix in LLVMConstIntOfStringAndSize (in LLVM C++ API, this is APInt c'tor) is not 2, 8, 10, 16 or 36.

Since that's a small finite list, we could use an enum to represent the valid values.

**To Reproduce**
The surprisingly thorough tests will already demonstrate the issue. tests/all/test_values.rs test_value_from_string() has this code:

```
    let i8_val = i8_type.const_int_from_string("0121", 3);

    assert_eq!(*i8_val.print_to_string(), *CString::new("i8 16").unwrap());
```

so simply executing the tests with an assert llvm build will fail:

```
$ cargo test --all --features=llvm7-0 -- test_value_from_string
   Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running target/debug/deps/inkwell-3860d788201e18fe

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running target/debug/deps/all-4e4d32e48e3ec760

running 1 test
all-4e4d32e48e3ec760: /home/nicholas/llvm-7.0.1.src/lib/Support/APInt.cpp:1968: void llvm::APInt::fromString(unsigned int, llvm::StringRef, uint8_t): Assertion `(radix == 10 || radix == 8 || radix == 16 || radix == 2 || radix == 36) && "Radix should be 2, 8, 10, 16, or 36!"' failed.
error: process didn't exit successfully: `/home/nicholas/inkwell2/target/debug/deps/all-4e4d32e48e3ec760 test_value_from_string` (signal: 6, SIGABRT: process abort signal)
```

**Expected Behavior**
Not able to compile code which makes LLVM assert or UB outside unsafe block.

**LLVM Version (please complete the following information):**
 - LLVM Version: 7.0.1
 - Inkwell Branch Used: master

**Desktop (please complete the following information):**
 - OS: Debian unstable

**Additional Context**
As an alternative, we could have a "string that is known to be valid in radix X" type, which has runtime-checked construction from string, and an API that guarantees valid string manipulation (you may only assign 0-7 to a character in the string, you may subset it, you may concatenate it with another of its radix or a lesser radix, that kind of thing).

TheDan64:
[...]

PR #99, opened 2019-07-08
Improve safety of const_int_from_string function by only allowing valid radixes.

nlewycky:
## Description

Improve safety of const_int_from_string function by only allowing valid radixes.

const_int_from_string() is still unsafe as the string may contain values outside the range permitted by the radix.

## Related Issue

Issue #98.

## How This Has Been Tested

```
$ cargo test --all --features=llvm7-0 -- test_value_from_string
```

On a Debian Unstable system, testing against my own build of llvm 7.0.1 with assertions enabled. I checked that the assertion fires without this patch.

## Option\<Breaking Changes\>

Callers of `int_type.const_int_from_string` will need to update to use the new enum instead of an integer literal. They can use `StringRadix::from_u8(value).unwrap()`.

## Checklist

### DIFF
--- a/src/types/int_type.rs
+++ b/src/types/int_type.rs
@@ -258,15 +258,15 @@ impl IntType {
     ///
     /// assert_eq!(i8_val.print_to_string().to_string(), "i8 121");
     ///
-    /// let i8_val = i8_type.const_int_from_string("0121", StringRadix::from_u8(10).unwrap());
+    /// let i8_val = i8_type.const_int_from_string("0121", StringRadix::from_u8(10).unwrap()).unwrap();
     ///
     /// assert_eq!(i8_val.print_to_string().to_string(), "i8 16");
     ///
-    /// // Unsafe executions, LLVM may terminate with an assertion failure or
-    /// // execute undefined behaviour in C++ code.
     /// let i8_val = i8_type.const_int_from_string("0121", StringRadix::Binary);
+    /// assert!(i8_val.is_none());
     ///
     /// let i8_val = i8_type.const_int_from_string("ABCD", StringRadix::Binary);
+    /// assert!(i8_val.is_none());
     /// ```
     pub fn const_int_from_string(&self, slice: &str, radix: StringRadix) -> Option<IntValue> {
         if !radix.to_regex().is_match(slice) {
--- a/src/types/int_type.rs
+++ b/src/types/int_type.rs
@@ -9,6 +9,34 @@ use crate::types::traits::AsTypeRef;
 use crate::types::{Type, ArrayType, BasicTypeEnum, VectorType, PointerType, FunctionType};
 use crate::values::{AsValueRef, ArrayValue, GenericValue, IntValue};
 
+/// How to interpret a string or digits used to construct an integer constant.
+#[derive(Clone, Copy, Debug, EnumAsGetters, EnumIntoGetters, EnumToGetters, Eq, Hash, PartialEq)]
+pub enum StringRadix {
+    /// Binary 0 or 1
+    Binary = 2,
+    /// Octal 0-7
+    Octal = 8,
+    /// Decimal 0-9
+    Decimal = 10,
+    /// Hexadecimal with upper or lowercase letters up to F.
+    Hexadecimal = 16,
+    /// Alphanumeric, 0-9 and all 26 letters in upper or lowercase.
+    Alphanumeric = 36,
+}
+impl StringRadix {
+    /// Create a radix from a base.
+    pub fn from_u8(value: u8) -> Option<StringRadix> {
+        match value {
+            2 => Some(StringRadix::Binary),
+            8 => Some(StringRadix::Octal),
+            10 => Some(StringRadix::Decimal),
+            16 => Some(StringRadix::Hexadecimal),
+            36 => Some(StringRadix::Alphanumeric),
+            _ => None
+        }
+   }
+}
+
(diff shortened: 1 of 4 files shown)


src/types/int_type.rs
```diff
@@ -9,6 +9,34 @@ use crate::types::traits::AsTypeRef;
 use crate::types::{Type, ArrayType, BasicTypeEnum, VectorType, PointerType, FunctionType};
 use crate::values::{AsValueRef, ArrayValue, GenericValue, IntValue};
 
+/// How to interpret a string or digits used to construct an integer constant.
+#[derive(Clone, Copy, Debug, EnumAsGetters, EnumIntoGetters, EnumToGetters, Eq, Hash, PartialEq)]
+pub enum StringRadix {
```
TheDan64: Great idea, thanks!

src/types/int_type.rs
```diff
@@ -209,29 +237,27 @@ impl IntType {
...
     /// assert_eq!(i8_val.print_to_string().to_string(), "i8 16");
     ///
-    /// // Unexpected outputs:
-    /// let i8_val = i8_type.const_int_from_string("0121", 2);
-    ///
-    /// assert_eq!(i8_val.print_to_string().to_string(), "i8 3");
-    ///
-    /// let i8_val = i8_type.const_int_from_string("ABCD", 2);
+    /// // Unsafe executions, LLVM may terminate with an assertion failure or
+    /// // execute undefined behaviour in C++ code.
```
TheDan64: I'm not sure I follow as to how this can be UB? If so we should make the function unsafe?

src/types/int_type.rs
```diff
@@ -209,29 +237,27 @@ impl IntType {
...
+    /// let i8_val = i8_type.const_int_from_string("0121", StringRadix::Binary);
     ///
-    /// assert_eq!(i8_val.print_to_string().to_string(), "i8 -15");
+    /// let i8_val = i8_type.const_int_from_string("ABCD", StringRadix::Binary);
     /// ```
-    pub fn const_int_from_string(&self, slice: &str, radix: u8) -> IntValue {
+    pub fn const_int_from_string(&self, slice: &str, radix: StringRadix) -> IntValue {
         let value = unsafe {
-            LLVMConstIntOfStringAndSize(self.as_type_ref(), slice.as_ptr() as *const i8, slice.len() as u32, radix)
+            LLVMConstIntOfStringAndSize(self.as_type_ref(), slice.as_ptr() as *const i8, slice.len() as u32, radix as u8)
```
TheDan64: If you make the enum `#[repr(u8)]` then I think you can avoid the cast

src/types/int_type.rs
nlewycky: LLVM's API is pretty crazy -- the caller is responsible for making sure that all inputs follow all of llvm's rules (and the documentation is not always there). If you don't, LLVM may do absolutely anything including UB. So, from llvm's point of view, asking it to parse "3.asd" in base 10, llvm is permitted to perform UB since you shouldn't be asking it that. (Their point of view is that their job is to provide maximum performance for customers who want it, and even the act of checking for validity takes time which is an expense upon those who don't need it.)

As for how this particular function call can trigger UB, here's a snippet of C++ code in APInt::fromString:
```
 void APInt::fromString(unsigned numbits, StringRef str, uint8_t radix) {
   // Check our assumptions here
   assert(!str.empty() && "Invalid string length");
   assert((radix == 10 || radix == 8 || radix == 16 || radix == 2 ||
           radix == 36) &&
          "Radix should be 2, 8, 10, 16, or 36!");
 
   StringRef::iterator p = str.begin();
   size_t slen = str.size();
   bool isNeg = *p == '-';
```
StringRef is not NUL terminated, so an empty string means there are zero bytes that are safe to dereference. In an asserts build of LLVM, it will assert on an empty string, but on a non-asserts build it will dereference a pointer to no particular place in `bool isNeg = *p == '-';`.

Should we make it unsafe? I was wondering what your goal for inkwell is. Given llvm's API design, it seems like nearly every function would end up being unsafe? I'd be interested in trying to figure out how to make them safe?

TheDan64: Well the function being safe is a matter of maintaining LLVM's invariants. So I think the best thing to do is either A) change the return type to be a `Result` so that the function is fallible (very verbose) or B) just assert that the string length is non zero, and document that the function call can panic if an empty string. In practice this shouldn't come up anyway, so I think this is the better of the two

nlewycky: It's not just the empty string, I think the best way is to match it against a regex verifying that only the correct digits are supplied for the given radix. A panic!() is fine with me.

TheDan64: Sounds good

nlewycky: Done.

src/types/int_type.rs
nlewycky: I added `#[repr(u8)]` above `pub enum StringRadix {` and got an error about the missing cast from StringRadix to u8 from rustc.

TheDan64: Weird, I guess I'll take a look after merging

TheDan64:
Looks like some tests broke?

nlewycky:
> Looks like some tests broke?

Yep, I didn't run doc-tests. Should be fixed!

nlewycky:
I think the travis CI failures are not due to this PR.

Issue #100, opened 2019-07-21
Add .ptr_type to AnyType trait


jyn514:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I want to create a pointer to a type and I have to match every member of the enum individually.

**Describe the solution you'd like**
Every variant of `AnyTypeEnum` has a `ptr_type` function. `ptr_type` should be part of the `AnyType` trait so I don't have to pattern match every time. This is similar to how `fn_type` is provided for all implementors of `BasicType`.

**Describe possible drawbacks to your solution**
No drawbacks I can think of.

**Describe alternatives you've considered**
Pattern match on every member of the enum and call `matched.ptr_type()` explicitly.

jyn514:
Here's the code for doing this without the trait:

```
macro_rules! gen_ptrs {
    ( $_self: ident, $addr: expr, $( $id: ident ),* ) => {
        match $_self {
            $( $id(t) => t.ptr_type($addr), )*
        }
    }
}

trait ToPointerType {
    fn ptr_type(&self, address_space: AddressSpace) -> types::PointerType;
}

impl ToPointerType for AnyTypeEnum {
    fn ptr_type(&self, addr: AddressSpace) -> types::PointerType {
        use AnyTypeEnum::*;
        gen_ptrs!(
            self,
            addr,
            VoidType,
            PointerType,
            FloatType,
            IntType,
            FunctionType,
            StructType,
            VectorType,
            ArrayType
        )
    }
}
```

jyn514:
Similar request for adding `.array_type` to `BasicType`

TheDan64:
> Every variant of AnyTypeEnum has a ptr_type function. 

As discovered in #96, you shouldn't be able create void pointers. So at best this function would have to be fallible.

> Here's the code for doing this without the trait: `...`

There's no need for this to be its own trait, it should just go on the `AnyType` trait. And a simple match would suffice over a macro since the macro won't be reused anywhere else

jyn514:
> As discovered in #96, you shouldn't be able create void pointers. So at best this function would have to be fallible.

That's fine, `.try_ptr_type` would still be convenient and useful.

> There's no need for this to be its own trait, it should just go on the AnyType trait.
I don't think there's a way to add functions to a trait in a different crate.

> And a simple match would suffice over a macro since the macro won't be reused anywhere else
The macro is because there's a lot of repetitive typing I didn't feel like doing. I also reused it for `array_type` with modifications.

---

[...]

PR #101, opened 2019-07-21
Add ptr_type and array_type to BasicType

jyn514:
## Description
Add `ptr_type` and `array_type` functions to BasicType. This allows users of the library to use `BasicTypeEnum`s to create arrays and pointers instead of having to match on every element.

Implementation note:
Since VoidType can't have a pointer to it, I didn't think ptr_type made sense to
implement for AnyType. array_type isn't valid for FunctionType and
VoidType in any case.

## Related Issue
https://github.com/TheDan64/inkwell/issues/100

## How This Has Been Tested
I ran `cargo test` and all new and existing tests passed.

## Checklist
- [X] My code follows the code style of this project (rustfmt and clippy have no complaints)
- [x] I have added documentation and doc tests to any new functions or types
- [X] I have updated documentation and doc tests to any modified functions or types as applicable
- [x] I have added tests to cover my changes
- [X] All new and existing tests passed
- [X] I am basing my changes off master, instead of one of the llvm version branches

### DIFF
--- a/src/types/traits.rs
+++ b/src/types/traits.rs
@@ -38,17 +38,52 @@ pub trait BasicType: AnyType {
         BasicTypeEnum::new(self.as_type_ref())
     }
 
-    /// Create a function type from this `BasicType`.
+    /// Create a `FunctionType` with this `BasicType` as its return type.
+    ///
+    /// Example:
+    /// ```
+    /// use inkwell::context::Context;
+    /// use inkwell::types::BasicType;
+    ///
+    /// let context = Context::create();
+    /// let int = context.i32_type();
+    /// let int_basic_type = int.as_basic_type_enum();
+    /// assert_eq!(int_basic_type.fn_type(&[], false), int.fn_type(&[], false));
+    /// ```
     fn fn_type(&self, param_types: &[BasicTypeEnum], is_var_args: bool) -> FunctionType {
         Type::new(self.as_type_ref()).fn_type(param_types, is_var_args)
     }
 
-    /// Create an array type from this `BasicType`.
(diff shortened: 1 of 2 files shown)


jyn514:
Where should the tests for `BasicType` go? I don't see any in `tests/all`.

TheDan64:
`test_types.rs`

jyn514:
Hmm that segfault on the first test run is strange. I reproduced it locally by changing `llvm-sys = "70.0"` to `llvm-sys = "70"` in `Cargo.toml`, but only first time I ran the tests! Maybe there's some sort of race condition? Cargo runs tests in parallel by default, I know you've mentioned LLVM doesn't deal well with threading.

TheDan64:
Yes, it's a spurious bug that we haven't totally figured out yet. I can rerun the test if needed.

jyn514:
Found what's causing it at least: https://github.com/TheDan64/inkwell/blob/master/tests/all/test_module.rs#L428

To replicate, put an `eprintln!("before")` and `eprintln!("after")`, and then this:
```
rm -f target/debug/all*  # remove cached test binaries
cargo test  # let cargo compile the test binary
rm -f target/debug/all-*.d  # not sure what this file is but it messes up the glob expansion
sh -c 'for i in `seq 1 100`; do target/debug/all-* --test-threads=1 --nocapture || exit 1; done'  
```
'before' will print but not 'after'. Strangely if you just run that one test it works fine, maybe it's something to do with uninitialized memory on the stack?

It turns out it's the doctests that take most of the time so the for loop doesn't take too long.

jyn514:
This is ready to go!

TheDan64:
Thanks for the PR!

Issue #91, opened 2019-07-01
test calls LLVMInitializeFunctionPassManager on object that isn't a FunctionPassManager


nlewycky:
**To Reproduce**
```
$ target/debug/deps/all-a7768349d1d36c50 test_passes::test_init_all_passes_for_module

running 1 test
test test_passes::test_init_all_passes_for_module ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 94 filtered out

$ valgrind --exit-on-first-error=yes --error-exitcode=1 target/debug/deps/all-a7768349d1d36c50 test_passes::test_init_all_passes_for_module
==3741== Memcheck, a memory error detector
==3741== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3741== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==3741== Command: target/debug/deps/all-a7768349d1d36c50 test_passes::test_init_all_passes_for_module
==3741== 

running 1 test
==3741== Thread 2 test_passes::tes:
==3741== Invalid read of size 8
==3741==    at 0xF689E4: llvm::legacy::FunctionPassManager::doInitialization() (in /home/nicholas/inkwell/target/debug/deps/all-a7768349d1d36c50)
==3741==    by 0xF0A695: LLVMInitializeFunctionPassManager (in /home/nicholas/inkwell/target/debug/deps/all-a7768349d1d36c50)
==3741==    by 0x3586A6: <inkwell::passes::PassManager<T>>::initialize (passes.rs:227)
==3741==    by 0x30A1D9: all::test_passes::test_init_all_passes_for_module (test_passes.rs:85)
==3741==    by 0x3273D9: all::test_passes::test_init_all_passes_for_module::{{closure}} (test_passes.rs:8)
==3741==    by 0x3286FD: core::ops::function::FnOnce::call_once (function.rs:231)
==3741==    by 0x39550E: <F as alloc::boxed::FnBox<A>>::call_box (in /home/nicholas/inkwell/target/debug/deps/all-a7768349d1d36c50)
==3741==    by 0x13A2289: __rust_maybe_catch_panic (in /home/nicholas/inkwell/target/debug/deps/all-a7768349d1d36c50)
[...]

PR #104, opened 2019-07-26
Move PassManager::{initialize,finalize} to PassManager<FunctionValue>.

nlewycky:
## Description

LLVMInitializeFunctionPassManager and LLVMFinalizeFunctionPassManager are only applicable to FunctionPassManagers, not all types of PassManager. Move those two functions to their own `impl` block so that they are only provided for that specialization.

## Related Issue

Issue #91 .

## How This Has Been Tested

```
$ /home/nicholas/inkwell/target/debug/deps/all-3f9595a010399385 --nocapture --test-threads 1 test_passes

running 3 tests
test test_passes::test_init_all_passes_for_module ... ok
test test_passes::test_pass_manager_builder ... ok
test test_passes::test_pass_registry ... ok

test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 92 filtered out
```

## Option\<Breaking Changes\>

Two methods are removed from `PassManager<T>` and are only available on `PassManager<FunctionValue>`. This may cause a build failure for callers of those methods on `PassManager<Module>`. Such callers should probably simply remove the calls to `initialize()` and `finalize()`, or if all their passes are function passes, they can create a function pass manager via `PassManager::create(&module)`.

## Checklist

### DIFF
--- a/src/passes.rs
+++ b/src/passes.rs
@@ -209,6 +209,21 @@ pub struct PassManager<T> {
     sub_type: PhantomData<T>,
 }
 
+impl PassManager<FunctionValue> {
+    // return true means some pass modified the module, not an error occurred
+    pub fn initialize(&self) -> bool {
+        unsafe {
+            LLVMInitializeFunctionPassManager(self.pass_manager) == 1
+        }
+    }
+
+    pub fn finalize(&self) -> bool {
+        unsafe {
+            LLVMFinalizeFunctionPassManager(self.pass_manager) == 1
+        }
+    }
+}
+
 impl<T: PassManagerSubType> PassManager<T> {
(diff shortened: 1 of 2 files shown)


nlewycky:
LLVM 3.9 test failure appears to be real. I'll have to look into it, but my rough guess is that initialization of those passes over that IR is indeed changing the IR.

TheDan64:
If you think it's a bug - ie all other llvm versions say one thing but 3.9 says another - then we can just have 3.9 only assert and a non 3.9 assert. I think something similar happened elsewhere

nlewycky:
I looked at Travis CI and I don't believe any of those failures are due to this PR.

PR #107, opened 2019-08-02
Remove get_entry_basic_block and make get_first_basic_block return None when llvm returns nullptr.

nlewycky:
## Description

It's an invariant in LLVM that the first basic block in a function is the entry block, therefore `get_entry_basic_block` and `get_first_basic_block` are redundant. The main difference is that `get_entry_basic_block` assumes that there is at least one child block (aka. the llvm::Function is a definition instead of a declaration) and will execute undefined behaviour if it does not, while `get_first_basic_block` returns a nullptr on function declarations.

## Related Issue

n/a

## How This Has Been Tested

Modified kaleidoscope runs, test_basic_block tests pass.

## Option\<Breaking Changes\>

Removes the `get_entry_basic_block` function, callers should use `get_first_basic_block.unwrap()` instead. The behaviour is the same except that when the unwrap fails it does so in a Rust way, instead of executing undefined behaviour inside of LLVM C++ code.

## Checklist

### DIFF
--- a/examples/kaleidoscope/main.rs
+++ b/examples/kaleidoscope/main.rs
@@ -21,7 +21,6 @@ use std::iter::Peekable;
 use std::str::Chars;
 use std::ops::DerefMut;
 
-use self::inkwell::basic_block::BasicBlock;
 use self::inkwell::builder::Builder;
 use self::inkwell::context::Context;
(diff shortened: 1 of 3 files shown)


TheDan64:
Thanks!!

PR #112, opened 2019-08-15
Add `atomicrmw` and `cmpxchg` instructions to the builder.

nlewycky:
## Description

Adds the `atomicrmw` and `cmpxchg` instructions to the builder.

Makes `!enum_replace` ordered to make AtomicOrdering comparisons easier, adds `LLVMAtomicRMWBinOp` enum.

## Related Issue

Didn't file an issue.

## How This Has Been Tested

Tested against LLVM 8.0.1 on Ubuntu x86-64 only, running cargo test filtered to only test_builder tests, and cargo test --doc.

## Checklist


src/builder.rs
```diff
@@ -1464,6 +1464,102 @@ impl Builder {
...
+    /// builder.build_atomicrmw(AtomicRMWBinOp::Add, i32_ptr_param, i32_seven, AtomicOrdering::Unordered);
+    /// builder.build_return(None);
+    /// ```
+    // https://llvm.org/docs/LangRef.html#atomicrmw-instruction
+    pub fn build_atomicrmw(&self, op: AtomicRMWBinOp, ptr: PointerValue, value: IntValue, ordering: AtomicOrdering) -> Result<InstructionValue, ()> {
+        // "The type of ‘<value>’ must be an integer type whose bit width is a power of two greater than or equal to eight and less than or equal to a target-specific size limit. The type of the ‘<pointer>’ operand must be a pointer to that type." -- https://releases.llvm.org/3.6.2/docs/LangRef.html#atomicrmw-instruction
+        // Newer LLVM's (9+) support additional FAdd and FSub operations as well as xchg on floating point types.
+        if value.get_type().get_bit_width() < 8 ||
+           !value.get_type().get_bit_width().is_power_of_two() {
+            return Err(());
```
TheDan64: Can we provide users with helpful static string error messages? Like `"Value must be a power of two"`?

nlewycky: Done!

src/lib.rs
```diff
@@ -185,6 +185,22 @@ enum_rename!{
     }
 }
 
+enum_rename!{
```
TheDan64: I think we should start moving `enum_rename` uses to the more powerful (because it can apply feature cfgs) proc macro, `llvm_enum` which does this and more. See `InstructionOpcode` for example: https://github.com/TheDan64/inkwell/blob/master/src/values/instruction_value.rs#L16-L18

Once that's done, could you add doc comments to each variant and briefly describe what each one is used for?  IE

```rust
#[llvm_enum(LLVMAtomicRMWBinOp)]
pub enum AtomicRMWBinOp {
    /// Used to add two things, atomically
    Add,
```

nlewycky: Done.

PR #121, opened 2019-09-19
Add parent checking to BB methods. Fixes #93

### LINKED ISSUE #93: improve support for detached instructions, basic blocks and functions
**Describe the Bug**
LLVM's memory management model is comprised of circular intrusive linked lists. The module owns an ordered list of global values, which may be Functions, GlobalVariables or GlobalAliases, a Function is a linked list of BasicBlocks, and a BasicBlock is a linked list of Instructions. The `remove_from_parent()` method detaches a node from its linked list parent leaving node with no concept of next or prev nodes.

From test_basic_block.rs:
```
    basic_block.remove_from_function();

    assert!(basic_block.get_next_basic_block().is_none());
```
At the moment, `get_next_basic_block()` doesn't check whether the block is detached, so it attempts to get the next block which is a misuse of LLVM API.

This also goes for similar functions `get_previous_basic_block`, `move_before` (`self` may be detached but `basic_block` may not), `move_after` (same) and `prepend_basic_block`.

**Expected Behavior**
It seems clear that `get_next_basic_block()` and `get_previous_basic_block()` should return `None()`.

What should `move_before()`, `move_after()` and `prepend_basic_block()` do if asked to order something relative to a detached basic block?

 - LLVM Version: 7.0.1
 - Inkwell Branch Used: master

 - OS: macOS 10.14.5

**Additional Context**
I've only looked at detached basic blocks, but the same rules apply for detached functions, detached instructions, and detached global values.

TheDan64:
## Related Issue

#93 

## How This Has Been Tested

Tests pass

## Option\<Breaking Changes\>

Slightly changes the api to return a Result

## Checklist

### DIFF
--- a/src/basic_block.rs
+++ b/src/basic_block.rs
@@ -146,7 +146,7 @@ impl BasicBlock {
     }
 
     /// Prepends one `BasicBlock` before another.
-    /// It returns `Err(())` when either `BasicBlock` has no parent, as LLVM assumes it has a parent.
+    /// It returns `Err(())` when either `BasicBlock` has no parent, as LLVM assumes they both have parents.
     ///
     /// # Example
     /// ```no_run
@@ -183,7 +183,7 @@ impl BasicBlock {
     }
 
     /// Appends one `BasicBlock` after another.
-    /// It returns `Err(())` when either `BasicBlock` has no parent, as LLVM assumes it has a parent.
+    /// It returns `Err(())` when either `BasicBlock` has no parent, as LLVM assumes they both have parents.
     ///
     /// # Example
     /// ```no_run
@@ -355,8 +355,8 @@ impl BasicBlock {
         Some(InstructionValue::new(value))
     }
 
-    /// Removes this `BasicBlock` from its parent `FunctionValue`. Does nothing if it has no parent.
-    /// It returns `Err(())` when it has no parent to remove from, as LLVM assumes it has a parent.
+    /// Removes this `BasicBlock` from its parent `FunctionValue`.
+    /// It returns `Err(())` when it has no parent to remove from.
     ///
     /// # Example
     /// ```no_run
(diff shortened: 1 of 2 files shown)


TheDan64:
@nlewycky Could you take a look at this when you get a chance?

TheDan64:
In particular, I wasn't sure if `move_before`/`move_after` needed parental checks on both BBs or just one. I did both just to be safe based on the snippet you put in the issue.

src/basic_block.rs
```diff
@@ -146,6 +146,7 @@ impl BasicBlock {
     }
 
     /// Prepends one `BasicBlock` before another.
+    /// It returns `Err(())` when either `BasicBlock` has no parent, as LLVM assumes it has a parent.
```
nlewycky: "as LLVM assumes it has a parent." --> "as LLVM assumed they both have parents."

src/basic_block.rs
```diff
@@ -340,6 +356,7 @@ impl BasicBlock {
     }
 
     /// Removes this `BasicBlock` from its parent `FunctionValue`. Does nothing if it has no parent.
+    /// It returns `Err(())` when it has no parent to remove from, as LLVM assumes it has a parent.
```
nlewycky: I would just cut ", as LLVM assumes it has a parent" here. It makes sense that we issue an error when removing a BB from its parent, and it doesn't have a parent. No further explanation is required.

nlewycky:
> In particular, I wasn't sure if `move_before`/`move_after` needed parental checks on both BBs or just one. I did both just to be safe based on the snippet you put in the issue.

I was surprised too so I looked it up in LLVM, and indeed both blocks need to have parents for `move_before` or `move_after` to work.

TheDan64:
Thanks!!

PR #125, opened 2019-11-01
Add IntValue::is_constant_int() which is true iff is_const() and the int value is not a constant expression.

nlewycky:
## Description

Of course in LLVM you can have a value which is a Constant, and an int, but not a ConstantInt. Add `is_constant_int()` method on `IntValue`.

Also fixes the safety test on `get_zero_extended_constant` and `get_sign_extended_constant` since those both require ConstantInt, not Constant with IntegerType.

## Related Issue

n/a. Needed this in wasmer, finally wrote a test and sending in the PR.

## How This Has Been Tested

New testcase creates constant expression `i32 ptrtoint (void ()* @0 to i32)` and checks it against `is_const` and `is_constant_int`. This testcase is a bit special in that we use the builder but never create an instruction with it. It must create a constant expression instead.

## Checklist


src/values/int_value.rs
```diff
@@ -385,7 +385,10 @@ impl IntValue {
         BasicValueEnum::new(value)
     }
 
-    /// Determines whether or not an `IntValue` is a constant.
+    /// Determines whether or not an `IntValue` is an `llvm::Constant`.
+    ///
+    /// Constants includes values that are not known at compile time, for
+    /// the example the address of a function casted to an integer.
```
TheDan64: Extra `the`
```suggestion
    /// example the address of a function casted to an integer.
```

src/values/int_value.rs
```diff
@@ -402,6 +405,25 @@ impl IntValue {
...
+    /// ```no_run
+    /// use inkwell::context::Context;
+    ///
+    /// let context = Context::create();
+    /// let i64_type = context.i64_type();
+    /// let i64_val = i64_type.const_int(12, false);
+    ///
+    /// assert!(i64_val.is_constant_int());
+    /// ```
+    pub fn is_constant_int(&self) -> bool {
```
TheDan64: Could you move this function to be after the other is const fn? Just would like both to be closer together in the docs

PR #127, opened 2019-11-05
Make vec_type() return a Result so that it can issue an error on vec_type(0).

nlewycky:
## Description

Vectors of size zero trigger an assert:
```
all-90f412f7832b506c: /home/nick/llvm-8.0.1.src/lib/IR/Type.cpp:607: static llvm::VectorType* llvm::VectorType::get(llvm::Type*, unsigned int): Assertion `NumElements > 0 && "#Elements of a VectorType must be greater than 0"' failed.
```

This fixes the test to not create zero-sized vectors, and fixes the API to forbid it in the future.

## How This Has Been Tested

`cargo test`

## Option\<Breaking Changes\>

`vec_type()` now returns a `Result`, so that it can indicate success and failure. Callers will need to be updated to call `unwrap()`.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1270,7 +1270,7 @@ impl Builder {
     /// let module = context.create_module("av");
     /// let i32_type = context.i32_type();
     /// let i32_zero = i32_type.const_int(0, false);
-    /// let vec_type = i32_type.vec_type(2).unwrap();
+    /// let vec_type = i32_type.vec_type(2);
     /// let fn_type = i32_type.fn_type(&[vec_type.into()], false);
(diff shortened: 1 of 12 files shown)


TheDan64:
Looks like, according to travis, there are still broken tests? But I wonder if we should just assert that the size isn't 0 and document it to keep the API simple

nlewycky:
I think I left some local changes out of the commit, sorry. I've rewritten them, this should work now!

As for Result, that makes sense, just so long as the assert is deterministic (like a `panic!`) and not optional like a C assert which is only enabled in certain builds.

PR #128, opened 2019-11-05
Remove MetadataValue::create_*, use context explicitly

MarkMcCaskey:
Hi!

This removes the `MetadataValue::create` functions which cause LLVM to use global context.

This makes these functions not thread safe.  The solution is that users should always supply a context when creating metadata values.

## Description

Remove `MetadataValue::create_node` and `MetadataValue::create_string`

## How This Has Been Tested

`while cargo test --features "llvm8-0" test_metadata -- --nocapture ; do : ; done` and wait 5 seconds to 5 minutes; before this change there are segfaults.

## Option\<Breaking Changes\>

`MetadataValue::create_node` and `MetadataValue::create_string` are removed.  They should be removed because they're not safe in general and there's a good alternative to them.

## Checklist

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -824,7 +824,6 @@ impl Module {
     ///
     /// ```no_run
     /// use inkwell::context::Context;
-    /// use inkwell::values::MetadataValue;
     ///
     /// let context = Context::create();
     /// let module = context.create_module("my_module");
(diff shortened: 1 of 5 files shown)


TheDan64:
I don't think we should get rid of these methods. The problem is that they need to ensure the global context is locked, which I've started to implement in the `feature/context_lifetimes` branch

TheDan64:
What we should do is 1) Properly document these functions as taking the global context and 2) properly ensure they lock the global context

nlewycky:
The problem with functions that use the global context implicitly is that we can't even write tests for them safely, as tests run in parallel. Will the `feature/context_lifetime` branch, or locking the global context, fix that?

Issue #130, opened 2019-11-11
Metadata handling is broken


seanyoung:
**Describe the Bug**
cargo test sometimes fails

**To Reproduce**
Run under valgrind:
1. Run cargo test
2. Find the line that looks like "Running target/debug/deps/all-9eb5b5ba7c1d722f"
3. Run `valgrind ./target/debug/deps/all-9eb5b5ba7c1d722f`

Please provide a simple sample of code which reproduces the issue on your machine.

**Expected Behavior**

**LLVM Version (please complete the following information):**
 - LLVM Version: 8.0
 - Inkwell Branch Used: llvm8-0

**Desktop (please complete the following information):**
 - OS: Fedora 31

**Additional Context**
LLVMHasMetadata(), LLVMGetMetadata() and LLVMSetMetadata() all take Instructions as
arguments, see:
https://github.com/llvm/llvm-project/blob/release/8.x/llvm/lib/IR/Core.cpp#L851
So:
https://github.com/TheDan64/inkwell/blob/master/src/values/mod.rs#L172
Is broken.

TheDan64:
Good catch, I think the solution is to just add

```rust
if !self.is_instruction() {
    return false;
}
```

to the top of `has_metadata`. Feel free to PR this change if you wish

PR #131, opened 2019-11-12
metadata can only be attached to instructions

seanyoung:
See: https://llvm.org/docs/LangRef.html#metadata

Attempting to set metadata on other LLVMValueRef will cause undefined
behaviour in llvm, see here for example:

https://github.com/llvm/llvm-project/blob/release/8.x/llvm/lib/IR/Core.cpp#L851

See #130

### DIFF
--- a/tests/all/test_values.rs
+++ b/tests/all/test_values.rs
@@ -443,35 +443,35 @@ fn test_metadata() {
     assert_eq!(md_string.get_string_value().unwrap(), &*CString::new("lots of metadata here").unwrap());
 
     let bool_type = context.bool_type();
-    let i8_type = context.i8_type();
+    // let i8_type = context.i8_type();
     // let i16_type = context.i16_type();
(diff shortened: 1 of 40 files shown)


tests/all/test_values.rs
```diff
@@ -524,83 +519,6 @@ fn test_metadata() {
     // assert!(!vec_val.has_metadata());
     // assert!(!fn_val.has_metadata());
 
-    bool_val.set_metadata(md_string, 3);
```
TheDan64: If you get rid of this code then we don't test get/set metadata at all. Could you update it to test a couple of instructions and non-instructions, please?

seanyoung: You're totally right. In doing that I also spotted that it could never work since set_metadata et all were not implemented for InstructionValue. So I've moved it all over to InstructionValue and added some tests.

seanyoung:
This should have been based off the master branch, closing

PR #136, opened 2019-11-21
The WebAssembly target is experimental on llvm 7.0

seanyoung:
So allow initialize_webassembly() to be called, but only if the
target-webassembly feature is enabled. This should not be enabled by
default by target-all, since it is experimental.

This makes it possible to build https://github.com/hyperledger-labs/solang against llvm 7.0. I know you want merges to go into master and this merge request is not. However this change is specific to the llvm7-0 branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,6 @@ target-all = [
     "target-sparc",
     "target-bpf",
     "target-lanai",
-    "target-webassembly"
 ]
 experimental = ["static-alloc"]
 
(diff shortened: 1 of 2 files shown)


TheDan64:
If it doesn't go into master, it will be lost

TheDan64:
We currently don't have a good way to support experimental target features (and there are a lot of them) since they'd need to be added to `target-all` on a per llvm version basis. I don't think there's reason to make a special exception just for web assembly

seanyoung:
It's regrettable since it would be nice to build on debian buster without backports, but I see you point.

PR #147, opened 2019-12-30
WIP: LLVM 9.0

bitwalker:
@TheDan64 Here's my changes, after switching from my own fork of llvm-sys to the mainline fork. Everything compiles cleanly, but naturally there are merge conflicts with master, as this branch is a bit behind, albeit not by that much. It also includes things that aren't strictly required for the LLVM 9 update, but which I needed, so they ended up on this branch. Nothing convtroversial I don't think, but worth keeping an eye on just in case.

I'm not sure I'll have time to get things caught up with master this week or next, but if nobody else gets to it, I'll return to it as soon as I can. Feel free to take this branch and hack it up as needed though!


Cargo.toml
```diff
@@ -59,11 +60,12 @@ experimental = ["static-alloc"]
 either = "1.5"
 inkwell_internals = { path = "./internal_macros", version = "0.1.0" }
 libc = "0.2"
-llvm-sys = "80.1"
+llvm-sys = "90.0.0"
 once_cell = "1.2"
 parking_lot = "0.9"
 regex = "1"
 static-alloc = { version = "0.1", optional = true }
+cfg-if = "0.1"
```
TheDan64: We've been getting a lot of complaints about compilations times. I don't think we should add `cfg-if` since the stuff you use it for look like they should be able to be done easily without it.

src/context.rs
```diff
@@ -1,6 +1,9 @@
...
+use llvm_sys::core::{LLVMAppendBasicBlockInContext, LLVMContextCreate, LLVMContextDispose, LLVMCreateBuilderInContext, LLVMDoubleTypeInContext, LLVMFloatTypeInContext, LLVMFP128TypeInContext, LLVMInsertBasicBlockInContext, LLVMInt16TypeInContext, LLVMInt1TypeInContext, LLVMInt32TypeInContext, LLVMInt64TypeInContext, LLVMInt8TypeInContext, LLVMIntTypeInContext, LLVMModuleCreateWithNameInContext, LLVMStructCreateNamed, LLVMStructTypeInContext, LLVMVoidTypeInContext, LLVMHalfTypeInContext, LLVMGetGlobalContext, LLVMPPCFP128TypeInContext, LLVMConstStructInContext, LLVMGetMDKindIDInContext, LLVMX86FP80TypeInContext, LLVMConstStringInContext, LLVMContextSetDiagnosticHandler};
+#[cfg(not(any(feature = "llvm3-6", feature = "llvm3-7")))]
```
TheDan64: Duplicate cfg here

src/context.rs
```diff
@@ -91,6 +96,11 @@ impl Context {
         Context::new(context)
     }
 
+    /// Creates a `ContextRef` referencing this context, and bound to its lifetime
+    pub fn as_ref<'ctx>(&'ctx self) -> ContextRef<'ctx> {
```
TheDan64: I'm not sure that this method is necessary. Could you explain why it was added?

src/context.rs
```diff
@@ -702,7 +733,9 @@ impl Context {
     // REVIEW: Maybe more helpful to beginners to call this metadata_tuple?
     // REVIEW: Seems to be unassgned to anything
     // REVIEW: Should maybe make this take &[BasicValueEnum]?
+    #[llvm_versions(3.6..9.0)]
```
TheDan64: Both this and the added version for 9 have the exact signature. Rather than having two definitions, we should have one and put the cfgs in the body itself. Like

```rust
#[llvm_versions(3.6..9.0)]
let metadata_value = unsafe { ... };
#[llvm_versions(9.0..=latest)]
let metadata_value = unsafe { ... };
```

And same for `metadata_string`

src/module.rs
```diff
@@ -190,6 +190,60 @@ impl<'ctx> Module<'ctx> {
...
+                                                           .map(|val| val.as_type_ref())
+                                                           .collect();
+        let value = unsafe {
+            LLVMGetIntrinsicDeclaration(self.module.get(),
+                id,
+                param_types.as_mut_ptr(),
+                param_types.len()
+            )
+        };
+        // We use unwrap() here because GetIntrinsicDeclaration will fail if the parameter types
```
TheDan64: This comment doesn't make sense to me. If `GetIntrinsicDeclaration` fails, won't that mean that `value` is null? If so then we should probably return a `Result` for the error case rather than unwrapping on the null check.

src/object_file.rs
```diff
@@ -1,322 +1,28 @@
...
-
-    pub fn get_address(&self) -> u64 {
-        unsafe {
-            LLVMGetSymbolAddress(self.symbol)
-        }
-    }
-}
+use cfg_if::cfg_if;
+
+cfg_if! {
```
TheDan64: We should just do
```rust
#[cfg(any(feature = "llvm3-6", feature = "llvm3-7", feature = "llvm3-8", feature = "llvm3-9", feature = "llvm4-0", feature = "llvm5-0", feature = "llvm6-0", feature = "llvm7-0", feature = "llvm8-0"))]
mod object_file;
#[llvm_versions(3.6..=8.0)]
pub use self::object_file::{Binary, BinaryRef, ObjectFile};
#[cfg(not(any(feature = "llvm3-6", feature = "llvm3-7", feature = "llvm3-8", feature = "llvm3-9", feature = "llvm4-0", feature = "llvm5-0", feature = "llvm6-0", feature = "llvm7-0", feature = "llvm8-0")))]
mod binary;
#[llvm_versions(9.0..=latest)]
pub use self::binary::{Binary, BinaryRef, BinaryType};
```

Could you explain this big file change a bit? Link to related docs maybe? Both? I don't have any context for why this is necessary

src/object_file/relocation.rs
```diff
@@ -0,0 +1,111 @@
...
+        };
+        Symbol::new(symbol)
+    }
+}
+
+#[derive(Debug)]
+pub struct RelocationIterator<'a> {
+    iter: LLVMRelocationIteratorRef,
+    section: LLVMSectionIteratorRef,
+    _marker: PhantomData<&'a Section<'a>>,
```
TheDan64: This doesn't look right to me. Since `RelocationIterator` seems to be derived from a `Section`, shouldn't it have `section: &'s Section` as a field rather than the current `section` and `_marker`?

src/object_file/section.rs
```diff
@@ -0,0 +1,179 @@
...
+use std::iter::FusedIterator;
+
+use llvm_sys::object::LLVMSectionIteratorRef;
+
+use super::{Binary, BinaryRef, RelocationIterator};
+
+#[derive(Debug)]
+pub struct Section<'a> {
+    section: LLVMSectionIteratorRef,
+    _marker: PhantomData<&'a Binary<'a>>,
```
TheDan64: Not totally sold that there's a good reason for this to be phantom data either

src/object_file/section.rs
```diff
@@ -0,0 +1,179 @@
...
+#[derive(Debug)]
+pub struct SectionIterator<'a> {
+    binary: BinaryRef,
+    iter: LLVMSectionIteratorRef,
+    done: bool,
+    _marker: PhantomData<&'a Binary<'a>>,
+}
+impl<'a> SectionIterator<'a> {
+    #[llvm_versions(9.0..=latest)]
+    pub(super) fn new(binary: BinaryRef) -> Self {
```
TheDan64: Since both versions of `new` have the same signature we should consolidate them into one function with `cfg`s on the call site(s).

Same with the other methods below

src/types/mod.rs
```diff
@@ -19,6 +19,9 @@ mod traits;
 mod vec_type;
 #[deny(missing_docs)]
 mod void_type;
+#[cfg(not(any(feature = "llvm3-6", feature = "llvm3-7")))]
+#[deny(missing_docs)]
```
TheDan64: Thanks for documenting new module(s)!

src/values/metadata_value.rs
```diff
@@ -131,14 +133,24 @@ impl<'ctx> MetadataValue<'ctx> {
     pub fn replace_all_uses_with(&self, other: &MetadataValue<'ctx>) {
         self.metadata_value.replace_all_uses_with(other.as_value_ref())
     }
-}
 
-impl AsValueRef for MetadataValue<'_> {
-    fn as_value_ref(&self) -> LLVMValueRef {
```
TheDan64: Why was the `AsValueRef` impl removed?

src/values/metadata_value.rs
```diff
@@ -131,14 +133,24 @@ impl<'ctx> MetadataValue<'ctx> {
...
+    pub fn as_value_ref(&self) -> LLVMValueRef {
         self.metadata_value.value
     }
 }
 
+//impl AsValueRef for MetadataValue<'_> {
```
TheDan64: Unnecessary comments?

src/values/metadata_value.rs
```diff
@@ -131,14 +133,24 @@ impl<'ctx> MetadataValue<'ctx> {
...
+    //fn as_value_ref(&self) -> LLVMValueRef {
+        //self.metadata_value.value
+    //}
+//}
+
+impl<'ctx, V: 'ctx + AsValueRef> From<V> for MetadataValue<'ctx> {
```
TheDan64: If I understand correctly, this lets you turn any inkwell `AsValueRef` impl'd type into a metadata type? I don't think this is LLVM-correct.

src/values/traits.rs
```diff
@@ -14,6 +14,17 @@ pub trait AsValueRef {
     fn as_value_ref(&self) -> LLVMValueRef;
 }
 
+impl<T, V> AsValueRef for T
```
TheDan64: Hmm, what is this for?

tests/all/test_module.rs
```diff
@@ -98,6 +98,7 @@ fn test_module_owned_data_layout_disposed_safely() {
     context.create_module("test");
 }
 
+#[llvm_versions(3.6..9.0)]
 #[test]
 fn test_write_and_load_memory_buffer() {
```
TheDan64: We shouldn't need to duplicate test code here. Just add a llvm_versions/cfg to the `create_object_file` call line?

tests/all/test_passes.rs
```diff
@@ -139,7 +139,7 @@ fn test_pass_manager_builder() {
     let module2 = module.clone();
 
     // TODOC: In 3.6, 3.8, & 3.9 it returns false. Seems like a LLVM bug?
-    #[cfg(not(any(feature = "llvm3-7", feature = "llvm6-0", feature = "llvm7-0", feature = "llvm8-0")))]
+    #[cfg(not(any(feature = "llvm3-7", feature = "llvm6-0", feature = "llvm7-0", feature = "llvm8-0", feature = "llvm9-0")))]
     assert!(!module_pass_manager.run_on(&module));
     #[cfg(any(feature = "llvm3-7", feature = "llvm6-0", feature = "llvm7-0", feature = "llvm8-0"))]
```
TheDan64: I think you need to add llvm 9 here

TheDan64:
@bitwalker do you think you'll have a chance to look at this in the near future?

TheDan64:
Closing in favor of #167, though there are likely some pieces of code from here that would be worth preserving.

Issue #150, opened 2020-01-22
build_malloc segfaults with certain arguments


willtunnels:
If you call `Builder::build_malloc` with an argument for which `is_sized()` returns `false`, it will segfault. A simple solution to this is to check `is_sized()` before shelling out to `llvm-sys` and panic if it is `false`. I can submit a PR for this.

One interesting thing about `build_malloc` is that neither clang, swiftc, nor rustc call it (or their LLVM API's equivalent), preferring to build a call to libc malloc instead -- at least this is my understanding after digging through their repositories a bit. Also, based on the code `build_malloc` generates and based on the fact that a malloc intrinsic is not mentioned in the LLVM 10 reference, my impression is that it might just be little more than a convenience function provided by the LLVM API. Perhaps it would be good to give it a `deprecated` annotation or otherwise discourage its use?

TheDan64:
Rust lets you create a `Box<()>` which allocates no space, and I think, sets the pointer equal to 1 in that case because `()` is unsized. I'd guess they avoid calling malloc for that reason.

I think
1) We should see if the malloc code/docs indicate it should never be called on an unsized type
2) Return a result rather than panic

willtunnels:
1. I couldn't find any docs but `LLVMBuildMalloc` makes an internal call to `SizeOf`, which is where it panics. This means that the segfault is in line with what we should expect as it's fundamentally the same issue that forces us to check `is_sized` in `ArrayType::size_of` etc.
2. Returning a `Result` is a good idea.

PR #151, opened 2020-01-22
Build malloc fix

willtunnels:
## Description

I added checks that the types passed to `build_malloc` and `build_array_malloc` are sized before shelling out to `llvm-sys`. One annoying thing about the implementation is that it has to match on the `BasicTypeEnum` because there isn't another way (as far as I know) to check `is_sized`. There are a few functions like this in `inkwell` where QOL could be improved in the future. For now I think this is fine, though maybe I should have left a TODO there.

## Related Issue

[Fixes this issue.](https://github.com/TheDan64/inkwell/issues/150)

## How This Has Been Tested

I added two new checks to `test_allocations` which ensure `build_malloc` and `build_array_malloc` panic when handed unsized types (instead of segfaulting). I also ran the entire existing test suite and all tests pass.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -348,7 +348,7 @@ impl<'ctx> Builder<'ctx> {
     }
 
     // TODOC: Heap allocation
-    pub fn build_malloc<T: BasicType<'ctx>>(&self, ty: T, name: &str) -> PointerValue<'ctx> {
+    pub fn build_malloc<T: BasicType<'ctx>>(&self, ty: T, name: &str) -> Result<PointerValue<'ctx>, &'static str> {
         // LLVMBulidMalloc segfaults if ty is unsized
         let is_sized = match ty.as_basic_type_enum() {
             BasicTypeEnum::ArrayType(ty) => ty.is_sized(),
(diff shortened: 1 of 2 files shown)


src/builder.rs
```diff
@@ -348,25 +348,56 @@ impl<'ctx> Builder<'ctx> {
     }
 
     // TODOC: Heap allocation
-    pub fn build_malloc<T: BasicType<'ctx>>(&self, ty: T, name: &str) -> PointerValue<'ctx> {
+    pub fn build_malloc<T: BasicType<'ctx>>(&self, ty: T, name: &str) -> Result<PointerValue<'ctx>, &'static str> {
+        // LLVMBulidMalloc segfaults if ty is unsized
+        let is_sized = match ty.as_basic_type_enum() {
```
TheDan64: I think we should add an `is_sized` method to the `BasicValue` trait so that this is a little more seamless

willtunnels: That's a good idea. Since you put yourself as the assignee, I assume you plan on making the change.

TheDan64: Whoops, didn't mean to assign myself

TheDan64: In the interest of time, I can make the change though

TheDan64:
Thanks for the PR btw!

Issue #149, opened 2020-01-21
Module::set_target doesn't correctly set the target!


willtunnels:
`LLVMSetTarget` (which is called by `Module::set_target`) takes, as its argument, a target triple. This is *not* the same thing as the result of a call to inkwell's `Target::get_name`, which is the value inkwell currently passes `Module::set_target`.

If I create a `Target` with triple "x86_64-pc-linux-gnu" and then call `Module::set_target`, the resulting .ll file will be annotated with `target triple = "x86-64"` and `llc` will not be able to compile it. The correct annotation would be "x86_64-pc-linux-gnu".

TheDan64:
Good catch, we probably want `TargetMachine::get_triple()`?

TheDan64:
Or maybe even just a string since LLVM doesn't seem to care what string you give it?

willtunnels:
I think we want `TargetMachine::get_triple()` like you mentioned. My understanding is that it matters what you give LLVM in so far as giving it the correct triple can help it optimize your IR (I read this in one of the LLVM references though I can't remember exactly where off the top of my head).

The only reason I didn't submit a pull request for this is that I do not know how to fix this without a breaking change (since it seems like we need a `TargetMachine` rather than a `Target`). Though arguably code that calls this function is already broken since attempting to compile to generated IR will, as a rule, fail.

TheDan64:
Sure, I just meant that it seems to be memory safe to provide it an invalid string as far as I can tell. In any case, feel free to PR with the `TargetMachine` change. A breaking change is fine here for correctness (also we're pre-1.0)

willtunnels:
I've been looking at this and there actually is a somewhat meaningful choice to be made. In particular, what should happen with `Module::get_target`. Right now it returns a `Target`. You could argue that now it should return a `TargetMachine` but `Module::set_target` is lossy in that you're really only storing the target triple and only the target triple will be returned from `LLVMGetTarget`, which is not enough to reconstruct the `TargetMachine`. Now, you could store a ref to the machine. But I think this issue actually points to a moderately deeper problem. Somethings take an "entire" machine and somethings just take a triple and these two things are not the same (though a machine does include a triple).

This leads to solution two: just pass strings in and out of set/get target. However, I think a much better solution is to reify the notion of a triple by adding a new struct, `inkwell::targets::TargetTriple`, and adjusting functions across the API to deal in such objects.

PR #152, opened 2020-01-24
Set target fix

willtunnels:
## Description

I added a new struct, `inkwell::targets::TargetTriple` that models a target triple more explicitly in the type system. I fixed the bug where `Module::set_target` and `Module::get_target` were operating on a target rather than a target triple. I also updated all of the functions in the inkwell API which deal with target triples (I think I got all of them at least) to use the `TargetTriple` struct. Corresponding changes were made to the documentation.

`TargetTriple` is a thin wrapper around an `LLVMStringOrRaw`. I modeled `TargetTriple` after `DataLayout`, which achieves a similar aim albeit for a different part of the API. So, even though there are a good number of breaking changes here, the end result is consistent with other parts of the already existing API.

The only part of the implementation I'm not completely satisfied with is the two string copies that have to be made in `TargetTriple::create`, which is the user facing constructor. But, even though this isn't ideal, it should probably be OK because it's unlikely to be on the critical path and, on the off chance it is, things are probably still OK.

## Related Issue

[Fixes this issue.](https://github.com/TheDan64/inkwell/issues/149)

## How This Has Been Tested

There were already test cases for all (most?) of the functions I modified. I updated the existing test suite where needed to be compatible with the new API. I ran `cargo test` and everything passed.

## Breaking Changes

- `Module::set_target` takes a `&TargetTriple` rather than a `&Target`
- `Module::get_target` returns a `TargetTriple` rather than a `Target`
- `Target::create_target_machine` takes a `&TargetTriple` rather than a ` &str` to specify the target triple
- `Target::from_triple` takes a `&TargetTriple` rather than a `&str`
- `TargetMachine::get_triple` returns a `TargetTriple` rather than an `LLVMString`
- `TargetMachine::get_default_triple` returns a `TargetTriple` rather than an `LLVMString`
- `TargetMachine::normalized_target_triple` takes a `&TargetTriple` rather than an `Either<&str, &CStr>` and returns a `TargetTriple` rather than a `LLVMString`

## Checklist

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -395,12 +395,12 @@ impl<'ctx> Module<'ctx> {
             .map_err(|mut err_string| {
                 err_string.push('\0');
 
-                LLVMString::create(&err_string)
+                LLVMString::create_from_str(&err_string)
             })?;
 
         if self.owned_by_ee.borrow().is_some() {
             let string = "This module is already owned by an ExecutionEngine.\0";
-            return Err(LLVMString::create(string));
+            return Err(LLVMString::create_from_str(string));
         }
 
         let mut execution_engine = MaybeUninit::uninit();
@@ -446,12 +446,12 @@ impl<'ctx> Module<'ctx> {
             .map_err(|mut err_string| {
                 err_string.push('\0');
 
-                LLVMString::create(&err_string)
+                LLVMString::create_from_str(&err_string)
             })?;
 
         if self.owned_by_ee.borrow().is_some() {
             let string = "This module is already owned by an ExecutionEngine.\0";
-            return Err(LLVMString::create(string));
+            return Err(LLVMString::create_from_str(string));
         }
 
         let mut execution_engine = MaybeUninit::uninit();
@@ -499,12 +499,12 @@ impl<'ctx> Module<'ctx> {
             .map_err(|mut err_string| {
                 err_string.push('\0');
 
-                LLVMString::create(&err_string)
+                LLVMString::create_from_str(&err_string)
             })?;
 
         if self.owned_by_ee.borrow().is_some() {
             let string = "This module is already owned by an ExecutionEngine.\0";
-            return Err(LLVMString::create(string));
+            return Err(LLVMString::create_from_str(string));
         }
 
         let mut execution_engine = MaybeUninit::uninit();
@@ -1209,7 +1209,7 @@ impl<'ctx> Module<'ctx> {
     pub fn link_in_module(&self, other: Self) -> Result<(), LLVMString> {
         if other.owned_by_ee.borrow().is_some() {
             let string = "Cannot link a module which is already owned by an ExecutionEngine.\0";
-            return Err(LLVMString::create(string));
+            return Err(LLVMString::create_from_str(string));
         }
 
         #[cfg(any(feature = "llvm3-6", feature = "llvm3-7"))]
--- a/src/module.rs
+++ b/src/module.rs
@@ -16,7 +16,6 @@ use llvm_sys::LLVMLinkage;
 use llvm_sys::LLVMModuleFlagBehavior;
 
 use std::cell::{Cell, RefCell, Ref};
-#[llvm_versions(3.9..=latest)]
 use std::ffi::CStr;
 use std::ffi::CString;
 use std::fs::File;
--- a/src/module.rs
+++ b/src/module.rs
@@ -327,19 +327,20 @@ impl<'ctx> Module<'ctx> {
     ///
     /// ```rust,no_run
     /// use inkwell::context::Context;
-    /// use inkwell::targets::TargetTriple;
+    /// use inkwell::targets::{Target, TargetTriple};
     ///
+    /// Target::initialize_x86(&Default::default());
     /// let context = Context::create();
     /// let module = context.create_module("mod");
     /// let triple = TargetTriple::create("x86_64-pc-linux-gnu");
     ///
-    /// assert_eq!(module.get_target(), TargetTriple::create(""));
+    /// assert_eq!(module.get_target_triple(), TargetTriple::create(""));
     ///
-    /// module.set_target(&triple);
+    /// module.set_target_triple(&triple);
     ///
-    /// assert_eq!(module.get_target(), triple);
+    /// assert_eq!(module.get_target_triple(), triple);
     /// ```
-    pub fn set_target(&self, target_triple: &TargetTriple) {
+    pub fn set_target_triple(&self, target_triple: &TargetTriple) {
         unsafe {
             LLVMSetTarget(self.module.get(), target_triple.as_ptr())
         }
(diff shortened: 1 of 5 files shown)


src/targets.rs
```diff
@@ -96,6 +96,63 @@ impl Default for InitializationConfig {
...
+
+    pub(crate) fn new_borrowed(target_triple: *const ::libc::c_char) -> TargetTriple {
+        debug_assert!(!target_triple.is_null());
+
+        TargetTriple {
+            target_triple: LLVMStringOrRaw::Borrowed(target_triple),
+        }
+    }
+
+    pub fn create(target_triple: &str) -> TargetTriple {
```
TheDan64: It's not really clear to me what advantage `TargetTriple` has over just passing a string since you can just give `create` any old invalid string? It seems like the same problem as before?

willtunnels: I was thinking of this as an application of [newtype pattern](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html). The advantage is just that it can help people avoid mixing up strings in their code that represent different things.

One could theoretically add some validation in the future though I'm not actually sure if it's feasible.

TheDan64: I'm not sure that it is feasible to add validation because you can create custom triples I think? Or at least Rust lets you define triple specifications. Not sure how that works when it gets to LLVM

TheDan64: I need to look through this again and think on it some more

src/targets.rs
```diff
@@ -96,6 +96,63 @@ impl Default for InitializationConfig {
...
+        let c_string = CString::new(target_triple).expect("Conversion to CString failed unexpectedly");
```
TheDan64: Nit: This will allocate twice, I think? You should just use `create_from_str` on `target_triple`?

willtunnels: My thinking was this:
What if the `&str` does not end with a `'\0'`? In general I would not expect it to. So, I would need to append `'\0'` to the input. To append `'\0'` I would need an owned copy, which would reallocate anyway. Am I missing something here?

TheDan64: You're right, but there's gotta be a better way..

src/targets.rs
TheDan64: I think there's value in a newtype when you can guarantee that only valid `TargetTriple` are generated, which we cannot

src/module.rs
```diff
@@ -321,59 +321,56 @@ impl<'ctx> Module<'ctx> {
...
     /// ```
-    pub fn get_target(&self) -> Option<Target> {
+    pub fn get_target(&self) -> TargetTriple {
         // REVIEW: This isn't an owned LLVMString, is it? If so, need to deallocate.
         let target_str = unsafe {
             LLVMGetTarget(self.module.get())
         };
 
-        Target::from_name_raw(target_str)
+        TargetTriple::new_borrowed(target_str)
```
TheDan64: A big problem is that we're leaking a raw pointer into the wild. Really this should be `-> &TargetTriple` for borrowed triples

src/targets.rs
willtunnels: I see both perspectives here, so I'm not super inclined to argue hard for the newtype. It's my personal inclination, but you've got a point. Just let me know what you decide on this so I can finish up the PR and we can merge it.

src/module.rs
```diff
@@ -321,59 +321,56 @@ impl<'ctx> Module<'ctx> {
...
+    /// assert_eq!(module.get_target(), TargetTriple::create(""));
     ///
-    /// module.set_target(&target);
+    /// module.set_target(&triple);
     ///
-    /// assert_eq!(module.get_target().unwrap(), target);
+    /// assert_eq!(module.get_target(), triple);
     /// ```
-    pub fn set_target(&self, target: &Target) {
+    pub fn set_target(&self, target_triple: &TargetTriple) {
```
TheDan64: We should change these to `get/set_target_triple` for clarity, despite LLVM's poorly named function that implies an actual Triple object

tests/all/test_module.rs
```diff
@@ -316,27 +316,25 @@ fn test_print_to_file() {
 
 #[test]
 fn test_get_set_target() {
-    Target::initialize_x86(&Default::default());
```
TheDan64: Why was this removed? This is generally to ensure LLVM's globals are initialized properly

src/targets.rs
willtunnels: This is abstraction breaking so probably best to avoid, but if you look at the LLVM source code you can see that creating a message just does a `strdup` and freeing it just does a `free` so you could actually pretended any old owned cstring was allocated by LLVM.

tests/all/test_module.rs
willtunnels: The test runs fine without it. Getting/setting the module's target triple does not touch the `TargetRegistry`. It's probably better to put it back in though, as a defensive measure, because I'm not sure that no code path touches the `TargetRegistry` given a module on which you've set the target triple.

willtunnels: LLVM's sparse documentation makes it hard for me to tell exactly what preconditions the APIs have

src/module.rs
willtunnels: Apparently that's the name the corresponding function in the [C++ API](https://llvm.org/doxygen/classllvm_1_1Module.html#af66bad1510da8a71d20002609d3b4a25) has

src/module.rs
willtunnels: I think a similar problem may exist with DataLayout since the origin of the problem is really with LLVMStringOrRaw not having a proper lifetime on its borrow variant. Maybe the solution is to define an LLVMStr that implements ToOwned and stick it inside an actual Cow?

willtunnels: I fixed by making a copy. I can't imagine that this is a hot path anyway and doing something more efficient really increased code complexity. Similarly, there are more efficiently ways to implement `TargetTriple;:create` (maybe you could hold an enum of `LLVMString` and `CString`) but I'm not really sure they're worth it. Also, as mentioned above, there is definitely a leakage issue with `DataLayout::get_borrowed_data_layout`. I guess I will make a separate issue for this...

src/module.rs
willtunnels: I called it `get/set_triple` for consistency with the rest of the `inkwell` api.

src/module.rs
TheDan64: Sounds good. We can always optimize the copy out in the future. Please just add a TODO comment so that we don't forget.

TheDan64: And maybe an issue too

TheDan64:
It looks like something needs to be fixed for most of the older LLVM versions

willtunnels:
I have been extremely busy with school. Sorry for the delay. I went back and fixed the issue with old LLVM versions.

TheDan64:
No problem. And thanks, I'll take a look soon. Sorry about the whack-a-mole situation. I usually just rely on the travis tests to tell me which imports aren't correct and then just look up their paths in llvm-sys.

src/targets.rs
```diff
@@ -96,6 +95,53 @@ impl Default for InitializationConfig {
...
+        TargetTriple {
+            triple,
+        }
+    }
+
+    pub fn create(triple: &str) -> TargetTriple {
+        let c_string = CString::new(triple).expect("Conversion to CString failed unexpectedly");
+
+        TargetTriple {
+            triple: LLVMString::create_from_c_str(c_string.as_c_str())
```
TheDan64: You should just do `LLVMString::create_from_str(triple)` so as to not allocate with `CString`

willtunnels: Hmm, when tests started failing I remembered the reason why I did this: if the user's &str is not NUL terminated then calling LLVMString::create_from_str will fail (and in general we don't expect it to be). This is a pretty fundamental Rust and C interop problem and I'm curious about your thoughts on it. We could avoid one of the allocations by cheating because we know the implementation of `LLVMCreateMessage` and `LLVMDisposeMessage`...

TheDan64: Ack, you're right. I keep forgetting about this. lol. Let's leave it as it is then and I'll merge

Issue #145, opened 2019-12-29
Add Builder functions for LLVM C library intrinsics


willtunnels:
LLVM offers a number of intrinsics roughly corresponding to C library functions, for instance `memcpy` and `memset`. From the [LLVM reference](https://llvm.org/docs/LangRef.html): "these intrinsics allow source-language front-ends to pass information about the alignment of the pointer arguments to the code generator, providing opportunity for more efficient code generation."

It would be nice if inkwell provided access to these intrinsics.

The implementation would probably consist of a few trivial wrappers in the `Builder` for llvm-sys functions.

TheDan64:
This is already implemented in the changes mentioned in #143 I believe

TheDan64:
CC #147

rtfeldman:
> This is already implemented in the changes mentioned in #143 I believe

I looked at the branch and it looks like they aren't.

If you're open to it, I'd be happy to make a separate PR into `master` to introduce `build_memcpy` - it'd be a wrapper around [`LLVMBuildMemCpy`](https://thedan64.github.io/inkwell/llvm_sys/core/fn.LLVMBuildMemSet.html) very similar to how [`build_malloc`](https://thedan64.github.io/inkwell/inkwell/builder/struct.Builder.html#method.build_malloc) wraps [`LLVMBuildMalloc`](https://thedan64.github.io/inkwell/llvm_sys/core/fn.LLVMBuildMalloc.html).

TheDan64:
Sure, go for it!

duci9y:
While I don't feel confident making a PR myself (yet), would love to see a `build_memset` as well.

PR #158, opened 2020-03-19
Add build_memcpy and build_memmove

rtfeldman:
## Description

Add `build_memcpy` and `build_memmove` LLVM intrinsics.

## Related Issue

https://github.com/TheDan64/inkwell/issues/145#issuecomment-600399438

## How This Has Been Tested

TODO

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -357,13 +357,13 @@ impl<'ctx> Builder<'ctx> {
     ///
     /// [`TargetData::ptr_sized_int_type_in_context`](https://thedan64.github.io/inkwell/inkwell/targets/struct.TargetData.html#method.ptr_sized_int_type_in_context) will get you one of those.
     #[llvm_versions(8.0..=latest)]
-    pub fn build_memcpy<T: IntMathValue<'ctx>>(
+    pub fn build_memcpy(
         &self,
         dest: PointerValue<'ctx>,
         dest_align_bytes: u32,
         src: PointerValue<'ctx>,
         src_align_bytes: u32,
-        size: T,
+        size: IntValue<'ctx>,
     ) -> Result<PointerValue<'ctx>, &'static str> {
         let value = unsafe {
             LLVMBuildMemCpy(
@@ -385,13 +385,13 @@ impl<'ctx> Builder<'ctx> {
     ///
     /// [`TargetData::ptr_sized_int_type_in_context`](https://thedan64.github.io/inkwell/inkwell/targets/struct.TargetData.html#method.ptr_sized_int_type_in_context) will get you one of those.
     #[llvm_versions(8.0..=latest)]
-    pub fn build_memmove<T: IntMathValue<'ctx>>(
+    pub fn build_memmove(
         &self,
         dest: PointerValue<'ctx>,
         dest_align_bytes: u32,
         src: PointerValue<'ctx>,
         src_align_bytes: u32,
-        size: T,
+        size: IntValue<'ctx>,
     ) -> Result<PointerValue<'ctx>, &'static str> {
         let value = unsafe {
             LLVMBuildMemMove(
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -1,7 +1,7 @@
 //! A `Builder` enables you to build instructions.
 
 use either::{Either, Left, Right};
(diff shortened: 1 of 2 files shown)


rtfeldman:
This is WIP - I haven't written tests yet, but wanted to get feedback on the approach so far before proceeding.

Does anything look wrong other than the fact that it doesn't have tests yet?

src/builder.rs
```diff
@@ -1,7 +1,7 @@
...
...uildUnreachable, LLVMBuildXor, LLVMDisposeBuilder, LLVMGetElementType, LLVMGetInsertBlock, LLVMGetReturnType, LLVMGetTypeKind, LLVMInsertIntoBuilder, LLVMPositionBuilderAtEnd, LLVMTypeOf, LLVMBuildExtractElement, LLVMBuildInsertElement, LLVMBuildIntToPtr, LLVMBuildPtrToInt, LLVMInsertIntoBuilderWithName, LLVMClearInsertionPosition, LLVMPositionBuilder, LLVMPositionBuilderBefore, LLVMBuildAggregateRet, LLVMBuildStructGEP, LLVMBuildInBoundsGEP, LLVMBuildPtrDiff, LLVMBuildNSWAdd, LLVMBuildNUWAdd, LLVMBuildNSWSub, LLVMBuildNUWSub, LLVMBuildNSWMul, LLVMBuildNUWMul, LLVMBuildSDiv, LLVMBuildSRem, LLVMBuildURem, LLVMBuildFRem, LLVMBuildNSWNeg, LLVMBuildNUWNeg, LLVMBuildFPToUI, LLVMBuildFPToSI, LLVMBuildSIToFP, LLVMBuildUIToFP, LLVMBuildFPTrunc, LLVMBuildFPExt, LLVMBuildIntCast, LLVMBuildFPCast, LLVMBuildSExtOrBitCast, LLVMBuildZExtOrBitCast, LLVMBuildTruncOrBitCast, LLVMBuildSwitch, LLVMAddCase, LLVMBuildShl, LLVMBuildAShr, LLVMBuildLShr, LLVMBuildGlobalString, LLVMBuildGlobalStringPtr, LLVMBuildExactSDiv, LLVMBuildTrunc, LLVMBuildSExt, LLVMBuildZExt, LLVMBuildSelect, LLVMBuildAddrSpaceCast, LLVMBuildBitCast, LLVMBuildShuffleVector, LLVMBuildVAArg, LLVMBuildIndirectBr, LLVMAddDestination};
```
TheDan64: It looks like these methods are only exposed for LLVM 8+ so you'll have to add `#[llvm_versions(8.0..=latest)]` to the individual import as well as the methods.

src/builder.rs
```diff
@@ -347,6 +347,60 @@ impl<'ctx> Builder<'ctx> {
...
+    /// The final argument should be a pointer-sized integer.
+    ///
+    /// [`TargetData::ptr_sized_int_type_in_context`](https://thedan64.github.io/inkwell/inkwell/targets/struct.TargetData.html#method.ptr_sized_int_type_in_context) will get you one of those.
+    pub fn build_memcpy<T: IntMathValue<'ctx>>(
+        &self,
+        dest: PointerValue<'ctx>,
+        dest_align_bytes: u32,
+        src: PointerValue<'ctx>,
+        src_align_bytes: u32,
+        size: T,
```
TheDan64: It's not clear to me why `size` is an `IntMathValue`. Shouldn't it always just be an int(`IntValue`)? Same for memmove.

src/builder.rs
```diff
@@ -347,6 +347,60 @@ impl<'ctx> Builder<'ctx> {
         PointerValue::new(value)
     }
 
+    /// Alignment arguments are specified in bytes, and should always be a power of 2.
```
TheDan64: Please be sure to test what happens when you provide a non-power of two. It may cause the function to return a null ptr (which needs to be handled) or worse cause UB - in which case we need to validate the input and return an `Err` in such cases. (Rust ints have an `is_power_of_two` method)

src/builder.rs
rtfeldman: Seems reasonable! :+1: Changed to `IntValue`.

src/builder.rs
```diff
@@ -4,6 +4,10 @@ use either::{Either, Left, Right};
...
 #[llvm_versions(3.9..=latest)]
 use llvm_sys::core::LLVMBuildAtomicCmpXchg;
+#[llvm_versions(8.0..=latest)]
+use llvm_sys::core::LLVMBuildMemCpy;
```
TheDan64: Nit: You should combine these two into one for simplicity:
```Rust
#[llvm_versions(8.0..=latest)]
use llvm_sys::core::{LLVMBuildMemCpy, LLVMBuildMemMove};
```

src/builder.rs
```diff
@@ -347,6 +351,62 @@ impl<'ctx> Builder<'ctx> {
...
+    /// [`TargetData::ptr_sized_int_type_in_context`](https://thedan64.github.io/inkwell/inkwell/targets/struct.TargetData.html#method.ptr_sized_int_type_in_context) will get you one of those.
+    #[llvm_versions(8.0..=latest)]
+    pub fn build_memcpy(
+        &self,
+        dest: PointerValue<'ctx>,
+        dest_align_bytes: u32,
+        src: PointerValue<'ctx>,
+        src_align_bytes: u32,
+        size: IntValue<'ctx>,
+    ) -> Result<PointerValue<'ctx>, &'static str> {
```
Hywan: Correct me if I'm wrong, but this function cannot fail for the moment. Why returning `Result` then? Is it because it's a WIP 🙂 ?

src/builder.rs
```diff
@@ -347,6 +351,62 @@ impl<'ctx> Builder<'ctx> {
...
+    /// [`TargetData::ptr_sized_int_type_in_context`](https://thedan64.github.io/inkwell/inkwell/targets/struct.TargetData.html#method.ptr_sized_int_type_in_context) will get you one of those.
+    #[llvm_versions(8.0..=latest)]
+    pub fn build_memmove(
+        &self,
+        dest: PointerValue<'ctx>,
+        dest_align_bytes: u32,
+        src: PointerValue<'ctx>,
+        src_align_bytes: u32,
+        size: IntValue<'ctx>,
+    ) -> Result<PointerValue<'ctx>, &'static str> {
```
Hywan: Same comment as above, this function cannot fail for what I read for the moment.

src/builder.rs
rtfeldman: I tried this out, and the module fails verification with:

```
alignment of arg 0 of memory intrinsic must be 0 or a power of 2
  call void @llvm.memcpy.p0i8.p0i8.i64(i8* align 3 %1, i8* align 3 %2, i64 8, i1 false)
```

Given that LLVM already has a check for this with a descriptive error message, do we need a separate test for it?

src/builder.rs
rtfeldman: Done!

src/builder.rs
rtfeldman: Yeah, because it's WIP - I was thinking about having it return `Err` if either alignment value isn't a power of 2, but it turns out if you provide something that isn't a power of 2, LLVM fails module verification anyway.

Based on that, I'm thinking this (and `build_memmove`) can just return `PointerValue<'ctx>` instead of a `Result`.

What do you think?

rtfeldman:
I added a `memcpy` test and would love any feedback on it!

src/builder.rs
TheDan64: Have you also tried really large values of alignment? Looks like LLVM might assert: https://github.com/llvm/llvm-project/blob/70c0a9675dcaeb74a114a129f3d998bd6dee926f/llvm/lib/IR/Attributes.cpp#L146-L149 which would be really bad on the Rust end.

rtfeldman: * 0 and 1 both work
* 3 fails validation
* `std::u32::MAX` fails validation

This is surprising to me, considering it looks like it's [hardcoded in a .h file](https://github.com/llvm/llvm-project/blob/f16f139db40e6bf6462ca831eb1ec423c50aeef2/llvm/include/llvm/IR/Value.h#L683) to be [equal to 536870912](https://github.com/llvm/llvm-project/blob/0e62011df891d0e7ad904524edf705d07d12d5d4/llvm/unittests/IR/ValueTest.cpp#L63) and yet it accepts both 536870912 and 2 * 536870912 without complaining.

At any rate, I haven't found an alignment value where it does anything other than either accepting it or failing validation.

TheDan64: Really weird. I'm gonna give it a test myself too just to be sure. In the meantime, the tests need to be updated to only run on llvm8

TheDan64: You should also write tests that use those sample values, ie `0`, `1`, `3`, `std::u32::max_value()`. However, you don't need to make them as in-depth as the current test with an execution engine. Just create a really simple function that uses your added functions with invalid values and assert that the module fails validation because of it. That way we're more likely to catch UB - if any - at some point.

rtfeldman: Added in 1ac9cfdc7917191f721e267f59c2bb9e46e52f41

src/builder.rs
rtfeldman: Changed both in 9a263ebabd3f9f3b7d59ff001cd55d076be366fe

rtfeldman:
I've now revised `memcpy` tests based on all the great feedback!

Let me know if they look good - if so, I'll follow the same approach for `memset` tests.

TheDan64:
Sorry for the delay. This looks good. Feel free for add the `memmove` tests and then we can get this merged in!

rtfeldman:
Done and done!

TheDan64:
@rtfeldman We just merged in llvm 10 support and the memcpy seems to be failing there: https://travis-ci.org/github/TheDan64/inkwell/jobs/672346135

Any ideas?

TheDan64:
@rtfeldman We just merged in llvm 10 support and the memcpy seems to be failing there: https://travis-ci.org/github/TheDan64/inkwell/jobs/672346135

Any ideas?

Issue #159, opened 2020-03-28
can't call replace all uses with on a basic block


nlewycky:
**Is your feature request related to a problem? Please describe.**
In my frontend I sometimes create the basic blocks before I need them (in "start" and "end" pairs), then a block ends up containing nothing but a branch to another basic block (multiple "end" blocks in a stack). When that happens, I could simply RAUW the block to that destination instead.

**Describe the solution you'd like**
LLVM lets me call replaceAllUsesWith on a BasicBlock. Expose that in inkwell.

**Describe possible drawbacks to your solution**
I can't think of anything.

**Describe alternatives you've considered**
There's other ways to architect a frontend, but this is pretty simple.

**Additional context**

PR #160, opened 2020-03-28
Add an implementation of replace_all_uses_with for BasicBlock.

nlewycky:
## Description

LLVM has replaceAllUsesWith for BasicBlocks as well as other Values. Add an implementation of that method to BasicBlock.

## Related Issue

#159

## How This Has Been Tested

Added a doc-test and checked that it passes, and doesn't pass if I change the assert. Tested on llvm8-0 in a release+asserts build.

## Checklist


nlewycky:
> This looks fine, could you just copy your doc test into an actual test for the sake of coverage?

Done.

Issue #162, opened 2020-03-30
BasicValueUse::get_user always returns InstructionValue


nlewycky:
**Describe the Bug**
Value users do not always have to be Instructions.

The user of an Instruction must be an Instruction. However, the user of a Function may be a Constant (say, the LLVM IR for a C++ virtual table, a constant array holding the pointers to functions).

**To Reproduce**
This was noticed by inspection of src/values/basic_value_use.rs.

You could create two `GlobalValue`s,  `@A = external global i32` and `@B = constant i32* @A`, then get a use-list of `@B` and get its first use which will be `@A`, then call `get_user` which will call InstructionValue::new on a non-Instruction.

**Expected Behavior**
`get_user` should probably return a BasicValue, or maybe an Either?

**LLVM Version (please complete the following information):**
 - LLVM Version: llvm8-0
 - Inkwell Branch Used: master

TheDan64:
I think returning `BasicValue` makes sense since you can always turn it into an `InstructionValue` afterwards?

nlewycky:
Yes, that sounds right to me.

nlewycky:
Ah, but not a BasicValueEnum. A store instruction uses other values but always has void type, and therefore isn't a BasicValueEnum. We could use AnyValueEnum which I think always works because any void-typed value would be an InstructionValue.

PR #163, opened 2020-04-02
Change BasicValueUse::get_user to return an AnyValueEnum.

nlewycky:
## Description

Change BasicValueUse::get_user to return an AnyValueEnum.

To make the tests pass, we need to improve AnyValueEnum too. When construction an AnyValueEnum with a void Value, choose the InstructionValue enum, for example a StoreInst. Implement the into_X_value() methods so that we can use the contained value.

## Related Issue

#162 

## How This Has Been Tested

`cargo test --features=llvm8-0`

## Option\<Breaking Changes\>

The return type of the function changed, callers may have to adjust.

## Checklist

### DIFF
--- a/src/values/basic_value_use.rs
+++ b/src/values/basic_value_use.rs
@@ -5,9 +5,9 @@ use llvm_sys::prelude::LLVMUseRef;
 use std::marker::PhantomData;
 
 use crate::basic_block::BasicBlock;
-use crate::values::{BasicValueEnum, InstructionValue};
+use crate::values::{AnyValueEnum, BasicValueEnum};
 
-/// A usage of a `BasicValue` in an `InstructionValue`.
+/// A usage of a `BasicValue` in another value.
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct BasicValueUse<'ctx>(LLVMUseRef, PhantomData<&'ctx ()>);
 
@@ -18,7 +18,7 @@ impl<'ctx> BasicValueUse<'ctx> {
         BasicValueUse(use_, PhantomData)
     }
 
-    /// Gets the next use of an `InstructionValue` or `BasicValue` if any.
+    /// Gets the next use of a `BasicValue` if any.
     ///
     /// The following example,
     ///
@@ -83,7 +83,7 @@ impl<'ctx> BasicValueUse<'ctx> {
         Some(Self::new(use_))
     }
 
-    /// Gets the user(an `InstructionValue`) of this use.
+    /// Gets the user (an `AnyValueEnum`) of this use.
     ///
     /// ```no_run
     /// use inkwell::AddressSpace;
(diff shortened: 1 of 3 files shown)


TheDan64:
The thing that throws me off about Adding `InstructionValue` to `AnyValueEnum` is that other variants might also be instructions (and can be converted into one). So the variants are no longer distinct types. hmm

nlewycky:
InstructionValue was already part of the AnyValueEnum, it was just not reachable through the standard constructor. I took that as a TODO rather than a design change. 

I agree that we have rust types for llvm types, and rust types for llvm value kinds, and that this is confusing. However, I don't think this is new, when you construct a load it is both an InstructionValue and some type, such as IntValue, depending on what it loaded. Or an build_alloca always returns an instruction that has pointer-type. We just have conversions between them.

I think I've suggested this before, I think the right way is to use mixins. We should have a trait for each value kind, and for each type. "Instruction" and a more specific trait for "LoadInstruction", which adds things like get/set alignment, while IntValue needs no methods but exists for type safety of inkwell's user. Builder::build_alloca would return dyn Instruction + dyn PointerValue, while ConstInt would imply that it's both a Constant (value kind) and an Int (type) while providing additional methods (the const_X methods currently on IntValue), however I'm not sure whether this is really possible to implement. In the absence of that, I figure we can just create whatever type-specific rust type or valuekind-specific rust type is most appropriate in the situation and call an into_X or as_X method when you need to deal with its other nature.

nlewycky:
It just dawned on me. Were you suggesting that I remove InstructionValue and PhiValue from AnyValueEnum, and add VoidValue?

TheDan64:
> It just dawned on me. Were you suggesting that I remove InstructionValue and PhiValue from AnyValueEnum, and add VoidValue?

No, I've just been trying to grok the consequences of this change. But I guess you're right that `InstructionValue` and `PhiValue` were already a part of `AnyValueEnum`. For some reason, I didn't think they were. I think this is the right approach then, but I want to think on it a little more.

TheDan64:
I suppose the only alternative is to return `Either<BasicValueEnum, InstructionValue>` as we do elsewhere. Do you think `AnyValueEnum` is the better choice?

nlewycky:
> I suppose the only alternative is to return `Either<BasicValueEnum, InstructionValue>` as we do elsewhere. Do you think `AnyValueEnum` is the better choice?

I think `AnyValueEnum` is the better choice. It's basically an Either<> of the types we care about, plus we can control its methods and traits.

src/values/basic_value_use.rs
```diff
@@ -18,7 +18,7 @@ impl<'ctx> BasicValueUse<'ctx> {
         BasicValueUse(use_, PhantomData)
     }
 
-    /// Gets the next use of an `InstructionValue` or `BasicValue` if any.
+    /// Gets the next use of a `BasicValue` if any.
```
TheDan64, line 21: Isn't the original comment still accurate? It's a use of either an instruction or basic value, no?

TheDan64: Or maybe it should say `AnyValueEnum`?

nlewycky: No, you're right, I shouldn't have changed this at all. The only way to create a BasicValueUse is with the method either on BasicValue or on InstructionValue and that hasn't changed.

PR #164, opened 2020-04-02
Check for RAUW'ing a value with itself. That's just a no-op.

nlewycky:
## Description

Replacing a value with itself is a no-op, but LLVM can enter an infinite loop. Infinite loops are UB in C++, let's just not do that.

## Related Issue

Ran into this after fixing #160, doh.

## How This Has Been Tested

`cargo test --features=llvm8-0`

## Checklist


nlewycky:
> Cool, looks good. Maybe just add a comment about the infinite loop thing so it doesn't accidentally get removed at some point?

Done.

PR #166, opened 2020-04-06
Yup, calling LLVMIsTailCall on a non-CallInst is UB.

nlewycky:
Noticed by inspection.


nlewycky:
For reference, the C++ code in LLVM is:
```c++
LLVMBool LLVMIsTailCall(LLVMValueRef Call) {
  return unwrap<CallInst>(Call)->isTailCall();
}
```
and that `unwrap` is an unchecked pointer cast to CallInst type.

nlewycky:
> Looks good, can you just add a comment prior to the `if` explaining it is required to prevent UB?

Done.

Issue #172, opened 2020-04-20
off by one in ObjectFile API iterators


nlewycky:
**Describe the Bug**
There is no API in LLVM Object library C API to clone a section iterator. The `LLVMMoveToNextSection(self.section_iterator)` method returns void, the contents of `self.section_iterator` stay the same but it refers to a different section.

The implementation of `SectionIterator` checks whether the iterator is one past the end and returns None if it is, then creates a new `Section` out of `self.section_iterator`, then advances `self.section_iterator`. However because there is no deep copy, the `Section` is now holding a one-past-the-end iterator and uses that to answer any queries.

The other bug is that the current implementation always skips the first section, because the first `Section` is returning information through the iterator that has been advanced once.

This bug is repeated for the `RelocationIterator` and `SymbolIterator`.

**To Reproduce**
```rust
use inkwell::memory_buffer::MemoryBuffer;
use std::path::Path;

fn main() {
    let mb = MemoryBuffer::create_from_file(Path::new("test.o")).unwrap();
    let obj = mb.create_object_file().unwrap();
    for section in obj.get_sections() {
        dbg!("<section>");
        dbg!(section.get_name());
        dbg!(section.size());
        dbg!(section.get_address());
    }
}
```
will crash on any .o file. You can use `/bin/ls` for instance. Here's a quick dump using `valgrind`:

```
==39244== Invalid read of size 1
==39244==    at 0x4839CC2: __strlen_sse2 (vg_replace_strmem.c:463)
==39244==    by 0x5A2909: std::ffi::c_str::CStr::from_ptr (c_str.rs:1034)
==39244==    by 0x204871: inkwell::object_file::Section::get_name (object_file.rs:113)
==39244==    by 0x203341: objfile::main (main.rs:9)
==39244==    by 0x20409A: std::rt::lang_start::{{closure}} (rt.rs:67)
==39244==    by 0x5A8A07: {{closure}} (rt.rs:52)
==39244==    by 0x5A8A07: do_call<closure-0,i32> (panicking.rs:331)
==39244==    by 0x5A8A07: try<i32,closure-0> (panicking.rs:274)
==39244==    by 0x5A8A07: catch_unwind<closure-0,i32> (panic.rs:394)
==39244==    by 0x5A8A07: std::rt::lang_start_internal (rt.rs:51)
==39244==    by 0x204076: std::rt::lang_start (rt.rs:67)
[...]

PR #174, opened 2020-04-20
Don't increment the iterator while a Section/Relocation/Symbol is using it.

nlewycky:
## Description

Don't increment the iterator while a Section/Relocation/Symbol is using it.

## Related Issue

#172 

## How This Has Been Tested

There aren't unit tests for object_file yet. I think that in order to add some, we will need some binary test files in the inkwell tree. Thoughts?

As for for I tested it, the simple test driver program is:

```rust
use inkwell::memory_buffer::MemoryBuffer;
use std::path::Path;

fn main() {
    let mb = MemoryBuffer::create_from_file(Path::new("test.o")).unwrap();
    let obj = mb.create_object_file().unwrap();
    let mut first = true;
    for section in obj.get_sections() {
        dbg!("<section>");
        if !first {
            dbg!(section.get_name());
        } else {
            first = false;
        }
        dbg!(section.size());
        dbg!(section.get_address());
        for rel in section.get_relocations() {
            dbg!(rel.get_offset());
            /*
            for symbol in rel.get_symbols() {
                dbg!("<relocation symbol>");
                dbg!(symbol.get_name());
                dbg!(symbol.size());
                dbg!(symbol.get_address());
            }
            */
            dbg!(rel.get_type());
            dbg!(rel.get_value());
        }
    }
    for symbol in obj.get_symbols() {
        dbg!("<symbol>");
        //dbg!(symbol.get_name());
        dbg!(symbol.size());
        dbg!(symbol.get_address());
    }
}
```

## Checklist

### DIFF
--- a/src/object_file.rs
+++ b/src/object_file.rs
@@ -48,6 +48,7 @@ impl Drop for ObjectFile {
 pub struct SectionIterator {
     section_iterator: LLVMSectionIteratorRef,
     object_file: LLVMObjectFileRef,
+    before_first: bool,
 }
 
 impl SectionIterator {
@@ -56,7 +57,8 @@ impl SectionIterator {
 
         SectionIterator {
             section_iterator,
-            object_file
+            object_file,
+            before_first: true
         }
     }
 }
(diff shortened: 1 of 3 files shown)


TheDan64:
Would it possible to add a test for this?

Issue #173, opened 2020-04-20
section and symbol names are optional


nlewycky:
**Describe the Bug**
In object files, names for sections and symbols are optional. In ELF they have numbers and there's an optional sparse table that associates names for them.

**To Reproduce**
I wanted to offer a way to create such things, but it turns out that normal tooling will attempt to not do so. For instance, the assembly code `"": .byte 1` will emit a named symbol with an empty name, not an unnamed symbol. Fortunately, most compiled binaries begin with an unnamed section. Unfortunately, due to #172, we can't actually query that section using inkwell yet. I noticed this behaviour while working on a fix to #172.

**Expected Behavior**
Inkwell should check whether the pointer returned by LLVM is a nullptr, and if so either return an empty string via `&CStr`, or the function should be changed to return an `Option`.

**LLVM Version (please complete the following information):**
 - LLVM Version: 8
 - Inkwell Branch Used: master, but with llvm-sys changed to 80.0

**Desktop (please complete the following information):**
 - OS: Debian unstable

**Additional Context**
You can see the unnamed section on `/bin/ls` with `llvm-objdump -section-headers /bin/ls`. I couldn't find a handy example of an unnamed symbol, but the relevant spec reads:

> st_name: This member holds an index into the object file's symbol string table, which holds the character representations of the symbol names. If the value is non-zero, it represents a string table index that gives the symbol name. Otherwise, the symbol table entry has no name.

-- https://refspecs.linuxbase.org/elf/gabi4+/ch4.symtab.html

PR #175, opened 2020-04-20
Make symbol and section names optional, for when llvm returns a nullptr.

nlewycky:
##  Description

The first section usually has no name, and LLVM returns a nullptr. Inkwell will dereference that to return the `&CStr`. The same thing should be able to happen with a `Symbol` but I haven't observed that myself. When LLVM returns a nullptr string as the name, return None.

## Related Issue

#173 

## How This Has Been Tested

Tried dumping names of all symbols and sections on a few object files.

## Option\<Breaking Changes\>

`get_name` now returns an Option instead of a `&CStr`.

It might be possible to return an empty string &CStr, but I tried and failed.

## Checklist

### DIFF
--- a/src/object_file.rs
+++ b/src/object_file.rs
@@ -108,9 +108,14 @@ impl Section {
         }
     }
 
-    pub fn get_name(&self) -> &CStr {
-        unsafe {
-            CStr::from_ptr(LLVMGetSectionName(self.section))
+    pub fn get_name(&self) -> Option<&CStr> {
+        let name = unsafe {
+            LLVMGetSectionName(self.section)
+        }
+        if !name.is_null() {
+            CStr::from_ptr(name)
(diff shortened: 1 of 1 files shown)


src/object_file.rs
```diff
@@ -108,9 +108,14 @@ impl Section {
...
 
-    pub fn get_name(&self) -> &CStr {
-        unsafe {
-            CStr::from_ptr(LLVMGetSectionName(self.section))
+    pub fn get_name(&self) -> Option<&CStr> {
+        let name = unsafe {
+            LLVMGetSectionName(self.section)
+        }
+        if !name.is_null() {
+            CStr::from_ptr(name)
```
TheDan64, line 116: Doesn't this need to be wrapped in `Some()`?

Issue #180, opened 2020-04-28
Add support for DIBuilder - debug info


epilys:
`llvm-sys` has bindings for the debug symbol API, but `inkwell` lacks it. I have drafted a quick implementation and I'd like to receive some feedback.
Haven't tried it with the Kaleidoscope example yet, but with my code I get this error I'm trying to figure out:
```text
warning: ignoring debug info with an invalid version (0) in output.ll
```

IR output:
```text
; ModuleID = 'bin'
source_filename = "bin"
target datalayout = "e-m:e-i64:64-f80:128-n8:16:32:64-S128"

@"Hello world!\0A" = private unnamed_addr constant [14 x i8] c"Hello world!\0A\00", align 1

define i64 @main() {
entry:
  call void @puts(i8* getelementptr inbounds ([14 x i8], [14 x i8]* @"Hello world!\0A", i32 0, i32 0)), !dbg !3
  ret i64 0, !dbg !3
}

!llvm.dbg.cu = !{!0}

!0 = distinct !DICompileUnit(language: DW_LANG_C, file: !1, producer: "toy compiler", isOptimized: false, runtimeVersion: 1, emissionKind: FullDebug, enums: !2, splitDebugInlining: false)
!1 = !DIFile(filename: "bin", directory: ".")
!2 = !{}
!3 = !DILocation(line: 3, scope: !4)
!4 = distinct !DISubprogram(name: "main", linkageName: "main", scope: null, file: !1, line: 3, type: !5, scopeLine: 3, flags: DIFlagPublic, spFlags: DISPFlagLocalToUnit | DISPFlagDefinition | DISPFlagOptimized, unit: !0, retainedNodes: !2)
!5 = !DISubroutineType(flags: DIFlagPublic, types: !6)
!6 = !{!7}
!7 = !DIBasicType(name: "()", size: 64, encoding: DW_ATE_signed, flags: DIFlagPublic)

```

TheDan64:
Added in #203

PR #182, opened 2020-04-28
DebugBuilder #181

epilys:
## Related Issue
#180 

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -13,6 +13,8 @@ use crate::{AtomicOrdering, AtomicRMWBinOp, IntPredicate, FloatPredicate};
 use crate::basic_block::BasicBlock;
 use crate::support::to_c_str;
 use crate::values::{AggregateValue, AggregateValueEnum, AsValueRef, BasicValue, BasicValueEnum, PhiValue, FunctionValue, IntValue, PointerValue, VectorValue, InstructionValue, GlobalValue, IntMathValue, FloatMathValue, PointerMathValue, InstructionOpcode, CallSiteValue};
+#[llvm_versions(7.0..=latest)]
+use crate::debug_info::DILocation;
 #[llvm_versions(3.9..=latest)]
 use crate::values::StructValue;
 use crate::types::{AsTypeRef, BasicType, IntMathType, FloatMathType, PointerType, PointerMathType};
@@ -1762,6 +1764,60 @@ impl<'ctx> Builder<'ctx> {
         };
 
         Ok(StructValue::new(val))
+    }
+
+    /// Set the debug info source location of the instruction currently pointed at by the builder
+    #[llvm_versions(7.0..=latest)]
+    pub fn set_current_debug_location(
+        &self,
+        context: &'ctx crate::context::Context,
+        location: DILocation<'ctx>,
+    ) {
+        use llvm_sys::core::LLVMMetadataAsValue;
+        use llvm_sys::core::LLVMSetCurrentDebugLocation;
+        unsafe {
+            LLVMSetCurrentDebugLocation(
+                self.builder,
+                LLVMMetadataAsValue(context.context, location.metadata_ref),
+            );
+        }
+    }
+
+    /// Get the debug info source location of the instruction currently pointed at by the builder,
+    /// if available.
+    #[llvm_versions(7.0..=latest)]
+    pub fn get_current_debug_location(
+        &self,
+    ) -> Option<DILocation<'ctx>>
+     {
+        use llvm_sys::core::LLVMGetCurrentDebugLocation;
+        use llvm_sys::core::LLVMValueAsMetadata;
+        let metadata_ref = unsafe {
+            LLVMGetCurrentDebugLocation(
+                self.builder,
+            )
+        };
+        if metadata_ref.is_null() {
+            return None;
+        }
+        Some(DILocation {
+            metadata_ref: unsafe { LLVMValueAsMetadata(metadata_ref)},
+            _marker: PhantomData,
+        })
+    }
+
+    /// Unset the debug info source location of the instruction currently pointed at by the
+    /// builder. If there isn't any debug info, this is a no-op.
+    pub fn unset_current_debug_location(
+        &self,
+    ) {
+        use llvm_sys::core::LLVMSetCurrentDebugLocation;
+        unsafe {
+            LLVMSetCurrentDebugLocation(
+                self.builder,
+                std::ptr::null_mut(),
+            );
+        }
     }
 }
 
--- a/src/debug_info.rs
+++ b/src/debug_info.rs
@@ -1,30 +1,1060 @@
-use llvm_sys::debuginfo::{LLVMDebugMetadataVersion, LLVMDisposeDIBuilder};
-use llvm_sys::prelude::LLVMDIBuilderRef;
+//! Debug symbols - `DebugInfoBuilder` interface
+//!
+//! # Example usage
+//!
+//! ## Setting up the module for holding debug info:
+//! ```ignore
+//! let context = Context::create();
+//! let module = context.create_module("bin");
+//!
+//! let debug_metadata_version = context.i32_type().const_int(3, false);
+//! module.add_basic_value_flag(
+//!     "Debug Info Version",
+//!     inkwell::module::FlagBehavior::Warning,
+//!     debug_metadata_version,
+//! );
+//! let builder = context.create_builder();
+//! let dibuilder = module.create_debug_info_builder(true);
+//!
+//! let compile_unit = dibuilder.create_compile_unit(
+//!     /* language */ inkwell::debug_info::DWARFSourceLanguage::C,
+//!     dibuilder.create_file(/* filename */ "source_file", /* directory */ "."),
+//!     /* producer */ "my llvm compiler frontend",
+//!     /* is_optimized */ false,
+//!     /* compiler command line flags*/ "",
+//!     /* runtime_ver */ 0,
+//!     /* split_name */ "",
+//!     /* kind */ inkwell::debug_info::DWARFEmissionKind::Full,
+//!     /* dwo_id */ 0,
+//!     /* split_debug_inling */ false,
+//!     /* debug_info_for_profiling */ false,
+//! );
+//! ```
+//! ## Creating function debug info
+//! ```ignore
+//!  let ditype = dibuilder.create_basic_type(
+//!      "type_name",
+//!      0_u64,
+//!      0x00,
+//!      inkwell::debug_info::DIFlags::Public,
+//!  );
+//!  let subroutine_type = dibuilder.create_subroutine_type(
+//!      compile_unit.get_file(),
+//!      /* return type */ ditype,
+//!      /* parameter types */ vec![],
+//!      inkwell::debug_info::DIFlags::Public,
+//!  );
+//!  let func_scope: DISubProgram<'_> = dibuilder.create_function(
+//!      /* scope */ compile_unit.as_debug_info_scope(),
+//!      /* func name */ "main",
+//!      /* linkage_name */ None,
+//!      /* file */ compile_unit.get_file(),
+//!      /* line_no */ 0,
+//!      /* DIType */ subroutine_type,
+//!      /* is_local_to_unit */ true,
+//!      /* is_definition */ true,
+//!      /* scope_line */ 0,
+//!      /* flags */ inkwell::debug_info::DIFlags::Public,
+//!      /* is_optimized */ false,
+//!  );
+//! ```
+//! The `DISubProgram` value must be attached to the generated `FunctionValue`:
+//! ```ignore
+//! /* after creating function: */
+//!     let fn_val = module.add_function(fn_name_str, fn_type, None);
+//!     fn_val.set_subprogram(func_scope);
+//! ```
+//!
+//! ## Setting debug locations
+//! ```ignore
+//! let lexical_block = dibuilder.create_lexical_block(
+//!         /* scope */ func_scope.as_debug_info_scope(),
+//!         /* file */ compile_unit.get_file(),
+//!         /* line_no */ 0,
+//!         /* column_no */ 0);
+//!
+//! let loc = dibuilder
+//!     .create_debug_location(context, /*line*/ 0, /* column */ 0,
+//!     /* current_scope */ lexical_block.as_debug_info_scope());
+//! builder.set_current_debug_location(context, loc);
+//! ```
+//!
+//! ## Finalize debug info
+//! Before any kind of code generation (including verification passes; they generate code and
+//! validate debug info), do:
+//! ```ignore
+//! dibuilder.finalize();
+//! ```
+
+use crate::basic_block::BasicBlock;
+use crate::context::Context;
+use crate::module::Module;
+use crate::support::to_c_str;
+use crate::values::AsValueRef;
+use crate::values::BasicValueEnum;
+use crate::values::{InstructionValue, PointerValue};
+pub use llvm_sys::debuginfo::LLVMDWARFTypeEncoding;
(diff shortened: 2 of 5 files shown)


src/debug_info.rs
```diff
@@ -1,30 +1,475 @@
...
+                if is_optimized { 0 } else { 1 },
+                flags.as_ptr(),
+                flags.as_bytes().len(),
+                runtime_ver,
+                split_name.as_ptr(),
+                split_name.as_bytes().len(),
+                kind,
+                dwo_id,
+                if split_debug_inlining { 0 } else { 1 },
+                if debug_info_for_profiling { 0 } else { 1 },
```
TheDan64, line 160: Shouldn't these be 1 else 0?

src/module.rs
```diff
@@ -136,6 +135,7 @@ enum_rename!{
 #[derive(Debug, PartialEq, Eq)]
 pub struct Module<'ctx> {
     data_layout: RefCell<Option<DataLayout>>,
+    dibuilder: RefCell<Option<DebugInfoBuilder<'ctx>>>,
```
TheDan64, line 138: It's not clear to me, why does the DIB need to live in the Module?

src/debug_info.rs
```diff
@@ -1,30 +1,475 @@
...
+    /// #[allow(non_upper_case_globals)]
+    /// const DW_ATE_unsigned: c_uint = 0x07;
+    /// #[allow(non_upper_case_globals)]
+    /// const DW_ATE_unsigned_char: c_uint = 0x08;
+    /// ```
+    pub fn create_basic_type(
+        &self,
+        name: &str,
+        size_in_bits: u64,
+        encoding: LLVMDWARFTypeEncoding,
```
TheDan64, line 331: You're going to need to create a new wrapper enum for this. See https://github.com/TheDan64/inkwell/blob/9dd5214a1c5ef5b8a9985847624eaf10c535af9e/src/values/global_value.rs#L336-L350 for example

TheDan64:
This looks good so far. I think you're on the right track.

src/module.rs
epilys: To call `dispose` when dropped. I was thinking of tracking if `finalize` has been called inside DIB in order to decide if it needs to be called before `dispose`.

I'd also like to call `finalize` automatically whenever code generation happens but that doesn't seem possible, it should be called even whenever a verification pass is performed (for example in function or module verify) and there's no clear way to do that.

src/debug_info.rs
epilys: For this and every other LLVM flag re-export, right?

src/debug_info.rs
epilys: I got wrong values ( when llvm reported errors, and searching for "LLVMBool" got me [this](http://lists.llvm.org/pipermail/llvm-dev/2016-September/104766.html) and [this](https://ubiquitydotnet.github.io/Llvm.NET/articles/InternalDetails/marshal-LLVMBool.html) This seems incorrect now.

The problem was I was setting `is_definition` false on `create_function` and I got an unresolved `<temporary>` error: 

```
!4 = !DISubprogram(name: "main", linkageName: "main", scope: null, file: !1, line: 3, type: !5, scopeLine: 3, flags: DIFlagPublic, spFlags: 0, retainedNodes: !8)
!5 = !DISubroutineType(flags: DIFlagPublic, types: !6)
!6 = !{!7}
!7 = !DIBasicType(name: "()", size: 64, encoding: DW_ATE_signed, flags: DIFlagPublic)
!8 = <temporary!> !{}

and the error:
llc-8: output.ll:63:6: error: Expected '!' here
!8 = <temporary!> !{}
     ^
```

and setting it to true removed that `<temporary>`. I don't know what `retainedNodes` is, unfortunately.

epilys: I got it working, I will try updating the Kaleidoscope example and update the PR.

TheDan64: As I understand it, `LLVMBool` being `0` is success/true only applies to status codes and/or out params. So as an input function param, `1` should mean true. (I think)

It's madness that we even have to discuss this...

Good find on those links, btw.

src/module.rs
TheDan64: > To call dispose when dropped.

Why can't we call dispose on `DebugInfoBuilder`'s drop impl, though?

>  I was thinking of tracking if finalize has been called inside DIB in order to decide if it needs to be called before dispose.

This is a valid option. Especially if LLVM doesn't have a getter for this already...

> I'd also like to call finalize automatically whenever code generation happens but that doesn't seem possible, it should be called even whenever a verification pass is performed (for example in function or module verify) and there's no clear way to do that.

Should be, or must be? If nothing bad happens (ie UB) if you neglect to do so then I think the API user should be responsible for calling finalize prior to codegen / passes / etc.

TheDan64: Also, does calling finalize make the DIBuilder immutable (ie can't make changes to it?). If we we might want two different states. Like `DIBuilder<Unfinalized>` and `DIBuilder<Finalized>` so that a finalized DIB can't call its methods (I'm assuming bad things happen in that scenario here...)

src/debug_info.rs
```diff
@@ -1,30 +1,730 @@
...
+//!  );
+//! ```
+//! The `DISubProgram` value must be attached to the generated `FunctionValue`:
+//! ```ignore
+//! /* after creating function: */
+//!     let fn_val = module.add_function(fn_name_str, fn_type, None);
+//!     fn_val.set_subprogram(func_scope);
+//! ```
+//!
+//! ## Setting debug locations
```
TheDan64, line 68: Thanks for all of these great documentation examples. This is awesome!

src/module.rs
epilys: > Why can't we call dispose on DebugInfoBuilder's drop impl, though?

I made it `Copy` when I was testing stuff and `Copy` implementors can't `Drop`. I can change that back.

 > Should be, or must be? If nothing bad happens (ie UB) if you neglect to do so then I think the API user should be responsible for calling finalize prior to codegen / passes / etc.

Definitely *must*, otherwise debug info is in incomplete state and invalid. 

> Also, does calling finalize make the DIBuilder immutable (ie can't make changes to it?). If we we might want two different states. Like DIBuilder<Unfinalized> and DIBuilder<Finalized> so that a finalized DIB can't call its methods (I'm assuming bad things happen in that scenario here...)

Haven't tested that yet. If it is prohibited to use DIBuilder after finalize, the function can take `self` instead of a reference (if it isn't `Copy`) and consume the builder.

src/values/fn_value.rs
```diff
@@ -505,6 +507,25 @@ impl<'ctx> FunctionValue<'ctx> {
     pub fn as_global_value(self) -> GlobalValue<'ctx> {
         GlobalValue::new(self.as_value_ref())
     }
+
+    /// Set the debug info descriptor
+    pub fn set_subprogram(self, subprogram: DISubProgram<'ctx>) {
```
TheDan64, line 512: The two imports and functions calls here need the `#[llvm_versions(6.0..=latest)]` attr applied to play nice with older versions

src/module.rs
TheDan64: > I made it Copy when I was testing stuff and Copy implementors can't Drop. I can change that back.

I see, I didn't notice that it was marked Copy.

> Definitely must, otherwise debug info is in incomplete state and invalid.

Is this only an IR issue though? We can't prevent all runtime IR errors. (We should certainly try to prevent as much as possible, but not everything can be prevented). I'm more concerned about memory safety here. Writing some tests and/or looking through LLVM source would probably be helpful (if you get stuck, I can try and take a look too)

> Haven't tested that yet. If it is prohibited to use DIBuilder after finalize, the function can take self instead of a reference (if it isn't Copy) and consume the builder.

Yeah that'd be a better option iff there's nothing to be called on a finalized builder

TheDan64: The fact that finalize and dispose are two separate functions makes me suspicious that there's something you can do with a finalized builder.

TheDan64:
Also, what did you discover wrt the finalize stuff? Does not calling `finalize` before pass manager/other types not invoke UB, just generate invalid IR?

epilys:
> Also, what did you discover wrt the finalize stuff? Does not calling `finalize` before pass manager/other types not invoke UB, just generate invalid IR?

It will result in errors. The error message isn't always clear that it's caused by not calling `finalize`, however.

TheDan64:
@epilys How is this going? Anything you need pointers or advice on?

epilys:
> @epilys How is this going? Anything you need pointers or advice on?

I'm kind of busy with other things atm, but I haven't forgotten about it. I plan on picking it up again eventually!

TheDan64:
@epilys Cool. Thanks for the update!

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+                unique_id.to_bytes().len(),
+            )
+        };
+        DICompositeType {
+            metadata_ref,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Create a function type
```
nlewycky, line 491:
```suggestion
    /// Create a subroutine type
```
Or if you like `subroutine (function)`? Depending on the source language, subroutines might or might not be functions.

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+                    .unwrap_or(std::ptr::null_mut()),
+                expr.unwrap_or(self.create_expression(vec![])).metadata_ref,
+                debug_loc.metadata_ref,
+                instruction.as_value_ref(),
+            )
+        };
+        InstructionValue::new(value_ref)
+    }
+
+    /// Insert a variable declaration (`llvm.dbg.declare` intrinsic)  at the end of `block`
```
nlewycky, line 581:
```suggestion
    /// Insert a variable declaration (`llvm.dbg.declare` intrinsic) at the end of `block`
```

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+            )
+        };
+        InstructionValue::new(value_ref)
+    }
+
+    /// Construct any deferred debug info descriptors. May generate invalid metadata if debug info
+    /// is incomplete.  Module/function verification can then fail.
+    ///
+    /// Call before any kind of code generation (including verification). Can be called more than
+    /// one time.
```
nlewycky, line 652:
```suggestion
    /// once.
```

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+                expr.unwrap_or(self.create_expression(vec![])).metadata_ref,
+                debug_loc.metadata_ref,
+                instruction.as_value_ref(),
+            )
+        };
+        InstructionValue::new(value_ref)
+    }
+
+    /// Construct any deferred debug info descriptors. May generate invalid metadata if debug info
+    /// is incomplete.  Module/function verification can then fail.
```
nlewycky, line 649:
```suggestion
    /// is incomplete. Module/function verification can then fail.
```

src/builder.rs
```diff
@@ -1763,6 +1765,60 @@ impl<'ctx> Builder<'ctx> {
...
+        }
+        Some(DILocation {
+            metadata_ref: unsafe { LLVMValueAsMetadata(metadata_ref)},
+            _marker: PhantomData,
+        })
+    }
+
+    /// Unset the debug info source location of the instruction currently pointed at by the
+    /// builder. If there isn't any debug info, this is a no-op.
+    pub fn unset_current_debug_location(
```
nlewycky, line 1811: Is this missing `#[llvm_versions(7.0..=latest)]`?

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+
+/// Specific scopes (i.e. `DILexicalBlock`) can be turned into a `DIScope` with the
+/// `AsDIScope::as_debug_info_scope` trait method.
+pub trait AsDIScope<'ctx> {
+    fn as_debug_info_scope(self) -> DIScope<'ctx>;
+}
+
+impl<'ctx> DebugInfoBuilder<'ctx> {
+    pub(crate) fn new(module: &Module, allow_unresolved: bool) -> Self {
+        use llvm_sys::debuginfo::{LLVMCreateDIBuilder, LLVMCreateDIBuilderDisallowUnresolved};
```
nlewycky, line 133: When this isn't cfg-attr conditional, please move it to the top of the file.

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+        unsafe { LLVMDITypeGetAlignInBits(self.metadata_ref) }
+    }
+
+    pub fn get_offset_in_bits(&self) -> u64 {
+        use llvm_sys::debuginfo::LLVMDITypeGetOffsetInBits;
+        unsafe { LLVMDITypeGetOffsetInBits(self.metadata_ref) }
+    }
+}
+
+/// A debug info typedef, created by `create_typedef` method of `DebugInfoBuilder`
```
nlewycky, line 747: `DIDerivedType` includes "qualified types, pointers, references, friends, typedefs, and class members".

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+            _marker: PhantomData,
+        }
+    }
+}
+
+/// A debug location within the source code. Contains the following information:
+///
+/// - line, column
+/// - scope
+/// - is inlined
```
nlewycky, line 822: Where it's been inlined into. The idea is that when code is initially parsed this is blank, then when the LLVM optimizer begins inlining it fills this in as it duplicates the LLVM IR frmo the callee into the caller. LLVM will use this to emit DWARF inlined subroutine debug info.

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+    pub(crate) metadata_ref: LLVMMetadataRef,
+    _marker: PhantomData<&'ctx Context>,
+}
+
+pub use flags::*;
+mod flags {
+    use llvm_sys::debuginfo::{LLVMDIFlags, LLVMDWARFEmissionKind, LLVMDWARFSourceLanguage};
+
+    /// Debug info flags. Corresponds to `LLVMDIFlags` enum from LLVM.
+    #[llvm_enum(LLVMDIFlags)]
```
nlewycky, line 877: These flags are intended to be OR'd together.

src/values/fn_value.rs
nlewycky: FYI, LLVMSetSubprogram was added in 7.0 not 6.0.

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+            metadata_ref,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Create a function type
+    pub fn create_subroutine_type(
+        &self,
+        file: DIFile<'ctx>,
+        return_type: DIType<'ctx>,
```
nlewycky, line 495: I don't see any way to make a function that returns void.

I think this should be Option<DIType<'ctx>>, and when it's None, the first entry in `p` would be `std::ptr::null_mut()`.

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+        parameter_types: Vec<DIType<'ctx>>,
```
nlewycky, line 496:
```suggestion
        parameter_types: &[DIType<'ctx>],
```

src/debug_info.rs
```diff
@@ -1,30 +1,1060 @@
...
+        DIScope {
+            metadata_ref: self.metadata_ref,
+            _marker: PhantomData,
+        }
+    }
+}
+
+/// Any kind of debug info type
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct DIType<'ctx> {
```
nlewycky, line 725: In LLVM C++ `DIType`s is derived from `DIScope`. Any reason not to have the DITypes implement `AsDIScope`?

TheDan64:
Closing in favor of #203 - thank you for getting the ball rolling!

PR #190, opened 2020-05-26
Get/set of volatile on atomicrmw and cmpxchg was added in LLVM 10.

nlewycky:
## Description

If building for llvm 10+, permit get_volatile and set_volatile on atomicrmw and cmpxchg instructions.

## Related Issue

n/a

## How This Has Been Tested

I've only checked that it builds.

## Option\<Breaking Changes\>

The error message changed to include the additional instructions in the list. Anyone with code relying on that string will find that it doesn't match. It would be surprising if anyone wrote code relying on that string, since it was the only possible string it could return.

## Checklist


src/values/instruction_value.rs
```diff
@@ -203,25 +213,49 @@ impl<'ctx> InstructionValue<'ctx> {
...
+        // does not export that functionality until 10.0.
         if !self.is_a_load_inst() && !self.is_a_store_inst() {
             return Err("Value is not a load or store.");
         }
         Ok(unsafe { LLVMSetVolatile(self.as_value_ref(), volatile as i32) })
     }
+    
+    // SubTypes: Only apply to memory access instructions
+    /// Sets whether or not a memory access instruction is volatile.
+    #[llvm_versions(10.0..=latest)]
```
TheDan64, line 251: I wonder if, rather than duplicating these functions we can just simplify like so:

```Rust
// Comment here why this is necessary
#[llvm_versions(3.6..=9.0)]
let not_atomicrmw_cmpxchg = true;
#[llvm_version(10.0..=latest)]
let not_atomicrmw_cmpxchg = !self.is_a_atomicrmw_inst() && !self.is_a_cmpxchg_inst();

if !self.is_a_load_inst() && !self.is_a_store_inst() && not_atomicrmw_cmpxchg {
```

src/values/instruction_value.rs
```diff
@@ -113,6 +115,14 @@ impl<'ctx> InstructionValue<'ctx> {
     fn is_a_alloca_inst(self) -> bool {
         !unsafe { LLVMIsAAllocaInst(self.as_value_ref()) }.is_null()
     }
+    #[llvm_versions(10.0..=latest)]
+    fn is_a_atomicrmw_inst(self) -> bool {
+        !unsafe { LLVMIsAAtomicRMWInst(self.as_value_ref()) }.is_null()
+    }
+    #[llvm_versions(10.0..=latest)]
+    fn is_a_cmpxchg_inst(self) -> bool {
```
TheDan64, line 123: Can you add a test for this? I think you should be able to slide it into the existing tests somewhere

src/values/instruction_value.rs
nlewycky: I tried that and got this error:
```
error[E0658]: custom attributes cannot be applied to statements
   --> src/values/instruction_value.rs:244:9
    |
244 |         #[llvm_versions(10.0..=latest)]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #54727 <https://github.com/rust-lang/rust/issues/54727> for more information
    = help: add `#![feature(proc_macro_hygiene)]` to the crate attributes to enable
```

I tried a couple implementations, one putting it above each `let` statement as you suggest, and another factoring out the let part and putting the feature test on the block that it's equal to. I couldn't get it to work.

src/values/instruction_value.rs
nlewycky: Done. It gets a new test function so that I can make it llvm10+ only.

src/values/instruction_value.rs
TheDan64: Weird, I thought they worked on statements. I think support for them on exprs is coming in a couple of versions. But I guess this is fine for now.

PR #203, opened 2020-07-08
Add support for debug info.

nlewycky:
## Description

This is built off of epilys' PR #180. I've addressed the review feedback and added the extensions that I need for my use case.

## Related Issue

#180 

## How This Has Been Tested

I've used this to generate line table info but indeed there's no tests included yet.

## Checklist

### DIFF
--- a/src/debug_info.rs
+++ b/src/debug_info.rs
@@ -125,8 +125,6 @@ pub fn debug_metadata_version() -> libc::c_uint {
 #[derive(Debug, PartialEq, Eq)]
 pub struct DebugInfoBuilder<'ctx> {
     pub(crate) builder: LLVMDIBuilderRef,
-    placeholders: Vec<LLVMMetadataRef>,
-    has_cu: bool,
     _marker: PhantomData<&'ctx Context>,
 }
 
@@ -146,7 +144,22 @@ pub trait AsDIScope<'ctx> {
 }
 
 impl<'ctx> DebugInfoBuilder<'ctx> {
-    pub(crate) fn new(module: &Module, allow_unresolved: bool) -> Self {
+    pub(crate) fn new(
+        module: &Module,
+        allow_unresolved: bool,
+        language: DWARFSourceLanguage,
+        filename: &str,
+        directory: &str,
+        producer: &str,
+        is_optimized: bool,
+        flags: &str,
+        runtime_ver: libc::c_uint,
+        split_name: &str,
+        kind: DWARFEmissionKind,
+        dwo_id: libc::c_uint,
+        split_debug_inlining: bool,
+        debug_info_for_profiling: bool,
+    ) -> (Self, DICompileUnit<'ctx>) {
         let builder = unsafe {
             if allow_unresolved {
                 LLVMCreateDIBuilder(module.module.get())
@@ -155,12 +168,28 @@ impl<'ctx> DebugInfoBuilder<'ctx> {
             }
         };
 
-        DebugInfoBuilder {
+        let builder = DebugInfoBuilder {
             builder,
-            placeholders: vec![],
-            has_cu: false,
             _marker: PhantomData,
-        }
+        };
+
+        let file = builder.create_file(filename, directory);
+
+        let cu = builder.create_compile_unit(
+            language,
+            file,
+            producer,
+            is_optimized,
+            flags,
+            runtime_ver,
+            split_name,
+            kind,
+            dwo_id,
+            split_debug_inlining,
+            debug_info_for_profiling,
+        );
+
+        (builder, cu)
     }
 
     /// A DICompileUnit provides an anchor for all debugging information generated during this instance of compilation.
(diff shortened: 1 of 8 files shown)


src/debug_info.rs
```diff
@@ -1,30 +1,1221 @@
...
+        flags: &str,
+        runtime_ver: libc::c_uint,
+        split_name: &str,
+        kind: DWARFEmissionKind,
+        dwo_id: libc::c_uint,
+        split_debug_inlining: bool,
+        debug_info_for_profiling: bool,
+    ) -> DICompileUnit<'ctx> {
+        if self.has_cu {
+            panic!("Only one compile unit per DebugInfoBuilder.");
```
TheDan64, line 194: Does this need to panic? Shouldn't we return a Result?

src/debug_info.rs
```diff
@@ -1,30 +1,1221 @@
...
+    pub fn create_basic_type(
+        &self,
+        name: &str,
+        size_in_bits: u64,
+        encoding: LLVMDWARFTypeEncoding,
+    ) -> DIBasicType<'ctx> {
+        if name.empty() {
+            // Also, LLVM returns the same type if you ask for the same
+            // (name, size_in_bits, encoding).
+            panic!("basic types must have names");
```
TheDan64, line 357: Result?

src/debug_info.rs
```diff
@@ -1,30 +1,1221 @@
...
+        unsafe { LLVMMetadataReplaceAllUsesWith(placeholder.metadata_ref, other.metadata_ref) };
+    }
+
+    /// Construct any deferred debug info descriptors. May generate invalid metadata if debug info
+    /// is incomplete. Module/function verification can then fail.
+    ///
+    /// Call before any kind of code generation (including verification). Can be called more than once.
+    pub fn finalize(&self) {
+        if !self.placeholders.is_empty() {
+            panic!("Can't finalize debug info builder with outstanding placeholder types!");
```
TheDan64, line 790: Result?

src/debug_info.rs
```diff
@@ -1,30 +1,1221 @@
...
 }
 
-impl Drop for DebugInfoBuilder {
+impl<'ctx> Drop for DebugInfoBuilder<'ctx> {
     fn drop(&mut self) {
-        // FIXME: Docs say DIB must be finalize before calling this
-        unsafe {
-            LLVMDisposeDIBuilder(self.0)
+        if !self.has_cu {
+            panic!("A DebugInfoBuilder must have a compile unit.");
```
TheDan64, line 799: Result?

src/debug_info.rs
```diff
@@ -1,30 +1,1221 @@
...
+        &mut self,
+        placeholder: DIDerivedType<'ctx>,
+        other: DIDerivedType<'ctx>,
+    ) {
+        // `placeholder` must be in the placeholder list.
+        // `other` may or may not be a placeholder.
+        let index = self
+            .placeholders
+            .iter()
+            .position(|ph| *ph == placeholder.metadata_ref)
```
TheDan64, line 778: Does this intentionally stop at the first one? Can there be multiple?

src/debug_info.rs
nlewycky: This panic! is removed.

src/debug_info.rs
nlewycky: Done.

src/debug_info.rs
nlewycky: This panic is removed.

src/debug_info.rs
nlewycky: This panic is removed.

src/debug_info.rs
nlewycky: It was stopping at the first one and they needed to be unique. Regardless, this entire technique doesn't work (DebugInfoBuilder is Copy and then the `placeholders` member gets out of sync) so this has been deleted.

src/module.rs
```diff
@@ -1358,20 +1357,9 @@ impl<'ctx> Module<'ctx> {
...
-            if allow_unresolved {
-                LLVMCreateDIBuilder(self.module.get())
-            } else {
-                LLVMCreateDIBuilderDisallowUnresolved(self.module.get())
-            }
-        };
-
-        DebugInfoBuilder::new(dib)
+    #[llvm_versions(7.0..=latest)]
+    pub fn create_debug_info_builder(&self, allow_unresolved: bool) -> DebugInfoBuilder<'ctx> {
```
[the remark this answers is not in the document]
nlewycky: I was going off of `create_debug_info_builder` which is marked as being 7+. I haven't actually checked llvm-sys. We could move it to 6 and see if it works on travis?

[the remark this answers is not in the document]
TheDan64: https://gitlab.com/taricorp/llvm-sys.rs/-/blob/llvm-6.0/src/debuginfo.rs
https://gitlab.com/taricorp/llvm-sys.rs/-/blob/llvm-7.0/src/debuginfo.rs

It definitely won't work wholesale but would in part. Do you think it's worth supporting 6? Might not be with how little is there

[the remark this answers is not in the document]
nlewycky: My general answer is that we don't need to add new features for old LLVMs, people using inkwell+llvm 7 are already happy with the features it supports and people who start using inkwell today will pick the newest llvm.

That aside. It looks like for LLVM 6 they tried to include just enough to make line tables work, except that I don't see Subprogram support. When I tried dbg! statements without declaring a subprogram in LLVM 10 the verifier complained that the scope wasn't a local scope. LLVM 6 might permit that though?

Zaxar163:
When will this Pull Request be accepted?

TheDan64:
@Zaxar163 Hopefully soon. We just need to figure out why LLVM 10 decided to stop linking: https://travis-ci.org/github/TheDan64/inkwell/jobs/708168713. I believe that's the only blocker

TheDan64:
Thanks!

Issue #206, opened 2020-08-23
Don't force linking to libffi.so


baszalmstra:
**Describe the Bug**
I'm trying to create a completely static binary including inkwell however inkwell forces to always dynamically link to libffi with this line in `build.rs`

```rust
if cfg!(not(target_os = "windows")) {
    println!("cargo:rustc-link-lib=dylib=ffi");
}
```

`llvm-sys` works around this issue by introducing an environment variable: LLVM_SYS_{}_FFI_WORKAROUND, if declared it forces the link, otherwise it doesnt.

Another option could be to add a feature flag?

**To Reproduce**
Create a build in an `alpine` docker. 

**Expected Behavior**
I expect the generated binary to not depend on `libffi.so` if LLVM does not link to libffi in the first place.

**LLVM Version (please complete the following information):**
 - LLVM Version: 8.0.1
 - Inkwell Branch Used: llvm-8.0

**Desktop (please complete the following information):**
 - OS: Alpine in a docker container from ubuntu 18.04

TheDan64:
Sure, there should be a way to opt-out

baszalmstra:
@TheDan64 I would love to help out with a PR Im just not sure what the right approach is. Here is what I came up with:

1) Build a workaround in Inkwell to disable linking to libffi
2) Remove the code completely and use the workaround already in place in llvm-sys
3) Another solution?

WDYT?

TheDan64:
I'm not really decided on it myself. It seems to me like you'd want libffi in the vast majority of cases. In my experience at least. So I'm currently leaning towards 1). But OTOH, it does seem a little bit silly that we're doing something that llvm-sys lets a user do for us.

TheDan64:
Because I think libffi is the common case, it seems like we're linking it out of convenience, and that we should have an opt out feature flag:

```rust
if cfg!(not(feature = "no-libffi-linking")) {
    println!("cargo:rustc-link-lib=dylib=ffi");
}
```

TheDan64:
Does that seem reasonable?

baszalmstra:
I think it's a little more complicated than that because I think it depends on the build environment. If you build on a machine with LLVM build with `--enable-libffi` then your binary must link to libffi otherwise you get linker errors. However, if that is not the case you dont want to link against libffi. Therefore, since you have to configure LLVM outside of the build system anyway, it would make sense to me to use an environment variable. Does that make sense?

TheDan64:
[...]

PR #208, opened 2020-09-21
Add feature to disable linking to libffi

baszalmstra:
## Description

Adds a feature `no-libffi-linking` (not enabled by default) that stops linking to libffi by default. 

## Related Issue

https://github.com/TheDan64/inkwell/issues/206

## How This Has Been Tested

Building with the feature enabled no longer forces any dependency on libffi which is what I wanted to achieve. 

## Checklist


TheDan64 (approved):
Thanks!

PR #214, opened 2020-10-05
Add debug checks to const gep argument bit sizes to resolve #213

### LINKED ISSUE #213: const_to_int of const_gep results in broken getelementptr call, const_gep segfaults with i64 indexes
**Describe the Bug**
I'm not sure if these are related or not. If not, I apologize for the two-for-one.
const_gep followed by const_to_int results in the gep getting butchered (wrong type and an index goes missing in this example). Also const_gep with i64 values segfaults.

**To Reproduce**
Run this. See comments in code for details.
```rust
use inkwell::AddressSpace;

fn main() {
    let module = context.create_module("my_module");
    let i32_type = context.i32_type();
    let i32_ptr_type = context.i32_type().ptr_type(AddressSpace::Generic);

    let struct_type = context.struct_type(&[i32_type.into(), i32_type.into()], false);

    unsafe { // just being lazy with gep calls.
        let init = struct_type.ptr_type(AddressSpace::Generic).const_zero().const_gep(&[
            i32_type.const_int(0, false).into(),
            i32_type.const_int(1, false).into()
        ]);

        let a = module.add_global(i32_ptr_type, Some(AddressSpace::Generic), "a");
        a.set_constant(true);
        a.set_initializer(&init);
        // Good, normal:
        // @a = constant i32* getelementptr ({ i32, i32 }, { i32, i32 }* null, i32 0, i32 1)

        let b = module.add_global(i32_type, Some(AddressSpace::Generic), "b");
        b.set_constant(true);
        b.set_initializer(&init.const_to_int(i32_type));
        // Why is GEP only passed one index? Where did my 0 go? -----------V
        // @b = constant i32 ptrtoint (i32* getelementptr (i32, i32* null, i32 1) to i32)
        // Types are wrong?^             ^

        let c = module.add_global(i32_ptr_type, Some(AddressSpace::Generic), "c");
        c.set_constant(true);
        c.set_initializer(&init);
        // Same as a - just showing the initializer isn't mutated or unusable or anything.
        // @c = constant i32* getelementptr ({ i32, i32 }, { i32, i32 }* null, i32 0, i32 1)
    }

    println!("{}", module.print_to_string().to_string());
    // This is where I got the above samples. Here's the whole output:
    // ; ModuleID = 'my_module'
    // source_filename = "my_module"
    //
    // @a = constant i32* getelementptr ({ i32, i32 }, { i32, i32 }* null, i32 0, i32 1)
    // @b = constant i32 ptrtoint (i32* getelementptr (i32, i32* null, i32 1) to i32)
    // @c = constant i32* getelementptr ({ i32, i32 }, { i32, i32 }* null, i32 0, i32 1)

    module.verify().unwrap();
    // No errors.

    // This instantly segfaults. Seems using i64s is not valid for the indexing.
    unsafe {
        struct_type.ptr_type(AddressSpace::Generic).const_zero().const_gep(&[
            context.i64_type().const_int(0, false).into(),
            context.i64_type().const_int(1, false).into(),
        ]);
    }
}
```

**Expected Behavior**
The gep on variable b should be the same as a, just with the pointer operation as well.
Less important (to me): The program should not segfault, unless gep with i64 indexes is just invalid or something - might be my mistake. If so, maybe it should take a u32 slice reference instead for the indexes?

 - LLVM version 10.0.1
 - Inkwell Branch Used: Cargo dependency configured as inkwell = { git = "https://github.com/TheDan64/inkwell", branch = "master", features = ["llvm10-0"] }

 - OS: Windows 10

**Additional Context**
LLVM built from source.

In LLVM clone dir:
```
>git show HEAD
commit ef32c611aa214dea855364efd7ba451ec5ec3f74 (HEAD -> release/10.x, tag: llvmorg-10.0.1-rc4, tag: llvmorg-10.0.1, origin/release/10.x)
```

In the clone of inkwell that cargo pulled:
```
>git show HEAD
commit 67981a09816b705a7304aaa995e29848cb832426 (HEAD -> master, origin/master)
```

djozis:
## Description

Add debug checks to const gep argument bit sizes.

## Related Issue

#213 

## How This Has Been Tested

Tested by running tests in a dependent project, which exercised passing i64 and i32 values. Tests are similar to those in #213.
Also added tests, but was unable to build the tests locally as Windows is not current supported in the tests. A trivial attempt to add Windows support failed. I copied the added tests to a dependent project and validated that they pass. Was unable to run all tests.

## Option\<Breaking Changes\>

This would break code which passes integers with bit widths other than 32, and were getting lucky and working due to constant folding. I'd say that's getting lucky with UB.

## Checklist

### DIFF
--- a/src/values/ptr_value.rs
+++ b/src/values/ptr_value.rs
@@ -68,7 +68,17 @@ impl<'ctx> PointerValue<'ctx> {
 
     // REVIEW: Should this be on array value too?
     /// GEP is very likely to segfault if indexes are used incorrectly, and is therefore an unsafe function. Maybe we can change this in the future.
+    /// GEP indexes must be 32 bit integer values. Constant folding may result in other sizes working.
     pub unsafe fn const_gep(self, ordered_indexes: &[IntValue<'ctx>]) -> PointerValue<'ctx> {
+        if cfg!(debug_assertions) {
+            for (index, value) in ordered_indexes.iter().enumerate() {
+                let bit_width = value.get_type().get_bit_width();
+                if bit_width != 32 {
+                    panic!("Index #{} in ordered_indexes argument to const_gep call was a {} bit integer instead of 32 bit integer - gep indexes must be i32 values", index, bit_width);
+                }
+            }
+        }
+
         let mut index_values: Vec<LLVMValueRef> = ordered_indexes.iter()
                                                                  .map(|val| val.as_value_ref())
                                                                  .collect();
@@ -80,7 +90,17 @@ impl<'ctx> PointerValue<'ctx> {
     }
 
     /// GEP is very likely to segfault if indexes are used incorrectly, and is therefore an unsafe function. Maybe we can change this in the future.
+    /// GEP indexes must be 32 bit integer values. Constant folding may result in other sizes working.
     pub unsafe fn const_in_bounds_gep(self, ordered_indexes: &[IntValue<'ctx>]) -> PointerValue<'ctx> {
+        if cfg!(debug_assertions) {
+            for (index, value) in ordered_indexes.iter().enumerate() {
+                let bit_width = value.get_type().get_bit_width();
+                if bit_width != 32 {
+                    panic!("Index #{} in ordered_indexes argument to const_in_bounds_gep call was a {} bit integer instead of 32 bit integer - gep indexes must be i32 values", index, bit_width);
+                }
+            }
+        }
+
         let mut index_values: Vec<LLVMValueRef> = ordered_indexes.iter()
                                                                  .map(|val| val.as_value_ref())
                                                                  .collect();
--- a/tests/all/test_values.rs
+++ b/tests/all/test_values.rs
@@ -1203,3 +1203,30 @@ fn test_constant_expression() {
     assert!(expr.is_const());
     assert!(!expr.is_constant_int());
 }
+
+#[test]
+#[should_panic]
+fn test_const_gep_i64() {
+    let context = Context::create();
+
+    let i64_type = context.i64_type();
+    let i64_ptr_type = i64_type.ptr_type(AddressSpace::Generic);
+    let i64_ptr_null = i64_ptr_type.const_zero();
+    unsafe {
+        // i64 indexes are not permitted - should panic.
+        i64_ptr_null.const_gep(&[i64_type.const_zero()]);
+    }
+}
+
+#[test]
+fn test_const_gep_i32() {
+    let context = Context::create();
+
+    let i32_type = context.i32_type();
+    let i32_ptr_type = i32_type.ptr_type(AddressSpace::Generic);
+    let i32_ptr_null = i32_ptr_type.const_zero();
+    unsafe {
+        // i32 indexes are permitted.
+        i32_ptr_null.const_gep(&[i32_type.const_zero()]);
+    }
+}


src/values/ptr_value.rs
```diff
@@ -68,7 +68,17 @@ impl<'ctx> PointerValue<'ctx> {
 
     // REVIEW: Should this be on array value too?
     /// GEP is very likely to segfault if indexes are used incorrectly, and is therefore an unsafe function. Maybe we can change this in the future.
+    /// GEP indexes must be 32 bit integer values. Constant folding may result in other sizes working.
```
TheDan64, line 71: Let's remove the "Constant folding may result in other sizes working." comment. It's a pretty neat insight but doesn't help users in any way. You can move it to a regular, non-doc comment if you'd like.

src/values/ptr_value.rs
```diff
@@ -68,7 +68,17 @@ impl<'ctx> PointerValue<'ctx> {
...
     pub unsafe fn const_gep(self, ordered_indexes: &[IntValue<'ctx>]) -> PointerValue<'ctx> {
+        if cfg!(debug_assertions) {
```
TheDan64, line 73: Let's not make this just on debug builds. Imo, preventing UB is something we should always do.

src/values/ptr_value.rs
```diff
@@ -68,7 +68,17 @@ impl<'ctx> PointerValue<'ctx> {
...
+            for (index, value) in ordered_indexes.iter().enumerate() {
+                let bit_width = value.get_type().get_bit_width();
+                if bit_width != 32 {
+                    panic!("Index #{} in ordered_indexes argument to const_gep call was a {} bit integer instead of 32 bit integer - gep indexes must be i32 values", index, bit_width);
```
TheDan64, line 77: Let's return results instead. IE `Result<PointerValue<'ctx>, String>`. We generally don't panic on errors. The only exceptions I can think of are all very unlikely edge-cases.

TheDan64 (commented):
We probably want to do this for `Builder::build_gep` and `Builder::build_in_bounds_gep` too?

TheDan64:
@djozis Does my feedback make sense? Is there anything I can expand upon?

djozis:
@TheDan64 Sorry for the delayed response. Yes, I understand. I had trouble with this conceptually because I wanted my use of LLVM to be as efficient as possible, and considered this type of error path to be something likely encountered statically - ie. a code path encountered while first developing something invoking this, occurring always until the invocation is corrected, rather than a code path which could be triggered at runtime by providing bad inputs. For that reason, I preferred the panic approach, and to do it during debug only, since it keeps the efficiency optimal for release, but also prevents the pain I experienced.

I appreciate that's just an opinion though. It felt strange to contribute something that would make it less appealing for my use-case.

TheDan64:
> I wanted my use of LLVM to be as efficient as possible

Panicking isn't any more efficient because it still has to make a check and you could achieve the same results by unwrapping the output Result anyway

TheDan64:
Static typing would be ideal, but I don't think there's enough Rust support for a sufficient amount of it yet

djozis:
>Panicking isn't any more efficient

The efficiency I was referring to would come from keeping the check to debug mode only, as I was thinking it's a code path you'd encounter in a static way (either the program always hits that path or never does).

TheDan64:
> The efficiency I was referring to would come from keeping the check to debug mode only, as I was thinking it's a code path you'd encounter in a static way (either the program always hits that path or never does).

I don't really think this approach makes for very idiomatic Rust. Error cases should always be enumerated through `Result::Err`. Panics should be reserved for unrecoverable scenarios the vast majority of the time

PR #221, opened 2020-10-24
Fix #220

### LINKED ISSUE #220: `GlobalValue::get_section` can segfault if there is no section on the global
**Describe the Bug**
`LLVMGetSection` returns a null pointer if the global does not have a section set on it. This means that `GlobalValue::get_section` should return `Option<&CStr>` rather than `&CStr`, and check whether null was returned and return `None` in that case. Similarly, `GlobalValue::set_section` should take `Option<&CStr>` to allow you to remove the section by passing `None`.

**To Reproduce**
Get a handle to a `GlobalValue` with no section set. Call `get_section()` on it. Segfault.

**Expected Behavior**
Function should check whether `LLVMGetSection` returns null.

 - LLVM Version: 10
 - Inkwell Branch Used: master

Windows 10

**Additional Context**
N/A

dylanede:
Fixes issue #220.

## Description

Changed the `GlobalValue::get_section` and `GlobalValue::set_section` to deal with null pointers. Signatures have changed as follows:

`pub fn get_section(&self) -> &CStr` becomes `pub fn get_section(&self) -> Option<&CStr>`

`pub fn set_section(self, section: &str)` becomes `pub fn set_section(self, maybe_section: Option<&str>)`

Tests have been updated to reflect the changes in the API. New lines in `test_globals` have been added to test setting and getting `None`.

## Related Issue

#220

## How This Has Been Tested

Ran `cargo test --features llvm10-0`

## Option\<Breaking Changes\>

API for accessing sections on globals has changed. This is unavoidable as the original API was broken.

## Checklist

### DIFF
--- a/src/values/global_value.rs
+++ b/src/values/global_value.rs
@@ -12,6 +12,7 @@ use llvm_sys::LLVMUnnamedAddr;
 use llvm_sys::prelude::LLVMValueRef;
 
 use std::ffi::CStr;
+use std::ptr;
 
 use crate::{GlobalVisibility, ThreadLocalMode, DLLStorageClass};
 use crate::module::Linkage;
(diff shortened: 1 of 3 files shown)


TheDan64 (approved):
Solid change, thanks!

TheDan64:
The only issue is that it's failing on earlier LLVM versions: https://travis-ci.org/github/TheDan64/inkwell/builds/738553065

TheDan64 (changes requested):
Needs tests fixed on older LLVM versions

TheDan64:
Closing in favor of #333 which also adds this functionality to fn values

PR #223, opened 2020-11-18
Add print_to_string() to Any{Value,Type}

seanyoung:
Make it easier to print a BasicValueEnum/BasicTypeEnum

### DIFF
--- a/src/types/fn_type.rs
+++ b/src/types/fn_type.rs
@@ -7,9 +7,8 @@ use std::mem::forget;
 
 use crate::AddressSpace;
 use crate::context::ContextRef;
-use crate::support::LLVMString;
 use crate::types::traits::AsTypeRef;
-use crate::types::{PointerType, Type, BasicTypeEnum};
+use crate::types::{AnyType, PointerType, Type, BasicTypeEnum};
 
 /// A `FunctionType` is the type of a function variable.
 #[derive(PartialEq, Eq, Clone, Copy)]
@@ -153,11 +152,6 @@ impl<'ctx> FunctionType<'ctx> {
     /// ```
     pub fn get_context(self) -> ContextRef<'ctx> {
         self.fn_type.get_context()
-    }
-
-    /// Prints the definition of a `FunctionType` to a `LLVMString`.
-    pub fn print_to_string(self) -> LLVMString {
-        self.fn_type.print_to_string()
     }
 
     // See Type::print_to_stderr note on 5.0+ status
--- a/src/types/traits.rs
+++ b/src/types/traits.rs
@@ -6,6 +6,7 @@ use crate::AddressSpace;
 use crate::types::{IntType, FunctionType, FloatType, PointerType, StructType, ArrayType, VectorType, VoidType, Type};
 use crate::types::enums::{AnyTypeEnum, BasicTypeEnum};
 use crate::values::{IntMathValue, FloatMathValue, PointerMathValue, IntValue, FloatValue, PointerValue, VectorValue};
+use crate::support::LLVMString;
 
 // This is an ugly privacy hack so that Type can stay private to this module
 // and so that super traits using this trait will be not be implementable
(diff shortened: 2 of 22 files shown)


TheDan64 (changes requested):
This is a great idea; but I think it's the wrong approach. Instead, we should default impl `print_to_string` for the `Basic{Type,Value}` traits, then the types, values, and enums get them for free. And we should remove the existing individual methods on the type/values since they're superceeded by the trait.

`BasicType` [already impls](https://github.com/TheDan64/inkwell/blob/master/src/types/traits.rs#L124-L126) `ptr_type` btw, so `BasicTypeEnum` already gets it with a `BasicType` trait import. We should just remove the individual impls for that too.

Yay, deleting code!

seanyoung:
Aha, I missed that I needed the `BasicType` trait in scope to use ptr_type() on a `BasicValueEnum`.

Anyway I reworked it. Let me know what you think.

TheDan64 (changes requested):

src/types/traits.rs
```diff
@@ -22,13 +23,17 @@ macro_rules! trait_type_set {
...
 /// Represents any LLVM type.
 pub trait AnyType<'ctx>: AsTypeRef + Debug {
     /// Returns an `AnyTypeEnum` that represents the current type.
     fn as_any_type_enum(&self) -> AnyTypeEnum<'ctx> {
         AnyTypeEnum::new(self.as_type_ref())
     }
+
+    /// Prints the definition of a Type to a `LLVMString`.
+    fn print_to_string(&self) -> LLVMString {
+        self.print_to_string()
```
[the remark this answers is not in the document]
seanyoung: Yes, that's obviously broken. I've fixed that and even tested it this time.

src/types/fn_type.rs
```diff
@@ -9,7 +9,7 @@ use crate::AddressSpace;
 use crate::context::ContextRef;
 use crate::support::LLVMString;
 use crate::types::traits::AsTypeRef;
-use crate::types::{PointerType, Type, BasicTypeEnum};
+use crate::types::{AnyType, PointerType, Type, BasicTypeEnum};
```
TheDan64, line 12: Is this import necessary here?

src/values/array_value.rs
```diff
@@ -6,7 +6,7 @@ use std::fmt;
 
 use crate::support::LLVMString;
 use crate::types::ArrayType;
-use crate::values::traits::AsValueRef;
+use crate::values::traits::{AnyValue, AsValueRef};
```
TheDan64, line 9: Is this import necessary?

src/values/fn_value.rs
```diff
@@ -20,7 +20,7 @@ use crate::basic_block::BasicBlock;
 use crate::debug_info::DISubprogram;
 use crate::module::Linkage;
 use crate::support::{to_c_str, LLVMString};
-use crate::types::{FunctionType, PointerType};
+use crate::types::{AnyType, FunctionType, PointerType};
```
TheDan64, line 23: 1. Is this import necessary?
2. Should get rid of `FunctionValue::print_to_string`

TheDan64 (changes requested):
Last set of concerns - then we should be good to go!

Still needing removal:
* `FunctionValue::print_to_string`
* ~`PhiValue::print_to_string`~ jk you already did this one

(both of these impl `AnyValue`)

src/types/fn_type.rs
seanyoung: Yes, `impl fmt::Debug` uses self.print_to_strng()

src/values/array_value.rs
seanyoung: Yes, `impl fmt::Debug` calls self.print_to_string()

src/values/fn_value.rs
seanyoung: Yes, this import is necessary.

Note that `cargo clippy` in CI would tell you about unused imports.

seanyoung:
I've removed phi_value/fn_type print_to_string()

TheDan64 (approved):
Looks good. Thanks! Will merge after CI finishes

TheDan64:
@seanyoung looks like fn_value.rs needs AnyValue imported: https://travis-ci.org/github/TheDan64/inkwell/jobs/744760345

seanyoung:
CI succeeded now

TheDan64:
Thanks!

Issue #217, opened 2020-10-16
Add support for LLVM11


baszalmstra:
**Is your feature request related to a problem? Please describe.**
[Mun](https://github.com/mun-lang/mun) cant upgrade beyond LLVM8 due to a bug introduced in LLVM9. This has been fixed from LLVM11 onwards. Inkwell doesnt support LLVM11 yet.

**Describe the solution you'd like**
We would like to be able to use LLVM11 with Inkwell.

**Describe possible drawbacks to your solution**
Since this can be implemented with a feature flag, I dont think there will be drawbacks for other users. 

**Describe alternatives you've considered**
We could use the llvm-sys bindings directly, support for LLVM11 has already landed there. For obvious reasons Id rather not.

**Additional context**
Nope

ghaith:
I would like to help with this.
I took a look at the changes in LLVM 11 and there are no deprecations on the Frontend level at first glance. There might be some on the Backend, I don't know how much of the backend API is supported in inkwell. 
Do you have a plan on how to start this? I'm open to help with issues once you devide them. I can also probably take a look at the weekend at the required work.

TheDan64:
The first step is to add in llvm-sys 11 and see what breaks. Some relevant locations:

* https://github.com/TheDan64/inkwell/blob/master/Cargo.toml#L27
* https://github.com/TheDan64/inkwell/blob/master/Cargo.toml#L43
* https://github.com/TheDan64/inkwell/blob/master/Cargo.toml#L98
* https://github.com/TheDan64/inkwell/blob/master/src/lib.rs#L65
* https://github.com/TheDan64/inkwell/blob/master/internal_macros/src/lib.rs#L23

From there fix any breakage. There are a couple locations that always break and just need a `cfg` to include `llvm11-0`, though I can't point to them by memory. If you run into any hard issues, feel free to bring it up here

PR #224, opened 2020-11-21
Add support for LLVM 11

ghaith:
## Description
Added support for an LLVM 11 future.
Updated the debug builder API to support the new parameters
Fixed some failing tests
Updated travis config 

## Related Issue
#217 

## How This Has Been Tested
Ran tests for the older features and the new feature
Did not add new tests as no new functionality was added
**2 Tests are still failing on my machine**, one for enums and one for alignment.

## Checklist

### DIFF
--- a/src/debug_info.rs
+++ b/src/debug_info.rs
@@ -162,6 +162,7 @@ pub trait AsDIScope<'ctx> {
 }
 
 impl<'ctx> DebugInfoBuilder<'ctx> {
+    #[llvm_versions(3.6..=10.0)]
     pub(crate) fn new(
         module: &Module,
         allow_unresolved: bool,
@@ -210,6 +211,59 @@ impl<'ctx> DebugInfoBuilder<'ctx> {
         (builder, cu)
     }
 
+    #[llvm_versions(11.0)]
+    pub(crate) fn new(
+        module: &Module,
+        allow_unresolved: bool,
+        language: DWARFSourceLanguage,
+        filename: &str,
+        directory: &str,
+        producer: &str,
+        is_optimized: bool,
+        flags: &str,
+        runtime_ver: libc::c_uint,
+        split_name: &str,
+        kind: DWARFEmissionKind,
+        dwo_id: libc::c_uint,
+        split_debug_inlining: bool,
+        debug_info_for_profiling: bool,
+        sysroot: &str,
+        sdk: &str,
+    ) -> (Self, DICompileUnit<'ctx>) {
+        let builder = unsafe {
+            if allow_unresolved {
+                LLVMCreateDIBuilder(module.module.get())
+            } else {
+                LLVMCreateDIBuilderDisallowUnresolved(module.module.get())
+            }
+        };
+
+        let builder = DebugInfoBuilder {
+            builder,
+            _marker: PhantomData,
+        };
+
+        let file = builder.create_file(filename, directory);
+
+        let cu = builder.create_compile_unit(
+            language,
+            file,
+            producer,
+            is_optimized,
+            flags,
+            runtime_ver,
+            split_name,
+            kind,
+            dwo_id,
+            split_debug_inlining,
+            debug_info_for_profiling,
+            sysroot,
+            sdk
+        );
+
+        (builder, cu)
+    }
+
     /// A DICompileUnit provides an anchor for all debugging information generated during this instance of compilation.
     ///
     /// * `language` - Source programming language
@@ -223,6 +277,7 @@ impl<'ctx> DebugInfoBuilder<'ctx> {
     /// * `dwo_id` - The DWOId if this is a split skeleton compile unit.
     /// * `split_debug_inlining` - Whether to emit inline debug info.
     /// * `debug_info_for_profiling` - Whether to emit extra debug info for profile collection.
+    #[llvm_versions(3.6..=10.0)]
     fn create_compile_unit(
         &self,
         language: DWARFSourceLanguage,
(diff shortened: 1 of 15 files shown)


ghaith:
I created this as a draft since 2 tests still fail.
One test deals with alignment values `test_instruction_values::test_mem_instructions` , where we were expecting them to be 0, but they are 4 in LLVM11, I did not find any reference to this change here: https://releases.llvm.org/11.0.0/docs/ReleaseNotes.html which is why i did not yet fix the test.
The other test has to do with the enum values for metadata `test_attributes::test_enum_attribute_kinds`. The [LLVM Documentation](https://llvm.org/doxygen/group__LLVMCCoreContext.html#ga4bd62e3a1b94b92e32d892c8f89dea1c) says these values are not maintained in the C API and can change at any time. So I'm not sure how we can test them reliably or why we should.

TheDan64:
If you see enum attribute kind numbers that changed, you can remove just the ones that are no longer the same and keep the working ones

ghaith:
> If you see enum attribute kind numbers that changed, you can remove just the ones that are no longer the same and keep the working ones

all tested attributes changed, the test only has the non-existing attributes now.

ghaith:
I pushed the updated tests, but I'm still having some issues in the build.
If I run all tests (using cargo test) then the test crash with : 
`Caused by:
  process didn't exit successfully: `/usr/local/src/target/debug/deps/all-4bbc9f42c9e86587` (signal: 11, SIGSEGV: invalid memory reference)
`
If I run individual tests they are fine.

I'm also not happy with all the #[cfg(feature...] I had to introduce since they need to be updated again with each new llvm version. Do you have suggestions for that? Should I just duplicate the methods and use llvm_version? Or do you know if llvm_version is usable inside parameter lists? I haven't worked much with features in rust.

ghaith:
I found the issue with the segfault, it was due to a missing feature flag in the test_globals section.
I also updated the travis setup.
I'll set the PR as ready from my side, tell me if I should do more cleanups there.

TheDan64:
Sorry for the delay, been mostly AFK for the holidays. Should be able to take a look soon

TheDan64:
> newer symver versions no longer compile with rust 1.42

What does this mean exactly??

src/debug_info.rs
```diff
@@ -210,6 +211,59 @@ impl<'ctx> DebugInfoBuilder<'ctx> {
...
+        is_optimized: bool,
+        flags: &str,
+        runtime_ver: libc::c_uint,
+        split_name: &str,
+        kind: DWARFEmissionKind,
+        dwo_id: libc::c_uint,
+        split_debug_inlining: bool,
+        debug_info_for_profiling: bool,
+        sysroot: &str,
+        sdk: &str,
```
TheDan64, line 231: Rather than duplicating these functions we should just add `llvm_version` to the params

src/types/enums.rs
```diff
@@ -103,13 +103,17 @@ impl<'ctx> AnyTypeEnum<'ctx> {
             LLVMTypeKind::LLVMX86_FP80TypeKind |
             LLVMTypeKind::LLVMFP128TypeKind |
             LLVMTypeKind::LLVMPPC_FP128TypeKind => AnyTypeEnum::FloatType(FloatType::new(type_)),
+            #[cfg(feature = "llvm11-0")]
+            LLVMTypeKind::LLVMBFloatTypeKind => AnyTypeEnum::FloatType(FloatType::new(type_)),
```
TheDan64, line 107: Any idea what this new type is for?

src/values/instruction_value.rs
```diff
@@ -277,6 +278,22 @@ impl<'ctx> InstructionValue<'ctx> {
         }
         Ok(unsafe { LLVMSetAlignment(self.as_value_ref(), alignment) })
     }
+    
+    // SubTypes: Only apply to memory access and alloca instructions
+    /// Sets alignment on a memory access instruction or alloca.
+    #[llvm_versions(11.0..=latest)]
```
TheDan64, line 284: Why was this function split out?

TheDan64 (commented):
Looks pretty good overall but have some thoughts. Also really confused about the version change.

ghaith:
> > newer symver versions no longer compile with rust 1.42
> 
> What does this mean exactly?? I don't see any reason to have to update to a newer rust?

The newest semver-parser package at the time, on which llvm-sys depends did not build with rust versions < 1.49.0, this is due to them using a non build-in inner attribute, they seem to have that fixed in 0.10.2 so i will revert this back

```
 Compiling semver-parser v0.10.1
error[E0658]: non-builtin inner attributes are unstable
 --> /home/ghaith/.cargo/registry/src/github.com-1ecc6299db9ec823/semver-parser-0.10.1/src/generated.rs:4:1
  |
4 | #![rustfmt::skip]
  | ^^^^^^^^^^^^^^^^^
  |
  = note: for more information, see https://github.com/rust-lang/rust/issues/54726


```

src/values/instruction_value.rs
ghaith: LLVM11 introduced some alignment guards, I thought we should not force <11 to use them, but I will move this to a feature check in the same function

src/types/enums.rs
ghaith: BFloat : https://releases.llvm.org/11.0.0/docs/ReleaseNotes.html#changes-to-the-aarch64-backend
Looks like an Aarch64 (64 bit ARM) specific field

The scalable vector looks like a new generic IR type: https://releases.llvm.org/11.0.0/docs/ReleaseNotes.html#changes-to-the-llvm-ir

src/debug_info.rs
ghaith: I had to use #[cfg(feature(..)] since custom attributes don't work on parameters.
I also still have to duplicate the function call since attributes on parameter calls are also experimental

TheDan64:
Ohh, thanks for clarifying!

src/values/instruction_value.rs
TheDan64: So older versions allowed alignment 0 but 11 no longer does?

ghaith: I could not find it in the release notes, so I looked for changes in the code and there has been some work on alignments.
Mainly _most_ of the `setAlignment` methods in [`llvm/IR/Core.cpp`](https://github.com/llvm/llvm-project/blob/release/11.x/llvm/lib/IR/Core.cpp) will now call `Align()` instead of `MaybeAlign()` which fails on 0 values. The only exception is `GlobalValue`. 
I can change this back to remove the check but it means setting alignment to 0 will panic instead of returning the Result<>. 
Since this is on instruction_value and instruction will check for 0, it might be worth keeping the check there.
The changes on LLVM for that are : 
https://reviews.llvm.org/D77454
https://reviews.llvm.org/D79968
https://reviews.llvm.org/D80044

TheDan64: Great catch. We should just add a feature check in the same function I think? Unless you feel strongly otherwise

TheDan64: We really don't want LLVM to hit an assertion under us because that's bad news

TheDan64 (approved):
Looks good. Left a comment on the alignment comment thread. If it's all set then we can merge

src/values/instruction_value.rs
ghaith: I already changed it to a feature check in a previous commit : https://github.com/TheDan64/inkwell/pull/224/files#diff-8e2e5aa572ea6e78c6571db5bae3c54a30f55b202b9c0995d76e145013ea96c1

TheDan64:
Thanks! Sorry it took so long.

PR #229, opened 2021-01-26
Add BasicTypeEnum::const_zero().

nlewycky:
## Description

Every BasicType can take on a zero value. Add BasicTypeEnum::const_zero() that returns a BasicValueEnum with the zero in it, of the matching basic type.

## Related Issue

n/a

## How This Has Been Tested

doc test

## Checklist


TheDan64:
This look like a good change but the wrong approach - we should add the method to the `BasicType` trait, and remove the individual methods since the trait will provide it for them

PR #247, opened 2021-04-30
Add metadata support

syrusakbary:
## Description

Add metadata support

## Checklist

- [ ] I have read the [Contributing Guide](https://github.com/TheDan64/inkwell/blob/master/.github/CONTRIBUTING.md)


src/types/enums.rs
```diff
@@ -2,7 +2,7 @@ use llvm_sys::core::LLVMGetTypeKind;
 use llvm_sys::LLVMTypeKind;
 use llvm_sys::prelude::LLVMTypeRef;
 
-use crate::types::{IntType, VoidType, FunctionType, PointerType, VectorType, ArrayType, StructType, FloatType};
+use crate::types::{IntType, VoidType, FunctionType, PointerType, VectorType, ArrayType, StructType, FloatType, MetadataType};
```
syrusakbary, line 5:
```suggestion
use crate::types::{IntType, VoidType, FunctionType, PointerType, VectorType, ArrayType, StructType, FloatType};
#[llvm_versions(6.0..=latest)]
use crate::types::MetadataType;
```

src/types/enums.rs
```diff
@@ -80,6 +82,8 @@ enum_type_set! {
         FloatType,
         /// An integer type.
         IntType,
+        /// A metadata type.
+        MetadataType,
```
syrusakbary, line 86:
```suggestion
        #[llvm_versions(6.0..=latest)]
        /// A metadata type.
        MetadataType,
```

TheDan64:
I'm not so confident on metadata being a basic value/type, I think it would break a lot of code. For starters, is it valid to pass metadata in as function parameters or return types?

TheDan64:
> As far as I know, metadata is something only intrinsics can take as argument or return.

I believe this is true - but we need tests to ensure it works as we expect and LLVM assertions are not triggered.

> Can you clarify what is a basic type/value exactly in Inkwell? I did some research but did not find something to answer my question so I went with trying to remove the panic! or unreachable! and go from here.

`BasicValue` and `BasicType` are inkwell traits that were implemented for `MetadataValue` and `MetadataType`, respectively, as part of this PR.

To accept this PR I need to see a few things:

1. All unit tests pass (currently older versions are failing - see the travis build. Don't think it'd be difficult to fix)
2. PR is rebased to fix conflicts (apologies - I had to make some big changes recently)
3. Tests are added that show that `MetadataValue` and `MetadataType` are valid `BasicValue` and `BasicType` implementors and don't trigger LLVM assertions. Here's a non-exhaustive list of functions to call them on:
    - `add_basic_value_flag`
    - `build_indirect_branch`
    - `build_insert_value`
    - `build_select`
    - `const_select`
    - `set_operand`
    - `const_insert_value`
    - `const_insert_element`
    - `const_select`
    - `const_vector`
    - `add_global`
    - `build_phi`
    - `build_alloca`
    - `build_array_alloca`
    - `build_malloc`
    - `build_array_malloc`
    - `build_cast`
    - `build_va_arg`

Explained another way - all existing implementors of `BasicValue`, `BasicType` (`FloatValue`, `IntType`, etc), should be valid for those generics above (and maybe more I missed some more). So if we want to add `MetadataValue` & `MetadataType` to that set of implementors, they also need to be valid to be passed in to the above functions. If they break in some way(test failure or LLVM assertion), then we need to reconsider the approach of this PR

jubianchi:
@TheDan64 thanks for all the clarifications.

I will do what you requested and take care of fixing tests and adding new ones (already tried but failed, will try more 😄 )

src/values/instruction_value.rs
```diff
@@ -665,6 +665,8 @@ impl<'ctx> InstructionValue<'ctx> {
     /// Determines whether or not this `Instruction` has any associated metadata
     /// `kind_id`.
     pub fn set_metadata(self, metadata: MetadataValue<'ctx>, kind_id: u32) {
+        assert!(metadata.is_node());
```
TheDan64, line 668: You mentioned `add_global_metadata` can't be provided a string; but why was this added here? Does a string also trigger an LLVM assertion here?

tests/all/test_values.rs
```diff
@@ -443,31 +443,34 @@ fn test_metadata() {
...
     assert_eq!(md_node.get_string_value(), None);
-    assert_eq!(node_values.len(), 2);
+    assert_eq!(node_values.len(), 4);
     assert_eq!(node_values[0].into_int_value(), bool_val);
     assert_eq!(node_values[1].into_float_value(), f32_val);
+    assert_eq!(node_values[2].into_metadata_value().get_string_value(), md_string.get_string_value());
+    assert!(node_values[3].into_metadata_value().is_node());
 
-    module.add_global_metadata("my_md", &md_string);
+    //module.add_global_metadata("my_md", &md_string);
```
TheDan64, line 458: `add_global_metadata` should return a result rather than asserting, and this then can remain a test assertion of Err

src/values/instruction_value.rs
TheDan64: And if so, this should become a result

src/module.rs
```diff
@@ -836,6 +836,8 @@ impl<'ctx> Module<'ctx> {
     /// assert_eq!(md_1[1].into_float_value(), f32_val);
     /// ```
     pub fn add_global_metadata(&self, key: &str, metadata: &MetadataValue<'ctx>) {
+        assert!(metadata.is_node());
```
TheDan64, line 839: Should be a result rather than asserting

src/values/instruction_value.rs
syrusakbary: Rather than being a result, can we make it an `Option`? It will make things a bit easier.

If we prefer a result, what the Error type should be?

src/module.rs
syrusakbary: Rather than being a result, can we make it an `Option`? It will make things a bit easier.

If we prefer a result, what the Error type should be?

TheDan64: A Result makes more sense because it's reporting an error condition. But you can use the [ok](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) method to turn it into an option for your purposes and discard the error. Error type would be a 'static string for now. We'll eventually have a proper enum

syrusakbary: Fixed!

tests/all/test_values.rs
syrusakbary: Fixed

src/values/instruction_value.rs
syrusakbary: Done

TheDan64 (approved):

TheDan64:
Looks like some tests still need to be fixed before this can be merged

PR #252, opened 2021-06-05
Invoke & some friends

folkertdev:
## Description

Adds support for the `invoke` instruction, and some support for building landing pads and resuming an exception. This does not cover the full exception functionality, but unless you're building a c++ compiler this should provide most useful features.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -243,20 +243,22 @@ impl<'ctx> Builder<'ctx> {
         }
     }
 
-    pub fn build_cleanup_landing_pad(
-        &self,
-        ty: &dyn BasicType<'ctx>,
-        value: &dyn BasicValue<'ctx>,
-        name: &str,
-    ) -> BasicValueEnum<'ctx> {
+    pub fn build_cleanup_landing_pad<T>(&self, exception_type: T, name: &str) -> BasicValueEnum<'ctx>
+    where
+        T: BasicType<'ctx>,
+    {
         let c_string = to_c_str(name);
         let num_clauses = 0u32;
+
+        // SAFETY: the personality function will not be used
+        // because there are no clauses to match
+        let personality_function : LLVMValueRef = unsafe { 0 as *mut _ };
 
         let value = unsafe {
             LLVMBuildLandingPad(
                 self.builder,
-                ty.as_type_ref(),
-                value.as_value_ref(),
+                exception_type.as_type_ref(),
+                personality_function,
                 num_clauses,
                 c_string.as_ptr(),
             )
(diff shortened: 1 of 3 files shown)


src/builder.rs
```diff
@@ -178,6 +175,145 @@ impl<'ctx> Builder<'ctx> {
...
+
+        unsafe {
+            CallSiteValue::new(value)
+        }
+    }
+
+    pub fn build_cleanup_landing_pad(
+        &self,
+        ty: &dyn BasicType<'ctx>,
+        value: &dyn BasicValue<'ctx>,
```
TheDan64, line 249: Any reason why `ty` and `value` aren't generics?

src/builder.rs
```diff
@@ -178,6 +175,145 @@ impl<'ctx> Builder<'ctx> {
...
+
+        unsafe {
+            BasicValueEnum::new(value)
+        }
+    }
+
+    pub fn build_catch_all_landing_pad(
+        &self,
+        ty: &dyn BasicType<'ctx>,
+        value: &dyn BasicValue<'ctx>,
```
TheDan64, line 277: Same Q re: generics

src/builder.rs
```diff
@@ -178,6 +175,145 @@ impl<'ctx> Builder<'ctx> {
...
+            LLVMAddClause(value, null.as_value_ref());
+        };
+
+        unsafe {
+            BasicValueEnum::new(value)
+        }
+    }
+
+    /// Resume propagation of an existing (in-flight) exception whose unwinding was interrupted with a landingpad instruction.
+    pub fn build_resume(&self, value: &dyn BasicValue<'ctx>) -> InstructionValue<'ctx> {
```
TheDan64, line 309: Generic?

TheDan64 (commented):
This looks pretty good; just have a Q on dyn traits that I think could be generics. I'd like to see tests for all of these new methods. And turning the asserts into Results would be ideal, too, and easier to test.

src/builder.rs
folkertdev: no reason, these were the particular types we needed in our code base. I'll use generics here.

folkertdev:
The `assert` mirrors the one in the `build_call` source. Invoke is semantically a call and then a branch on the result to see if there was an exception (it's implemented quite differently for performance reasons). So it would make sense for `build_call` and `build_invoke` to be consistent. 

If you do think this is the right time to turn that assert into a result (for both `build_call` and `build_invoke`, I'd like to suggest to use stronger types to constrain the input. Right now the function argument is 

```
        F: Into<FunctionOrPointerValue<'ctx>>,
```
we could instead add something like a `CallableValue` and `impl From<FunctionValue> for CallableValue` but crucially `impl TryFrom<PointerValue> for CallableValue`. 

Then, if we instead specify

```
        F: Into<CallableValue<'ctx>>,
```

The most common change, calling a function value, remains unchanged in user code. Only the actually dangerous code, treating a pointer value as a pointer to a function, requires changes.

TheDan64:
`Into<CallableValue<'ctx>>` won't work with `impl TryFrom<PointerValue> for CallableValue` though

folkertdev:
correct, and it shouldn't because not all pointer values point to functions. My point is that this variant, as opposed to returning a result on failure

- does not require user code changes in the FunctionValue case
- requires the user handles the "pointer points to a non-function" case explicitly up-front

src/builder.rs
```diff
@@ -178,6 +175,284 @@ impl<'ctx> Builder<'ctx> {
...
+    ///
+    /// The personality function determines how an exception is handled.
+    /// This example shows how to use the C++ personality function:
+    ///
+    /// ```no_run
+    /// let context = Context::create();
+    /// let module = context.create_module("sum");
+    /// let builder = context.create_builder();
+    ///
+    /// let f32_type = context.f32_type();
```
folkertdev, line 259: so, should this full example be here? it's big, but also I don't see a better place to put it.

src/builder.rs
```diff
@@ -178,6 +175,284 @@ impl<'ctx> Builder<'ctx> {
...
+                exception_type.as_type_ref(),
+                personality_function.as_value_ref(),
+                num_clauses,
+                c_string.as_ptr(),
+            )
+        };
+
+        // we will add one clause, a null pointer of the i8 type
+        // this is how c++ encodes the `catch(...)` case.
+        let i8_type = unsafe { crate::types::IntType::new(llvm_sys::core::LLVMInt8Type()) };
```
folkertdev, line 432: is this OK?

src/builder.rs
TheDan64: This is fine if you think it's helpful

src/builder.rs
TheDan64: At minimum, this is incorrect because `LLVMInt8Type` assumes the global context

TheDan64: Isn't it possible to do something like:

```rust
exeception_type.get_context().i8_type();
```

I know the `get_context` methods have not so great lifetimes associated with them (bug), but I think that won't be a problem here because we don't save it anywhere

folkertdev: well it works in the test... Anyway in this case you need to construct a null pointer of any type really. Is there a way to do that?

TheDan64: Also, the comment doesn't make it clear why this needs to be done. Something about exception catching was all I got out of it. But why does it need one clause?

TheDan64: or `personality_function.get_context()`

folkertdev: yes that worked!

folkertdev:
opted to expose the full `build_landing_pad` after actually figuring out how it works. I also give a bunch of code samples for setting up landing pads, because that information is very hard to find.

TheDan64 (approved):
Awesome, thanks for all of your hard work; esp around documentation and tests!

Issue #249, opened 2021-05-08
Support For LLVM version 12.0.


amm834:
**Describe the solution you'd like**

LLVM is upgraded to Version 12.0.
Please upgrade this.

https://releases.llvm.org/

TheDan64:
This is blocked by llvm-sys not having v12 bindings yet

TheDan64:
Correct. Inkwell is dependant on the `llvm-sys` crate, which has not yet released bindings for LLVM 12

Raspberry1111:
It appears that there is [now a version of `llvm-sys`](https://crates.io/crates/llvm-sys/120.0.0) with LLVM 12 bindings

seanyoung:
Now that llvm 12 support has been merged, can we have a crates.io release please? Something like 0.1.0-beta3.

I have tested llvm 12 support with [Solang](https://github.com/hyperledger-labs/solang) and it all looks fine to me.

TheDan64:
I'd like to wait until we get a couple more of the recent open PRs merged

PR #253, opened 2021-06-05
Llvm 12

folkertdev:
## Description

Add support for llvm 12

## How This Has Been Tested

ran test suite, worked on my machine

## Checklist

### DIFF
--- a/src/passes.rs
+++ b/src/passes.rs
@@ -376,7 +376,6 @@ impl<T: PassManagerSubType> PassManager<T> {
     /// In LLVM 12 and later, this instruction is replaced by the
     /// [`add_instruction_simplify_pass`].
     #[llvm_versions(3.6..=11.0)]
-    #[doc(alias = "add_instruction_simplify_pass")]
     pub fn add_ip_constant_propagation_pass(&self) {
         unsafe {
             LLVMAddIPConstantPropagationPass(self.pass_manager)
@@ -939,7 +938,6 @@ impl<T: PassManagerSubType> PassManager<T> {
     /// In LLVM 12 and later, this instruction is replaced by the
     /// [`add_instruction_simplify_pass`].
     #[llvm_versions(3.6..=11.0)]
-    #[doc(alias = "add_instruction_simplify_pass")]
     pub fn add_constant_propagation_pass(&self) {
         unsafe {
             LLVMAddConstantPropagationPass(self.pass_manager)
@@ -963,7 +961,6 @@ impl<T: PassManagerSubType> PassManager<T> {
     /// NOTE: this pass has a habit of making definitions be dead. It is a good idea to
     /// run a Dead Instruction Elimination pass sometime after running this pass.
     #[llvm_versions(12.0..=latest)]
-    #[doc(alias = "add_constant_propagation_pass")]
     pub fn add_instruction_simplify_pass(&self) {
         unsafe {
             LLVMAddInstructionSimplifyPass(self.pass_manager)
--- a/src/passes.rs
+++ b/src/passes.rs
@@ -1,12 +1,19 @@
 use llvm_sys::core::{LLVMDisposePassManager, LLVMInitializeFunctionPassManager, LLVMFinalizeFunctionPassManager, LLVMRunFunctionPassManager, LLVMRunPassManager, LLVMCreatePassManager, LLVMCreateFunctionPassManagerForModule, LLVMGetGlobalPassRegistry};
 use llvm_sys::initialization::{LLVMInitializeCore, LLVMInitializeTransformUtils, LLVMInitializeScalarOpts, LLVMInitializeObjCARCOpts, LLVMInitializeVectorization, LLVMInitializeInstCombine, LLVMInitializeIPO, LLVMInitializeInstrumentation, LLVMInitializeAnalysis, LLVMInitializeIPA, LLVMInitializeCodeGen, LLVMInitializeTarget};
 use llvm_sys::prelude::{LLVMPassManagerRef, LLVMPassRegistryRef};
-use llvm_sys::transforms::ipo::{LLVMAddArgumentPromotionPass, LLVMAddConstantMergePass, LLVMAddDeadArgEliminationPass, LLVMAddFunctionAttrsPass, LLVMAddFunctionInliningPass, LLVMAddAlwaysInlinerPass, LLVMAddGlobalDCEPass, LLVMAddGlobalOptimizerPass, LLVMAddIPConstantPropagationPass, LLVMAddIPSCCPPass, LLVMAddInternalizePass, LLVMAddStripDeadPrototypesPass, LLVMAddPruneEHPass, LLVMAddStripSymbolsPass};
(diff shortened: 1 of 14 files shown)


src/passes.rs
```diff
@@ -362,6 +372,11 @@ impl<T: PassManagerSubType> PassManager<T> {
     /// makes arguments dead, but does not remove them. The existing
     /// dead argument elimination pass should be run after this to
     /// clean up the mess.
+    ///
+    /// In LLVM 12 and later, this instruction is replaced by the
+    /// [`add_instruction_simplify_pass`].
+    #[llvm_versions(3.6..=11.0)]
+    #[doc(alias = "add_instruction_simplify_pass")]
```
folkertdev, line 379: doc aliases ensure that when searching for the old (now-deprectated) pass you also find the new one, and vice versa.

src/passes.rs
```diff
@@ -920,12 +935,41 @@ impl<T: PassManagerSubType> PassManager<T> {
...
+    /// ```
+    ///
+    /// becomes
+    ///
+    /// ```ir
+    /// i32 3
+    /// ```
+    ///
+    /// NOTE: this pass has a habit of making definitions be dead. It is a good idea to
+    /// run a Dead Instruction Elimination pass sometime after running this pass.
```
folkertdev, line 964: based on the LLVM docs/source this comment is still accurate

src/values/metadata_value.rs
```diff
@@ -36,6 +36,8 @@ pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 26;
 pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 28;
 #[cfg(any(feature = "llvm10-0", feature = "llvm11-0"))]
 pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 30;
+#[cfg(feature = "llvm12-0")]
+pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 31;
```
folkertdev, line 40: this number is what the tests told me it should be.

folkertdev (commented):
Left some notes. 

Also I'm not sure what the CI failure means and how I can fix it.

src/passes.rs
TheDan64: This is cool, I didn't know about doc alias. But I'm not sure it matters or will even work - this function will be compiled out of llvm 12 builds (which will include our publicly generated/hosted docs)

src/passes.rs
TheDan64: Cool, ty for details docs!

src/types/enums.rs
```diff
@@ -108,10 +108,12 @@ impl<'ctx> AnyTypeEnum<'ctx> {
             LLVMTypeKind::LLVMArrayTypeKind => AnyTypeEnum::ArrayType(ArrayType::new(type_)),
             LLVMTypeKind::LLVMPointerTypeKind => AnyTypeEnum::PointerType(PointerType::new(type_)),
             LLVMTypeKind::LLVMVectorTypeKind => AnyTypeEnum::VectorType(VectorType::new(type_)),
-            #[cfg(feature = "llvm11-0")]
+            #[cfg(any(feature = "llvm11-0", feature = "llvm12-0"))]
             LLVMTypeKind::LLVMScalableVectorTypeKind => AnyTypeEnum::VectorType(VectorType::new(type_)),
             LLVMTypeKind::LLVMMetadataTypeKind => panic!("FIXME: Unsupported type: Metadata"),
             LLVMTypeKind::LLVMX86_MMXTypeKind => panic!("FIXME: Unsupported type: MMX"),
+            #[cfg(feature = "llvm12-0")]
+            LLVMTypeKind::LLVMX86_AMXTypeKind => panic!("FIXME: Unsupported type: AMX"),
```
TheDan64, line 116: Do you have any insight or links to docs on what AMX is?

TheDan64: Could add it to a comment here

src/passes.rs
folkertdev: I verified locally that when building with `--features="llvm12-0"` when I search for `add_constant_propagation_pass` I see the alias for the new version. It has to be an exact match though, even `add_constant_propagation_pas` will not find it. But maybe this will improve further in the future.

src/types/enums.rs
folkertdev: I just found this https://llvm.org/docs/LangRef.html#x86-amx-type seems very similar to the MMX so I treated it similarly here

TheDan64 (commented):
This looks in pretty good shape. Travis test need to be passing on all versions, obviously. I also don't see a travis build for LLVM 12, so you'll need to add one to the .travis.yml file

If you check the travis builds it appears the failures are due to doc alias:

```rust
   --> src/passes.rs:379:5
    |
379 |     #[doc(alias = "add_instruction_simplify_pass")]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #50146 <https://github.com/rust-lang/rust/issues/50146> for more information

error[E0658]: `#[doc(alias)]` is experimental
   --> src/passes.rs:942:5
    |
942 |     #[doc(alias = "add_instruction_simplify_pass")]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #50146 <https://github.com/rust-lang/rust/issues/50146> for more information
```

We should probably just remove it as I don't think it'll be useful in practice due to the older methods being compiled out (though I do think it was a good idea!). Our MSRV is currently 1.42

folkertdev:
CI now mostly passes but the docs fail to build, giving the error 
```
Installing APT Packages
0.53s$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install libelf-dev lib32z1-dev libedit-dev libdw-dev binutils-dev libiberty-dev libssl-dev llvm-12-dev
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package llvm-12-dev
apt-get.diagnostics
apt-get install failed
The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install libelf-dev lib32z1-dev libedit-dev libdw-dev binutils-dev libiberty-dev libssl-dev llvm-12-dev" failed and exited with 100 during .
```

This package does exist, I wonder if maybe we need to use a more recent ubuntu version than bionic?

TheDan64:
Feel free to try a newer version than bionic if the required packages exist there

folkertdev:
so apparently the documentation build also needs `- libclang-common-12-dev` now? not sure how that works, but I'm guessing since the "normal" tests need it it will already be installed on the CI machine, so there is no downside to including it when building the docs.

TheDan64:
Could you update the req and usage sections of the readme?

TheDan64:
Otherwise I think this is good to be merged

TheDan64 (approved):
Thanks! Closes #249

PR #254, opened 2021-06-06
Add CallableValue

folkertdev:
Adds an opaque type that represents a value that can be called by the `build_call` instruction.

## Description

The main benefit of this change is that `build_call` will no longer panic. Previously it paniced when it was passed a pointer value that pointed to a non-function. A `CallableValue` is guaranteed to refer to a function (either by name or by pointer), so this panic is no longer necessary.

## Breaking Changes

It is no longer possible to pass a pointer value to `build_call` (A `FunctionValue` can still be passed in as before). A pointer must be converted explicitly:
```rust
// before
builder.build_call(callable_value, &[], "call");

// after
use std::convert::TryFrom;
let callable_value = CallableValue::try_from(i8_ptr_param).unwrap();
builder.build_call(callable_value, &[], "call");
```

But in practice I expect by-pointer calls to be rarer than by-name calls, and this is an error case that must be handled. With this approach it only needs to be handled in the case that could actually fail: a call by pointer.

## Checklist


folkertdev:
Wanted to work this out. I quite like this from an API perpective. One other alternative is to deprecate `build_call`, and instead have

```rust
pub fn build_call_by_name(&self, function: FunctionValue<'ctx>, args: &[BasicValueEnum<'ctx>], name: &str) -> CallSiteValue<'ctx>;

pub fn build_call_by_pointer(&self, function: PointerValue<'ctx>, args: &[BasicValueEnum<'ctx>], name: &str) -> Result<CallSiteValue<'ctx>, ()>
```

folkertdev:
actually, I really like that idea of having two functions, one for function values, one for pointer values better than this `CallableValue`  abstraction. Minimal extra stuff to learn (no Into, TryInto, etc), and the relevant documentation is exactly where it is required. Thoughts?

src/values/callable_value.rs
```diff
@@ -0,0 +1,127 @@
...
+///
+/// let ret_val = builder.build_call(callable_value, &[i32_arg], "call")
+///     .try_as_basic_value()
+///     .left()
+///     .unwrap();
+///
+/// builder.build_return(Some(&ret_val));
+/// ```
+#[derive(Debug)]
+pub struct CallableValue<'ctx>(FunctionOrPointerValue<'ctx>);
```
TheDan64, line 79: Is `FunctionOrPointerValue` used anywhere else? It kinda seems like `CallableValue` should just take its place

TheDan64: Nvm, I see the type alias below

src/values/callable_value.rs
```diff
@@ -0,0 +1,127 @@
...
+
+type FunctionOrPointerValue<'ctx> = Either<FunctionValue<'ctx>, PointerValue<'ctx>>;
+
+impl<'ctx> crate::values::traits::AsValueRef for CallableValue<'ctx> {
```
TheDan64, line 83: Nit: `crate::values::traits::AsValueRef` should be imported

src/values/callable_value.rs
```diff
@@ -0,0 +1,127 @@
...
+        use either::Either::*;
+
+        match self.0 {
+            Left(function) => function.as_value_ref(),
+            Right(pointer) => pointer.as_value_ref(),
+        }
+    }
+}
+
+impl<'ctx> crate::values::AnyValue<'ctx> for CallableValue<'ctx> {}
```
TheDan64, line 94: import

TheDan64:
I don't really like having two functions because since both are doing the same thing - building calls - we're just trying to "sanitize" the input

folkertdev:
I'd argue that the input validation is exactly why they are not doing the same thing.  To me it's better to be explicit about that, and having two function seems like the simplest way to do it.
```
pub fn build_call(&self, function: FunctionValue<'ctx>, args: &[BasicValueEnum<'ctx>], name: &str) -> CallSiteValue<'ctx>;

pub fn build_call_pointer(&self, function: PointerValue<'ctx>, args: &[BasicValueEnum<'ctx>], name: &str) -> Result<CallSiteValue<'ctx>, ()>
```

Anyway, your call. I imported some previously-qualified symbols, and removed the FunctionOrPointerValue type alias.

TheDan64 (approved):
I'm pretty happy with the change as is, and am willing to merge it. I hear you, however, and will keep your suggestion about separate functions in mind. Also, there appears to be some broken code that is causing tests across all versions to fail. Once that is fixed I'll merge

src/builder.rs
```diff
@@ -1,26 +1,29 @@
...
 #[llvm_versions(3.9..=latest)]
 use crate::values::StructValue;
+use crate::values::CallableValue;
 use crate::types::{AsTypeRef, BasicType, IntMathType, FloatMathType, PointerType, PointerMathType};
 
 use std::marker::PhantomData;
 
+// imports used in documentation
+#[allow(unused_imports)]
+use crate::values::FunctionValue;
```
TheDan64, line 25: We should just move these imports into the docs

folkertdev: finally figured out how to both make it look nice in the rendered text and also not pollute the surrounding code.

TheDan64:
Thanks!

PR #255, opened 2021-06-12
fix invoke tests

folkertdev:
now sure how both I and CI missed all this, but a bunch of tests had type errors...


TheDan64:
hmm, looks like CI might have broken somehow?

TheDan64 (approved):

Issue #75, opened 2019-04-11
Use rustfmt for code formatting


TheDan64:
**Is your feature request related to a problem? Please describe.**
@longwusha had a great suggestion - we should have code styling preferences. Particularly via rustfmt.

**Describe the solution you'd like**
We should use rustfmt to standardize code style.

**Describe possible drawbacks to your solution**
My ideal configuration for rustfmt uses unstable options, but this project is on stable. So I do not wish to enforce nightly rustfmt.

**Additional context**
Stabilization checklist **WIP**:
- [ ] comment_width = 100 - https://github.com/rust-lang/rustfmt/issues/3349
- [ ] format_doc_comments = true  - https://github.com/rust-lang/rustfmt/issues/3348
- [ ] match_arm_blocks = false - https://github.com/rust-lang/rustfmt/issues/3373
- [x] match_block_trailing_comma = true - https://github.com/rust-lang/rustfmt/issues/3380
- [x] max_width = 120
- [x] merge_derives = true
- [x] merge_imports = true - https://github.com/rust-lang/rustfmt/issues/3362
- [x] newline_style = "Unix"
- [x] use_field_init_shorthand = true
- [x] use_try_shorthand = true
- [ ] wrap_comments = true - https://github.com/rust-lang/rustfmt/issues/3347

TheDan64:
Blocked until rustfmt v2 which will let you use unstable features on stable

TheDan64:
Going to add the rustfmt settings that are available now & we can add more later

PR #256, opened 2021-07-05
chore: Run `rustfmt`

Hywan:
# Description

This patch is the result of running `rustfmt **/*.rs` from the root of
this project. The goal is the following: When a user (like me) edits a
file, the file is automatically formatted on save; which results in
larger diffs than needed. Let's fix the coding styles once and for all
thanks to `rustfmt` :-)!

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,5 +1,8 @@
 fn main() {
-    if cfg!(all(not(target_os = "windows"), not(feature = "no-libffi-linking"))) {
+    if cfg!(all(
+        not(target_os = "windows"),
+        not(feature = "no-libffi-linking")
+    )) {
         println!("cargo:rustc-link-lib=dylib=ffi");
     }
 }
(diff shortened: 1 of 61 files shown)


TheDan64:
Hi, thanks for the PR. I appreciate the effort but unfortunately it will not be accepted. We already have an open issue here: https://github.com/TheDan64/inkwell/issues/75 which documents that we're waiting for rustfmt 2.0 so that we can use unstable rustfmt features rather than requiring users to have nightly just to format inkwell.

Regarding formatting on save, editors usually allow you to disable format on save on a per-project basis

Hywan:
Yeah, OK. I'm not sure to understand why you don't want to get a clean code base with the current `rustfmt` version, which will format 99% of the code as you expect, but that's your choice :-).

TheDan64:
It should be out relatively soon I hope. There's only one issue left: https://github.com/rust-lang/rustfmt/milestone/8

Hywan:
Good to know :-)!

Issue #260, opened 2021-08-19
Missing api to deal with type attributes.


tarberd:
I am working at a compiler front-end to llvm-ir using `inkwell` to wrap `llvm-sys`. I am using llvm version `12.0.1` and found out that `llc` does not allow you to use the `sret` parameter attribute without a type. This is also the case for any of the attributes that have a type attached to itself: `byval(<ty>)`, `byref(<ty>)`, `preallocated(<ty>)`, etc. As can be seen at the [LLVM Language Reference](https://releases.llvm.org/12.0.1/docs/LangRef.html#parameter-attributes).

`llvm-sys` was missing the necessary ffi to deal with type attributes until today. This happened due to llvm team adding to the llvm-c interface api during a minor version release as was discussed [at the llvm-sys issue tracker](https://gitlab.com/taricorp/llvm-sys.rs/-/issues/19).

I managed to get inkwell's attributes api to support type attributes in the way that `enum` and `string` attributes are already being handled. During my time doodling with inkwells code-base I happened to see that the api for attributes is supposed to receive a major refactor at some point. Although a more typed api is preferable, I don't see any harm in expanding the api in this adhoc fashion in the meantime.

PR #261, opened 2021-08-19
Add type attribute api.

tarberd:
## Description

I have extended the api for `Context` and for `Attribute` to add support for type attributes.
This includes wrapping 3 new ffi calls: 
- [LLVMCreateTypeAttribute](https://docs.rs/llvm-sys/120.2.0/llvm_sys/core/fn.LLVMCreateTypeAttribute.html)
- [LLVMGetTypeAttributeValue](https://docs.rs/llvm-sys/120.2.0/llvm_sys/core/fn.LLVMGetTypeAttributeValue.html)
- [LLVMIsTypeAttribute](https://docs.rs/llvm-sys/120.2.0/llvm_sys/core/fn.LLVMIsTypeAttribute.html)

The added api was based in the old api for attributes.

## Related Issue

#260 

## How This Has Been Tested

I have added tests in `inkwell/tests/all/test_attributes.rs` as well as doc tests in each of the methods added. Only tested with llvm version `12.0.1` through `cargo test --features "target-all,llvm12-0"`.

Also I had been tested during development of my front-end at [ekitai](https://github.com/tarberd/ekitai).

## Checklist

### DIFF
--- a/tests/all/test_attributes.rs
+++ b/tests/all/test_attributes.rs
@@ -40,6 +40,44 @@ fn test_string_attributes() {
     assert_eq!(string_attribute.get_string_value().to_str(), Ok("my_val"));
 }
 
+#[llvm_versions(12.0..=latest)]
+#[test]
+fn test_type_attribute() {
+    use inkwell::types::{AnyType, BasicType};
+    use inkwell::AddressSpace;
+
+    let context = Context::create();
+    let kind_id = Attribute::get_named_enum_kind_id("sret");
+
+    let any_types = [
+        context.i32_type().as_any_type_enum(),
+        context.f32_type().as_any_type_enum(),
+        context.void_type().as_any_type_enum(),
+        context.i32_type().vec_type(1).as_any_type_enum(),
+        context.i32_type().array_type(1).as_any_type_enum(),
+        context.i32_type().fn_type(&[], false).as_any_type_enum(),
+        context
+            .i32_type()
+            .ptr_type(AddressSpace::Local)
+            .as_any_type_enum(),
+        context
+            .struct_type(&[context.i32_type().as_basic_type_enum()], false)
+            .as_any_type_enum(),
+    ];
+
+    let different_type = context.i128_type().as_any_type_enum();
+    assert!(!any_types.contains(&different_type));
+
+    for any_type in &any_types {
+        let type_attribute = context.create_type_attribute(kind_id, *any_type);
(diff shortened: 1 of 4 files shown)


TheDan64:
I think because these changes only really apply to LLVM12, the average change went down? If so that's not a big deal

tests/all/test_attributes.rs
```diff
@@ -40,6 +40,36 @@ fn test_string_attributes() {
...
+
+    let context = Context::create();
+    let kind_id = Attribute::get_named_enum_kind_id("sret");
+    let any_type = context.i32_type().as_any_type_enum();
+    let type_attribute = context.create_type_attribute(
+        kind_id,
+        any_type,
+    );
+
+    assert!(type_attribute.is_type());
```
TheDan64, line 56: Would be good to add asserts that it's not a string or enum attr

tests/all/test_attributes.rs
```diff
@@ -40,6 +40,36 @@ fn test_string_attributes() {
...
+
+    let context = Context::create();
+    let kind_id = Attribute::get_named_enum_kind_id("byval");
+    let any_type = context.i128_type().as_any_type_enum();
+    let type_attribute = context.create_type_attribute(
+        kind_id,
+        any_type,
+    );
+
+    assert!(type_attribute.is_type());
```
TheDan64, line 68: Could we run these tests against more types to verify any value of AnyTypeEnum is valid as an attribute? Maybe in a loop of types? Like

```rust
let types = [int_type.into_any_type(), float_type.into(), vec_type.into(), ...];

for type in types {
    // create_type_attr
    // asserts
    // ...
}
```

TheDan64 (commented):
These changes look pretty good but I'd like to see the tests cover more (ideally all) variations of `AnyTypeEnum`. ~Also, I don't really understand what's the point of creating an attribute from a type. Could you explain it or point to some docs for me please?~ Nvm, found it in the issue you posted :)

tests/all/test_attributes.rs
tarberd: Tested and confirmed they are mutually exclusive.

tests/all/test_attributes.rs
tarberd: It works for all `AnyTypeEnum` variants as tested in newer commit.

TheDan64 (approved):
Awesome, thanks!

PR #262, opened 2021-08-28
Add function to get the address of a basic block

StackDoubleFlow:
## Description

I have added a function to get the address of a basic block if it exists as a PointerValue. This function is a wrapper for `LLVMBlockAddress`. 
See [here](https://llvm.org/docs/LangRef.html#addresses-of-basic-blocks) for more information on `blockaddress`.

## How This Has Been Tested

A test was added in `tests/test_basic_block.rs`, and this has also been put to use in a toy language of mine which proved it to be working.

## Checklist


TheDan64:
The docs say:

> Taking the address of the entry block is illegal.
> This value only has defined behavior when used as an operand to the ‘indirectbr’ or ‘callbr’instruction, or for comparisons against null. Pointer equality tests between labels addresses results in undefined behavior

I'm concerned that this can be unsafe to use in some cases?

StackDoubleFlow:
Ah yes, you are right. Would it be reasonable to mark it unsafe, and add a check to verify that it is not an entry block? I don't see a simple way to make sure the value is only used for `inderectbr` or `callbr`.

StackDoubleFlow:
How about adding functions to `Builder` such as `build_indirect_br_block` and `call_block` with an argument of a `BasicBlock`? These functions would call `LLVMBlockAddress` and then build a branch to that address.

TheDan64:
I think the unsafe approach makes sense. I'm not really following how the alternate idea would work

Issue #267, opened 2021-09-14
Missing "merge functions pass" from PassManager


jrolli:
**Is your feature request related to a problem? Please describe.**
Yes because generated code often contains functions which are near identical at an ISA level which differ only in type information.  LLVM has exposed an optimization pass (`mergefunc`) in its C API (as `LLVMAddMergeFunctionsPass`) since version 10.0 that is intended to address this issue by collapsing duplicate functions into a single instance.  This can significantly save on final compiled size.  There is a detailed overview of the pass [here](https://www.llvm.org/docs/MergeFunctions.html#introduction).

**Describe the solution you'd like**
The function is exposed in `llvm-sys` and can be passed through just like the other functions from `ipo.h`.

**Describe possible drawbacks to your solution**
This approach does not create a systematic approach to tracking future changes in `ipo.h` (or any other passes exposed by `llvm-sys`).

**Describe alternatives you've considered**
Alternative is to bypass Inkwell and work with `llvm-sys` directly or run the pass via shelling out to `opt`.

**Additional context**
None

pejato:
Hi @TheDan64 / @jrolli, looks like a PR was merged to resolve this. Can the issue be closed now?

TheDan64:
Good catch, thanks!

PR #268, opened 2021-09-14
Add passthru for 'MergeFunctionsPass'

jrolli:
## Description

Exposes the `mergefunc` optimization pass to _Inkwell_ through `LLVMAddMergeFunctionsPass` function in _llvm-sys_

## Related Issue

#267 

## How This Has Been Tested

Built and tested under `llvm8-0` and `llvm11-0` (versions I readily have available).  Correct feature flag based off of presence of function in upstream `llvm-sys` (https://docs.rs/llvm-sys/100.2.2/src/llvm_sys/transforms/ipo.rs.html).

## Checklist

### DIFF
--- a/tests/all/test_passes.rs
+++ b/tests/all/test_passes.rs
@@ -12,6 +12,8 @@ fn test_init_all_passes_for_module() {
 
     pass_manager.add_argument_promotion_pass();
     pass_manager.add_constant_merge_pass();
+    #[cfg(any(feature = "llvm10-0", feature = "llvm11-0", feature = "llvm12-0"))]
+    pass_manager.add_merge_functions_pass();
     pass_manager.add_dead_arg_elimination_pass();
     pass_manager.add_function_attrs_pass();
     pass_manager.add_function_inlining_pass();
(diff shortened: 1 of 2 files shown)


tests/all/test_passes.rs
```diff
@@ -12,6 +12,8 @@ fn test_init_all_passes_for_module() {
 
     pass_manager.add_argument_promotion_pass();
     pass_manager.add_constant_merge_pass();
+    #[cfg(any(feature = "llvm10-0", feature = "llvm11-0", feature = "llvm12-0"))]
```
TheDan64, line 15: It's annoying but I'd rather invert the logic here so that it's future proofed. IE `#[cfg(not(any(feature = "llvm3-6", ..., feature = "9-0")))]`

TheDan64 (approved):
Looks good, just would like to see one tweak before merging

tests/all/test_passes.rs
jrolli: No worries, makes sense to pay the annoyance once up front instead of making you add another clause for each future version of LLVM.  I've made the edit and pushed it.

TheDan64:
Thanks!!

Issue #272, opened 2021-09-27
Allow replacing a `PhiValue` with any value kind, not just another `PhiValue`


bjorn3:
**Is your feature request related to a problem? Please describe.**
This is necessary to be able to use a `PhiValue` as placeholder until the value is actually defined when translating from another SSA form ir like Cranelift ir.

**Describe the solution you'd like**
Replace the `PhiValue` argument of `PhiValue::replace_all_uses_with` with `AnyValue`.

**Describe possible drawbacks to your solution**
None that I know of.

**Describe alternatives you've considered**
I don't know of any other alternatives.

**Additional context**
I am translating cranelift ir to llvm ir.

PR #273, opened 2021-09-27
Accept AnyValue in PhiValue::replace_all_uses_with

bjorn3:
## Description

Allow replacing a `PhiValue` with any value kind, not just another `PhiValue`. This is necessary to be able to use a `PhiValue` as placeholder until the value is actually defined when translating from another SSA form ir like Cranelift ir.

## Related Issue

Fixes https://github.com/TheDan64/inkwell/issues/272

## Option\<Breaking Changes\>

I don't think this breaks anything as `PhiValue` implements `AnyValue`.

## Checklist

### DIFF
--- a/src/values/phi_value.rs
+++ b/src/values/phi_value.rs
@@ -6,7 +6,9 @@ use std::ffi::CStr;
 
 use crate::basic_block::BasicBlock;
 use crate::values::traits::AsValueRef;
-use crate::values::{BasicValue, BasicValueEnum, InstructionOpcode, InstructionValue, Value};
+use crate::values::{
+    AnyValue, BasicValue, BasicValueEnum, InstructionOpcode, InstructionValue, Value,
+};
 
 // REVIEW: Metadata for phi values?
 /// A Phi Instruction returns a value based on which basic block branched into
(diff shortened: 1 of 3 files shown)


src/values/phi_value.rs
```diff
@@ -81,7 +81,7 @@ impl<'ctx> PhiValue<'ctx> {
         self.phi_value.as_instruction().expect("PhiValue should always be a Phi InstructionValue")
     }
 
-    pub fn replace_all_uses_with(self, other: &PhiValue<'ctx>) {
+    pub fn replace_all_uses_with(self, other: &dyn AnyValue<'ctx>) {
```
TheDan64, line 84: Would you be able to write tests to verify that various AnyValue types are valid here?

Also, we should turn this into a generic rather than a dyn trait.

bjorn3: Sorry for the late response. I added a test and turned it into a generic function.

TheDan64:
Hi @bjorn3 sorry that this is taken so long to get merged. Could you please 1) rebase this PR and 2) explain how this is valid? If the IR expects a phi value, wouldn't replacing it with another type potentially break the code?

bjorn3:
> sorry that this is taken so long to get merged.

No worries.

> If the IR expects a phi value, wouldn't replacing it with another type potentially break the code?

In my case other code doesn't expect a phi value. It expects any kind of value. It is just the case that I need a dummy value as placeholder (https://github.com/bjorn3/wasmtime/blob/0bb1639f68ba46d20919e2a8c4e83f8d20d19917/cranelift/llvm/src/function.rs#L205) until it gets replaced with the actual value (https://github.com/bjorn3/wasmtime/blob/0bb1639f68ba46d20919e2a8c4e83f8d20d19917/cranelift/llvm/src/function.rs#L221) once the defining block gets lowered. Phi instructions are the only instructions I could find that are suitable as placeholder for any type of value. If you have a suggestion for a better alternative, please tell me.

TheDan64:
I'm not totally sure what the correct answer is for your situation, but something about it does seem odd 🤔

PR #281, opened 2021-11-10
remove regex dependency (outside of tests)

folkertdev:
## Description

Replaces the use of regex matching in `int_type.rs` with features from the stdlib. The main benefit is that the `regex` crate can be moved to [ `dev-dependencies`](https://doc.rust-lang.org/rust-by-example/testing/dev_dependencies.html)  and is therefore not compiled when using inkwell as a dependency.

I also generally don't like regex too much and think the rust code is nicer, but that's a matter of taste.

Regex is still used in `tests/all/test_targets.rs`, with this beauty:
```rust
    let datalayout_specification_re = Regex::new("[Ee]|S\\d+|P\\d+|A\\d+|p(\\d+)?:\\d+:\\d+(:\\d+)?|i\\d+:\\d+(:\\d+)?|v\\d+:\\d+(:\\d+)?|f\\d+:\\d+(:\\d+)?|a:\\d+(:\\d)?|F[in]\\d+|m:[emoxw]|n\\d+(:\\d)*|ni:\\d+(:\\d)*").unwrap();
```
I'm not going to touch that one.

## How This Has Been Tested

tests pass. 

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -107,8 +107,10 @@ llvm-sys-120 = { package = "llvm-sys", version = "120.2", optional = true }
 llvm-sys-130 = { package = "llvm-sys", version = "130.0", optional = true }
 once_cell = "1.4.1"
 parking_lot = "0.11"
-regex = "1"
 static-alloc = { version = "0.2", optional = true }
+
+[dev-dependencies]
+regex = "1"
 
 [badges]
 travis-ci = { repository = "TheDan64/inkwell" }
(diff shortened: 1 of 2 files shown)


folkertdev:
Is there something else I should do here?

TheDan64:
Apologies, I haven't had a chance to look at this yet. Will try and get to it soon!

TheDan64 (approved):

TheDan64:
Thanks!

Issue #274, opened 2021-10-05
Allow conversion from instruction value to phi value, if not already possible


muke101:
I want to take an InstructionValue, check if it's a phi instruction, and then create a PhiInstruction struct pointing to the same underlying LLVM value. I've looked through all the relevant parts in the documentation and I can't find a way to achieve this, so if I'm correct in that it's not possible then adding this functionality would be very helpful. If it is already possible I'd appreciate being shown how.

Thanks.

TheDan64:
hmm, it might be possible to implement

PR #285, opened 2021-11-28
Support conversion of InstructionValue to PhiValue

kalzoo:
## Description

This change supports the conversion of `InstructionValue` to `PhiValue` when the underlying instruction is in fact a `phi` instruction. It is a naive approach but which works in practice in the use case I have for it (which is currently private but which will be public in the next few weeks).

I wrote it following the conventions/patterns already used in the library, namely that it panics when a condition is unmet. I'd rather make it `try_as_phi_instruction` and return a `Result`, but that pattern isn't in use here.

Comments welcome; if this approach is worthy of merging I can add tests.

## Related Issue

Closes https://github.com/TheDan64/inkwell/issues/274

## How This Has Been Tested

I am using it in a pre-production library to convert an InstructionValue; otherwise, no testing.

## Option\<Breaking Changes\>

All changes are additive.

## Checklist

`cargo clippy` emits 1155 errors unrelated to my changes.


src/values/instruction_value.rs
```diff
@@ -675,6 +675,18 @@ impl<'ctx> InstructionValue<'ctx> {
...
+
+    /// Return this `InstructionValue` as a `PhiValue`.
+    /// Panics if the instruction is not a Phi instruction.
+    pub fn as_phi_value(&self) -> PhiValue<'ctx> {
+        if self.get_opcode() == InstructionOpcode::Phi {
+            unsafe {
+                PhiValue::new(self.as_value_ref())
+            }
+        } else {
+            panic!("Found {:?} but expected the PhiValue variant", self)
```
TheDan64, line 687: This should be a None rather than panicing

TheDan64: Or even `TryFrom` impl

dbanty: This is switched to `TryFrom` now, can you give it another look @TheDan64? Thanks!

TheDan64:
I like it, could you please add a test for when it passes and for when it fails?

nilslice:
@TheDan64 - I've added a test for both expected success and failure cases. Lmk what you think, thanks!

TheDan64 (approved):

Issue #179, opened 2020-04-25
Expose some missing JIT API functionality for cross-module linking


nepp2:
**Is your feature request related to a problem? Please describe.**

I can't figure out how to link symbols across different modules using Inkwell (with LLVM 8). There may be some way that I missed, but if not it would be a major limitation. The [inkwell kaleidoscope example](https://github.com/TheDan64/inkwell/tree/master/examples/kaleidoscope) is excellent, but it avoids cross-module linking by doing a lot of redundant recompilation instead.

The inkwell kaleidoscope example's approach to linking functions is very simple. Each time you enter a new expression into the REPL, [every previous expression is recompiled into one monolithic module along with the new expression](https://github.com/TheDan64/inkwell/blob/4c9e29738bb2442d6e6e012d5da5d1d50262e0e1/examples/kaleidoscope/main.rs#L1282). This is done to make sure that any function you have defined is part of the same module as the new expression, which means there's no need to do cross-module linking. That works fine for a small demo, but it is not a scalable solution for a larger project.

Part of the problem is that the original C++ kaleidoscope example has lived through some big changes to the JIT API. In particular, it was previously possible to add new functions and symbols to an existing module so that no cross-module linking was required. When MCJIT was introduced, this was no longer supported. The new REPL implementation strategy was to compile each expression using a new module and execution engine, and then handle cross-module linking by overriding something called the SectionMemoryManager. [You can read about this transition on the official LLVM blog](http://blog.llvm.org/2013/07/using-mcjit-with-kaleidoscope-tutorial.html).

**Describe the solution you'd like**

I am not sure if it is possible to override the built-in linking behaviour through the C API. In my own Inkwell-based project [my solution](https://github.com/nepp2/live-programming-prototype/blob/591895b294d80b0b53aaf429396650cf062e9a7a/compiler/src/llvm_compile.rs#L129) was to link symbols manually when I compile a new module. Whenever the compiler sees an external symbol reference, it searches the previous modules for it, gets its address, and updates the new module with this address.

To do this I used the [`add_global_mapping`](https://thedan64.github.io/inkwell/inkwell/execution_engine/struct.ExecutionEngine.html#method.add_global_mapping) function, but I also had to expose two extra LLVM functions [in this commit on my Inkwell fork](https://github.com/nepp2/inkwell/commit/6c0296b1721ba7945b4b922ed3e016973c296273) (the generic parameters are an error I fixed in the next commit).

[...]

PR #292, opened 2022-01-21
Add API for the ORCv2 execution engine 

leifhelm:
## Description

This PR aims to add  the ability to safely interact with the ORCv2 execution engine.
The new orc2 module houses all the new structs and functions.
I wrote the `lljit` example to demonstrate the new execution engine.
## Related Issue
 #179 

## How This Has Been Tested

I tested the new module with the new `lljit` example. To verify that the memory management is working out I used `valgrind`.

## Checklist


leifhelm:
I changed the underlying implementation of `LLVMStringOrRaw` to be more generic. Furthermore I added lifetime specifiers to methods that return `LLVMStringOrRaw` and any wrapper types (`DataLayout`, `TargetTriple`). I hope this is fine with you (see b6c5d43). 
Tests for LLVM version 11 and 12 should be working.

tpisto:
Hi! Great work - is this compatible with LLVM 14?

leifhelm:
Inkwell does currently not support LLVM 14 see #317.
This PR is kind of stale, as I don't know how to best test some of the features. Especially test that break the current implementation are needed.
Furthermore I don't feel motivated currently to invest huge amounts time to move this forward.
There are still functions from LLVM 13 that are not wrapped. And I have not looked into multi-threading at all.

pablosichert:
This PR is indeed some nice work, @leifhelm! I've been working on wrapping the LLVM ORC API before I stumbled on this, and thought I'd rather invest my time to contribute here. Specifically, I have a wrapper for `LLVMOrcCreateCustomCAPIDefinitionGenerator` to be able to write a custom symbol resolver that I'd like to add.

> Especially test that break the current implementation are needed.

Could you elaborate what you're looking for? Does this PR introduce backwards-incompatible changes that are not caught by the compiler with the new lifetime parameters?

Considering that `inkwell` is currently distributed by linking to the `master` branch of this repo, I don't think `inkwell` provides any guarantees about compatibility, or that users would expect any versioning guarantees?

> There are still functions from LLVM 13 that are not wrapped. And I have not looked into multi-threading at all.

Providing an exhaustive implementation for all feature of the C API doesn't seem necessary to make this PR very useful nonetheless. Features that build on top of this can be contributed in more PRs in the future. Also since none of the structs are `Send` + `Sync`, not implementing multi-threading now won't break user code in the future.

leifhelm:
The current implementation  of orc2 is broken. I have an unpushed testcases that fails, regarding the materialization unit. But I was unable to write an example for that feature.

The changes should not break current functionality of `inkwell`. 

@pablosichert are you able to share how you use this function?

pablosichert:
@leifhelm sure, you can find my implementation that I added on top of your branch here: https://github.com/pablosichert/inkwell/commits/orc2.

The commit specifically to implement `JitDylib::add_generator`/`CustomDefinitionGenerator`: https://github.com/pablosichert/inkwell/commit/e335e56c4f293a6e0ef22d062bce8483cf3ec6d0.

Feel free to pick from that and apply to your branch as you see fit.

I noticed that you use `materializer: Box<(dyn Materializer + 'jit)>` in `MaterializationUnit::create`. I think we should be able to use a generic parameter instead, similar to how I did in `add_generator`.

Generally, I think we should be able to use a little less `transmute` calls, and use the inner fields that are available from within the crate, or add specific `pub(crate)` methods.

Another small nitpick: According to the [Rust naming conventions](https://github.com/rust-lang/rfcs/blob/master/text/0430-finalizing-naming-conventions.md):

> In UpperCamelCase, acronyms count as one word: use Uuid rather than UUID

Meaning that I'd suggest to rename `LLJIT` -> `Lljit`, `JITDylib` -> `JitDylib`, `RTDyldObjectLinkingLayer` -> `RtDyldObjectLinkingLayer`, `IRTransformLayer` -> `IrTransformLayer` and so on.

I'd be happy to help move progress forward on this PR. If you need a full code review or help with debugging and fixing test cases, let me know.

leifhelm:
@pablosichert do you have a better solution than `Wrapper` that I introduced in leifhelm@8643148?

pablosichert:
@leifhelm hah, yes good catch. I had a segfault locally that boiled down to the generator being dropped when the JIT was still alive. I left a suggestion on your commit

src/orc2/mod.rs
```diff
@@ -0,0 +1,2150 @@
...
+    pub fn create<'jit>(
+        name: &str,
+        mut symbols: SymbolFlagsMapPairs,
+        static_initializer: Option<SymbolStringPoolEntry>,
+        materializer: Box<(dyn Materializer + 'jit)>,
+    ) -> Self {
+        unsafe {
+            MaterializationUnit::new(LLVMOrcCreateCustomMaterializationUnit(
+                to_c_str(name).as_ptr(),
+                transmute::<*mut MaterializationUnitCtx, _>(Box::into_raw(Box::new(materializer))),
```
pablosichert, line 688:
```suggestion
                materializer.as_mut() as *mut _ as *mut MaterializationUnitCtx,
```

probably not needed anymore when this also uses a generic `Materializer` instead

src/orc2/mod.rs
```diff
@@ -0,0 +1,2150 @@
...
+}
+
+#[llvm_versions(13.0..=latest)]
+impl SymbolFlagsMapPair {
+    pub fn new(name: SymbolStringPoolEntry, flags: SymbolFlags) -> Self {
+        Self {
+            pair: LLVMOrcCSymbolFlagsMapPair {
+                Name: unsafe { transmute(name) },
+                Flags: unsafe { transmute(flags) },
+            },
```
pablosichert, line 1540:
```suggestion
            pair: LLVMOrcCSymbolFlagsMapPair {
                Name: name.entry,
                Flags: flags.flags,
            },
```

src/orc2/mod.rs
```diff
@@ -0,0 +1,2150 @@
...
+}
+
+#[llvm_versions(12.0..=latest)]
+impl SymbolMapPair {
+    pub fn new(name: SymbolStringPoolEntry, symbol: EvaluatedSymbol) -> Self {
+        SymbolMapPair {
+            pair: LLVMJITCSymbolMapPair {
+                Name: unsafe { transmute(name) },
+                Sym: unsafe { transmute(symbol) },
+            },
```
pablosichert, line 1711:
```suggestion
            pair: LLVMJITCSymbolMapPair {
                Name: name.entry,
                Sym: symbol.symbol,
            },
```

src/orc2/mod.rs
```diff
@@ -0,0 +1,2150 @@
...
+}
+
+#[llvm_versions(12.0..=latest)]
+impl EvaluatedSymbol {
+    pub fn new(address: u64, flags: SymbolFlags) -> Self {
+        EvaluatedSymbol {
+            symbol: LLVMJITEvaluatedSymbol {
+                Address: unsafe { transmute(address) },
+                Flags: unsafe { transmute(flags) },
+            },
```
pablosichert, line 1759:
```suggestion
            symbol: LLVMJITEvaluatedSymbol {
                Address: address,
                Flags: flags.flags,
            },
```

pablosichert (commented):
Some suggestions to elide a couple of `transmute` calls (which are an easy source of bugs with future changes).

Some of these should be obsolete when using rewriting `ObjectTransformer`, `ObjectLinkingLayerCreator`, `Materializer` to be generic arguments rather than used in `Box`

src/orc2/lljit.rs
```diff
@@ -0,0 +1,823 @@
...
+#[llvm_versions(13.0..=latest)]
+type ObjectTransformerCtx<'jit> = Box<dyn ObjectTransformer + 'jit>;
+
+#[llvm_versions(13.0..=latest)]
+#[no_mangle]
+extern "C" fn object_transform_layer_transform_function(
+    ctx: *mut c_void,
+    object_in_out: *mut LLVMMemoryBufferRef,
+) -> LLVMErrorRef {
+    let object_transformer: &mut ObjectTransformerCtx = unsafe { transmute(ctx) };
```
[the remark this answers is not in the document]
leifhelm: `ctx` is a `*mut Box<>` in disguise. Why is it bad to cast it to `&mut Box<>`?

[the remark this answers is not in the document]
leifhelm: Or do you mean that object_transformer should be a `Box<Box<>>`?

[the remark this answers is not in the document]
pablosichert: Ah, that makes sense, I thought it was the inner pointer. Could use

```rust
let object_transformer: &mut ObjectTransformerCtx = unsafe { &mut *(ctx as *mut _) };
```

then. That's not less error-prone than a transmute, but a bit more explicit about the cast.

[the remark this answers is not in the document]
pablosichert: Just noticed that you did exactly that in 090b715689e7ea378d1696cb650e16f04bb03b72 👍

src/orc2/mod.rs
```diff
@@ -0,0 +1,2186 @@
...
+            definition_generator: DefinitionGeneratorRef::new(definition_generator),
+        }
+    }
+
+    #[llvm_versions(12.0..=latest)]
+    pub fn create_custom_capi_definition_generator<G>(
+        capi_definition_generator: &'jit mut Wrapper<'jit, G>,
+    ) -> Self
+    where
+        G: CAPIDefinitionGenerator,
```
pablosichert, line 1321: Back to this, I'm still very confused why it allows dropping `capi_definition_generator` before the instance of `LLVMJIT<'jit>` (when using `&'jit mut G` without `Wrapper`).

I can model that problem in a small reproduction and ask what the guys in the Rust forums think

src/orc2/mod.rs
```diff
@@ -0,0 +1,2186 @@
...
+                <M as UnsafeMaterializer>::materialization_unit_destroy,
+            ))
+        }
+    }
+
+    pub fn from_absolute_symbols(mut symbols: SymbolMapPairs) -> Self {
+        unsafe {
+            MaterializationUnit::new(LLVMOrcAbsoluteSymbols(symbols.raw_ptr(), symbols.len()))
+        }
+    }
```
pablosichert, line 710:
```suggestion
    pub fn from_absolute_symbols(mut symbols: SymbolMapPairs) -> Self {
        for name in symbols.names_iter() {
            unsafe { LLVMOrcRetainSymbolStringPoolEntry(name.entry) };
        }
        unsafe {
            MaterializationUnit::new(LLVMOrcAbsoluteSymbols(symbols.raw_ptr(), symbols.len()))
        }
    }
```

From the LLVM docs for [`LLVMOrcAbsoluteSymbols`](https://llvm.org/doxygen/group__LLVMCExecutionEngineORC.html#ga2fa4d5572b2c93896a2999f5b5e3ca36):

> The Name fields of the array elements are taken to have been retained for this function.

src/orc2/mod.rs
```diff
@@ -0,0 +1,2186 @@
...
+}
+
+#[llvm_versions(12.0..=latest)]
+impl CLookupSetElement {
+    pub fn get_name(&self) -> SymbolStringPoolEntry {
+        unsafe { SymbolStringPoolEntry::new(self.element.Name) }
+    }
+
+    pub fn get_flags(&self) -> SymbolLookupFlags {
+        self.element.LookupFlags.clone().into()
```
pablosichert, line 2174: This fails to compile for me, is the `llvm_enum` macro supposed to provide a `From` implementation?

codecnotsupported:
Any update? 
Branch seems to be outdated with LLVM 14.

gavrilikhin-d:
The work done here must be partially added to inkwell. 

There are always problems with merging big pull requests (4k lines are a lot). This feature should be split into many small changes, frequently added to master. Otherwise, it will never be implemented.

pablosichert:
@gavrilikhin-d your point about incubating smaller incremental changes makes sense in isolation.

However, I think it would be great if you could consider the time constraints and motivation of individuals contributing to open source in their free time before making any demands to the structure or speed of their work.

In case there was some notion that has been lost in translation, I'd like to let you know that your comment has a bit of a rude tone.

leifhelm:
Well I think he should not merge it, because there are still some major flaws rooted deep within this API binding. One could say we will fix that later. However I think that the problems that will arise are not easily solvable and do not align with the approach of this PR. So a major rewrite of the orc module would be necessary.

I made this PR so big because I do not understand how LLVM ORC work. So I tried to add features one after the other. This resulted in me rewriting several core structs several times (adding/removing lifetimes, references, ...). This approach is valuable if you want to learn how this ORC thing works, but it is more of a hack.

In the end I concluded that I don't know enough about ORC that I cannot design a proper binding, because I do not understand what lifetimes are here because my implementation needs them and what lifetimes are here because they are intrinsic to the domain.

Issue #297, opened 2022-01-27
Add support for aligned loads and stores


DCNick3:
**Is your feature request related to a problem? Please describe.**
I want to be able to do loads and stores using unaligned pointers on platforms like ARM. Sometimes it requires special code generation, so it's nice to supply LLVM with this info

**Describe the solution you'd like**
Versions of `build_store` and `build_load` accepting an alignment argument

**Describe possible drawbacks to your solution**
Can't see any

**Describe alternatives you've considered**
It may be overcame with memcpy shenanigans, but is clearly suboptimal

PR #298, opened 2022-01-27
[WIP] Add support for aligned loads and stores

DCNick3:
## Description

LLVM C API does not expose functions for creating aligned loads and stores, but it can be achieved easy enough by setting the alignment after the creation of the corresponding instructions using `LLVMSetAlignment`.

## Related Issue

#297

## How This Has Been Tested

Create load instructions with alignment 1 (basically meaning unaligned). Generated LLVM IR seems to be fine, compiling successfully.

## Checklist

## Open questions:

- docs (should I mostly copy-paste the build_load & build_store docs or write only about differences in API?)

- how to handle invalid alignments?

LLVM supports only alignments that are powers of 2. And it ¿crashes? if provided with invalid alignment.

There is `is_alignment_ok` check used for `build_memcpy` and `build_memmove` with them returning a `Result` (and being an error if alignment does not match).

I kind of don't like this API, because invalid alignment being is a really exceptional case that I will not ever run into (using constant 1 as a parameter), it hinders my ability to use build_{load,store}_aligned as a drop-in replacement for existing functions, requiring unwraps or alike.

Maybe it will be more appropriate to panic in this case? Though it seems that this crate does not panic a lot..

### DIFF
--- a/src/targets.rs
+++ b/src/targets.rs
@@ -771,14 +771,14 @@ impl Target {
             unsafe { LLVMInitializeRISCVAsmPrinter() };
         }
 
-        if config.asm_printer {
-            let _guard = TARGET_LOCK.write();
-            unsafe { LLVMInitializeRISCVAsmPrinter() };
-        }
-
         if config.asm_parser {
             let _guard = TARGET_LOCK.write();
             unsafe { LLVMInitializeRISCVAsmParser() };
+        }
+
+        if config.disassembler {
+            let _guard = TARGET_LOCK.write();
+            unsafe { LLVMInitializeRISCVDisassembler() };
         }
 
         if config.machine_code {


TheDan64 (commented):
We already expose safe versions of `LLVMBuildStore` and `LLVMSetAlignment`, why not just use those methods rather than adding these new functions?

DCNick3:
>We already expose safe versions of LLVMBuildStore and LLVMSetAlignment

Whoops, I didn't find an exposed `LLVMSetAlignment` when I was looking for it.

Using it, however, does require a fair amount of boilerplate (especially for loads):

store:

```rust
builder.build_store(ptr, val)
    .set_alignment(1)
    .unwrap()
```

load:

```rust
let val = builder.build_load(ptr, "");
val.as_instruction_value()
    .unwrap()
    .set_alignment(1)
    .unwrap();
return val
```

So, it now seems to me, that this PR is not necessary in order to implement aligned loads and stores, but might be a nice improvement to an API.

PR #308, opened 2022-03-07
Add a new internal-getters feature to expose LLVM references

### LINKED ISSUE #275: Allow public calls to AsValueRef trait
I need to read the LLVMValueRef fields of LLVM objects I'm working with in my project that uses inkwell, and I don't think this is a very uncommon need either. The LLVM C API is significantly reduced compared to the regular API in areas and to use it in it's full capacity you occasionally have to wrap it's function with your own C code, which take LLVMValueRef values.

The obvious solution would be to make the implementations of the AsValueRef trait public. I understand writing to or even just holding this value can be unsafe but I don't think that warrants completely disallowing it altogether when there are very reasonable use cases for doing so.

ghaith:
## Description
Added a trait `LLVMReference` hidden behind the features `internal-getters`
The Trait can provide the `LLVMRef` type for various inkwell components.
I implemented the traits for all the types I found, but I could have missed some.

## Related Issue
Fixes #275

## How This Has Been Tested
I was not sure what kind of unit tests I could add here, I'm open to ideas.
I tested the changes on a local project where I used internal values and called into the `llvm-sys` API directly.

## Checklist


src/lib.rs
```diff
@@ -430,3 +430,15 @@ pub enum InlineAsmDialect {
     #[llvm_variant(LLVMInlineAsmDialectIntel)]
     Intel,
 }
+
+
+/// Exposes LLVM internal references to the outside.
+#[cfg(feature = "internal-getters")]
+pub trait LLVMReference<T> {
```
TheDan64, line 437: Since all types can only resolve to exactly one type, this should be an associated type and not a generic:

```rust
pub trait LLVMReference {
    type Ref;
    unsafe fn get_ref(&self) -> Self::Ref;
}
```

TheDan64 (changes requested):
Would like to see the trait be an associated type rather than a generic, & the merge conflict needs to be resolved before merging.

src/lib.rs
ghaith: I changed the trait as requested, the only problem I had there is that the implementation for `AsValueRef` and `AsTypeRef` will require `dyn` as the trait will conflict otherwise.
https://stackoverflow.com/questions/39159226/conflicting-implementations-of-trait-in-rust

ghaith:
For the sake of completion, once #312 is merged, we would need to add an extra Ref here (`LLVMPassBuilderOptionsRef`)
I assume keeping these refs up to date will have to happen on a case by case basis. 
I can add a test that will call all currently implemented refs to ensure the current ones don't disappear, but updates will also be hard to track.

src/lib.rs
TheDan64: I'm not sure I follow, what exactly is conflicting?

ghaith: I was trying to implement `LLVMReference` for all types implementing `AsValueRef` and all types implementing `AsTypeRef`
So my implementation looks like this
```rust
//values/mod.rs
#[cfg(feature="internal-getters")]
impl<T : AsValueRef> LLVMReference for T {
    type Ref = LLVMValueRef;
    unsafe fn get_ref(&self) -> Self::Ref {
       self.as_value_ref() 
    }
}
//types/mod.rs
#[cfg(feature="internal-getters")]
impl<T : AsTypeRef> LLVMReference for T {
    type Ref  = LLVMTypeRef;
    unsafe fn get_ref(&self) -> Self::Ref {
       self.as_type_ref() 
    }
}
```
This does not compile because of a conflicting implementation
```
   --> src/values/mod.rs:227:1
    |
227 | impl<T : AsValueRef> LLVMReference for T {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation
    |
   ::: src/types/mod.rs:207:1
    |
207 | impl<T : AsTypeRef> LLVMReference for T {
    | --------------------------------------- first implementation here
```
If I understand correctly, this is because there is no guarantee that types implementing `AsTypeRef` will not at a future point also implement `AsValueRef`
My solution was to use `dyn`. Maybe I'm missing something?

TheDan64: hmm, I see. I didn't know you could impl for a dyn trait. I wonder if it works as expected 🤔

ghaith: It seems from a quick [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=10cd3e70b9a8a687e0291dce7dc11bf2) try that this does not work as intended.

I'll look for options to keep this without generics. Some [option ](https://geo-ant.github.io/blog/2021/mutually-exclusive-traits-rust/)I found was to try and emulate mutually exclusive traits by using a wrapper trait. But I'm still not sure if that's the best option

TheDan64: It does work, but you have to explicitly cast to the dyn trait with is pretty ugly: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f49b5ce1409f4977235fbf9f201d3fd6

ghaith: I'll add some tests and use generics then. I was thinking if I limit the generics only to the AsRef implementations it would then keep the associated types everywhere else, which makes more sense for 1-1 mappings.
I'll give it a shot, and if I can't manage it I will revert to generics everywhere

ghaith:
Sorry for the delay, I uploaded a version where only the trait-based fields get generics (I used a new type for that: `LLVMTaitReference`)
I thought it might be better than using generics everywhere, although I have to admit the naming might not be clear.
If you prefer we stay with one type I will revert all types to their generic counterpart as I was not able to solve it otherwise.

TheDan64:
Let's just keep it simple and revert to the one trait with generics. Sorry for all the back and forth on this 😅 & thank you for taking the time to explore multiple options

TheDan64 (approved):
Thanks!

muke101:
Hey, can I request this trait also be implemented to instruction values? That's generally the main use in the first place, and from what I can tell the best alternative right now is to get the value ref of the basic block and then find the instruction you want within it, which moves a lot more of your logic than you'd like out of rust and into C. 

Thanks.

ghaith:
I think this was an oversight on my part. I implemented the trait for all types that had an LLVMValueRef so I missed the instruction which just wraps the internal value here. (The trait is implemented on the value)
If @TheDan64 is ok with it this is a relatively small change (the trait on `InstructionValue` would have to forward to the value)

TheDan64:
Sure, sounds good. Though fyi - I likely won't be able to review PRs til late next week

ghaith:
I was looking into this again, and the trait is actually implemented for all Value types that implement `AsValueRef` and for all types that implement `AsTypeRef` so instruction values should already be included.
@muke101 Could you verify this again, please? You should be able to call `get_ref` on an `InstructionValue` by importing the `LLVMReference trait.

muke101:
Ah, sorry you're right - I was looking through the pull request and couldn't see any changes to the instruction file and was also using `.as_value_ref()` instead, my bad!

PR #311, opened 2022-03-10
Add a function to build int casts that respect the target signed-ness

ayazhafiz:
The current `build_int_cast` uses `LLVMBuildIntCast`, which is [always
signed and is deprecated](https://github.com/hdoc/llvm-project/blob/a6ad3505abc7409abd2a4118338b9c85ec2e9f09/llvm/include/llvm-c/Core.h#L3939-L3941).
This commit adds a function that uses `LLVMBuildIntCast2`, which
correctly casts ints to unsigned int types if asked to do so.

### DIFF
--- a/tests/all/test_builder.rs
+++ b/tests/all/test_builder.rs
@@ -703,6 +703,22 @@ fn test_vector_convert_ops() {
     builder.build_return(Some(&casted_vec));
     assert!(fn_value.verify(true));
 
+    #[llvm_versions(8.0..=latest)]
+    {
+    // Here we're building a function that takes in a <3 x i8> (signed) and returns it casted to and from a <3 x i8> (unsigned)
+    let fn_type = int8_vec_type.fn_type(&[int8_vec_type.into()], false);
+    let fn_value = module.add_function("test_int_vec_cast", fn_type, None);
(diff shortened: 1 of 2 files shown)


ayazhafiz:
I took the liberty of sending this patch without first filing an issue; let me know if this is not something you'd like. Also, I'm no good with names so I'm happy for any better suggestions for the function.

tests/all/test_builder.rs
```diff
@@ -703,6 +703,22 @@ fn test_vector_convert_ops() {
     builder.build_return(Some(&casted_vec));
     assert!(fn_value.verify(true));
 
+    #[llvm_versions(8.0..=latest)]
+    {
+    // Here we're building a function that takes in a <3 x i8> (signed) and returns it casted to and from a <3 x i8> (unsigned)
```
TheDan64, line 708: Nit, should be indented

TheDan64 (approved):
Just have an indentation nit, otherwise looks good

TheDan64:
Also, tests are failing and need to be fixed

ayazhafiz:
I believe @folkertdev  fixed the two issues above

Issue #309, opened 2022-03-08
Implement the LLVMRunPasses


ghaith:
**Is your feature request related to a problem? Please describe.**
LLVM 13 Introduced an `LLVMRunPasses` method to run all passes on a module. [Relevent LLVM Commit](https://reviews.llvm.org/D102136)
I believe this relates to the new pass manager. It is supposed to take the same arguments as the `opt` application to optimize the IR

**Describe the solution you'd like**
Inkwell could expose this method either in the `passes` module or directly on the `Module`, the method expects a module reference in addition to a string of passes, a target and pass options.
I believe all the requirements can be currently constructed from information that inkwell already has.

**Describe possible drawbacks to your solution**
Should not affect existing code

**Describe alternatives you've considered**
With #275 in place, one could call this directly but this is more a hack/workaround than a solution

**Additional context**
Optimizing the IR (only tested with the opt app for now) [significantly](https://discourse.llvm.org/t/performance-of-clang-vs-llc/60515) improves the compilation speed of some of our programs

ghaith:
I'm looking into how to implement this, and I'm struggling a bit with the Architecture.
The new method logically belongs to the Module but requires a machine, and some new PassManager options. Implementing it in `module.rs` will cause targets and passes to be imported into the `module.rs` file.
Doing it in `passes.rs` introduces modules and target dependencies there.
Which option would you prefer for this?
Should I create a `impl Module {}` in the target.rs, should I pull in the depedencies to the `module.rs` file or do you have another way to handle this?

ghaith:
I went ahead with updating the `module.rs` as it already had dependencies to the `targets` module.

TheDan64:
We can close this issue now, right?

ghaith:
> We can close this issue now, right?

Yes the passes are already in.

PR #312, opened 2022-03-17
Add the runpasses option to the module

ghaith:
## Description
Added a method to the `Module` to run the llvm passes

## Related Issue
Issue #309 

## How This Has Been Tested
I tested this on a separate project.
I am not sure what kind of tests I could add to the module here, suggestions are appreciated.
The change only applies from LLVM13, as the API was introduced then, the feature flags are set appropriately

## Checklist


TheDan64 (approved):

TheDan64:
Is there really no way to add tests for this?

ghaith:
> Couldn't you create a module, enable all options, and assert that it passes? Also testing the error case of the function would be good too

I can indeed add such tests. I was thinking of tests that would verify the pass was executed, but those won't be necessary on inkwell's side, just that they were given to the module.

PR #313, opened 2022-03-28
Get all function/callsite attributes as a vector

folkertdev:
Add functions to `FunctionValue` and `CallSiteValue` that give back a vector with all attributes at a particular location on that value.

## Description

LLVM primitives have existed since llvm 3.9. The tricky bit here is that we need to give LLVM a pointer that it writes the attributes into. This requires `MaybeUninit` to do safely. Afterwards we transmute the `MaybeUninit` away (the alternative would create a second vector which is undesirable).

## How This Has Been Tested

Testing is done similarly to the `count_attributes` functions. We check the attributes we get back are as expected after adding/removing attributes.

## Checklist

---

as an aside: following the standard `cargo fmt` formatting rules would make contributing much easier/fun for me.

### DIFF
--- a/src/values/call_site_value.rs
+++ b/src/values/call_site_value.rs
@@ -240,7 +240,7 @@ impl<'ctx> CallSiteValue<'ctx> {
     #[llvm_versions(3.9..=latest)]
     pub fn attributes(self, loc: AttributeLoc) -> Vec<Attribute> {
         use llvm_sys::core::LLVMGetCallSiteAttributes;
-        use std::mem::MaybeUninit;
+        use std::mem::{ManuallyDrop, MaybeUninit};
 
         let count = self.count_attributes(loc) as usize;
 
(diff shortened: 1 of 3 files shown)


src/values/call_site_value.rs
```diff
@@ -210,6 +210,56 @@ impl<'ctx> CallSiteValue<'ctx> {
...
+        unsafe {
+            LLVMGetCallSiteAttributes(
+                self.as_value_ref(),
+                loc.get_index(),
+                attribute_refs.as_mut_ptr() as *mut _,
+            )
+        }
+
+        // Safety: everything is initialized
+        unsafe { std::mem::transmute(attribute_refs) }
```
TheDan64, line 260: Conceptually I'm fine with this change, but we should avoid transmute at all costs imo. The transmute docs do have an example on converting a `Vec<A>` to `Vec<B>` without transmute that we should try and leverage: https://doc.rust-lang.org/std/mem/fn.transmute.html

(search for "This is the proper no-copy, unsafe way of "transmuting" a `Vec`")

src/values/fn_value.rs
```diff
@@ -361,6 +361,50 @@ impl<'ctx> FunctionValue<'ctx> {
...
+        unsafe {
+            LLVMGetAttributesAtIndex(
+                self.as_value_ref(),
+                loc.get_index(),
+                attribute_refs.as_mut_ptr() as *mut _,
+            )
+        }
+
+        // Safety: everything is initialized
+        unsafe { std::mem::transmute(attribute_refs) }
```
TheDan64, line 405: Same here

TheDan64 (changes requested):

TheDan64:
re:
> as an aside: following the standard cargo fmt formatting rules would make contributing much easier/fun for me.

Plan to do this soon before releasing 0.1.0.

TheDan64 (approved):

PR #314, opened 2022-03-30
make get_enum_kind_id work on type attributes

folkertdev:
## Description

remove an assert that makes `get_enum_kind_id` panic when given a type attribute. 

I'm not entirely sure what the difference is between enum and type attributes. It looks like enum is a superclass. By default they have integer values, but since version 12.0 they can also have type as the value. And then some specific functions are added for type attributes, but specifically the kind ID is still only available with the "enum" functions, e.g. `get_named_enum_kind_id`.

So the choice I made here is that inkwell should give the same answer as LLVM for a type attribute, rather than panicking.

An alternative is to add a new function `get_type_kind_id` which has the same implementation as `get_enum_kind_id` and would assert that `attribute.is_type()`.

Finally, some docs in the module hint at a more radical change in how inkwell represents attributes. 

> Determines whether or not an `Attribute` is an enum. This method will
> likely be removed in the future in favor of `Attribute`s being generically
> defined.

Does that allude to attributes as described here https://llvm.org/docs/HowToUseAttributes.html? or is this some different idea?
## Checklist

### DIFF
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -124,6 +124,25 @@ impl Attribute {
     ///
     /// assert_eq!(enum_attribute.get_enum_kind_id(), 0);
     /// ```
+    #[llvm_versions(3.6..12.0)]
+    pub fn get_enum_kind_id(self) -> u32 {
+        assert!(self.get_enum_kind_id_is_valid()); // FIXME: SubTypes
+
+        unsafe { LLVMGetEnumAttributeKind(self.attribute) }
+    }
+
+    /// Gets the kind id associated with an enum `Attribute`.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    /// use inkwell::context::Context;
+    ///
+    /// let context = Context::create();
+    /// let enum_attribute = context.create_enum_attribute(0, 10);
+    ///
+    /// assert_eq!(enum_attribute.get_enum_kind_id(), 0);
+    /// ```
     ///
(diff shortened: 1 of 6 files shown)


folkertdev:
hmm, it's difficult to provide an example with a type kind id for the current implementation, because the relevant functions are only available in llvm 12 and higher. Maybe an argument for making it a separate function?

TheDan64:
> Finally, some docs in the module hint at a more radical change in how inkwell represents attributes.
The idea is to have like `Attribute<Enum>` and `Attribute<String>` one day. Sounds like we might need `Attribute<Type>` though?

src/attributes.rs
```diff
@@ -124,12 +124,60 @@ impl Attribute {
...
+    /// let type_attribute = context.create_type_attribute(
+    ///     kind_id,
+    ///     any_type,
+    /// );
+    ///
+    /// assert_eq!(type_attribute.get_enum_kind_id(), kind_id);
+    /// ```
+    #[llvm_versions(12.0..=latest)]
+    pub fn get_enum_kind_id(self) -> u32 {
+        assert!(self.get_enum_kind_id_is_valid()); // FIXME: SubTypes
```
TheDan64, line 166: Why are two versions needed when they're identical?

folkertdev: because the comment in the documentation uses a function that is only supported in later versions. Apparently cargo always at least does name resolution on those code snippets

Issue #316, opened 2022-04-09
Support LLVM14


nbaksalyar:
Starting from v1.60, the Rust compiler uses LLVM 14.
It doesn't have many breaking changes since LLVM 13.0, so it would be nice to have support for the new version in Inkwell.

PR #317, opened 2022-04-09
Add support for LLVM 14

nbaksalyar:
## Description

This is a WIP to support a new version of LLVM.
It's currently [blocked by `llvm-sys`](https://gitlab.com/taricorp/llvm-sys.rs/-/issues/30).

## Related Issue

#316

## How This Has Been Tested

This builds locally on my machine (I tested it both with v13 and v14). We'll need to wait until `llvm-sys-140` is released though.

## Checklist

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -225,7 +225,7 @@ impl Context {
     pub fn create_inline_asm(&self, ty: FunctionType, mut assembly: String, mut constraints: String, sideeffects: bool, alignstack: bool, dialect: Option<InlineAsmDialect>,
         can_throw: bool,
     ) -> PointerValue {
-        #[cfg(any(feature = "llvm13-0"))]
+        #[cfg(any(feature = "llvm13-0", feature = "llvm14-0"))]
         let can_throw_llvmbool = can_throw as i32;
 
         let value = unsafe {
--- a/src/context.rs
+++ b/src/context.rs
@@ -225,7 +225,6 @@ impl Context {
     pub fn create_inline_asm(&self, ty: FunctionType, mut assembly: String, mut constraints: String, sideeffects: bool, alignstack: bool, dialect: Option<InlineAsmDialect>,
         can_throw: bool,
     ) -> PointerValue {
-        #[cfg(any(feature = "llvm13-0", feature = "llvm14-0"))]
         let can_throw_llvmbool = can_throw as i32;
 
         let value = unsafe {
--- a/src/values/metadata_value.rs
+++ b/src/values/metadata_value.rs
@@ -13,6 +13,8 @@ use crate::values::{BasicMetadataValueEnum, Value};
 use std::ffi::CStr;
 use std::fmt;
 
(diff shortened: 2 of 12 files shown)


src/values/metadata_value.rs
```diff
@@ -40,6 +40,8 @@ pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 30;
 pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 31;
 #[cfg(feature = "llvm13-0")]
 pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 31;
+#[cfg(feature = "llvm14-0")]
+pub const FIRST_CUSTOM_METADATA_KIND_ID: u32 = 31;
```
nbaksalyar, line 44: I'm not sure this is the correct value - how do I obtain and verify this constant?

TheDan64: See the docs for the method

nbaksalyar: Thanks, I have confirmed that the value is correct :+1:

I also added documentation but unfortunately it doesn't get picked up by `rustdoc` because of `#[cfg(feature = ...)]`. So the rustdoc comment either has to be duplicated for each LLVM version or fixed separately when https://github.com/rust-lang/rust/issues/43781 is stabilised.

tpisto:
I tested this with my M1 Mac using the llvm 14 branch of llvm-sys and it's working great.

To test I used at Cargo.toml

```
[dependencies]
llvm-sys = "140"
inkwell = { git = "https://github.com/nbaksalyar/inkwell.git", branch = "llvm14", features = ["llvm14-0"] }

[patch.crates-io]
llvm-sys = { git = "https://gitlab.com/nikita-baksalyar/llvm-sys.rs.git", branch="llvm-14" }
```

nbaksalyar:
llvm-sys v140 has been published now, so this PR is ready to be reviewed.

I guess CI has to be restarted manually.

src/context.rs
```diff
@@ -225,7 +225,7 @@ impl Context {
     pub fn create_inline_asm(&self, ty: FunctionType, mut assembly: String, mut constraints: String, sideeffects: bool, alignstack: bool, dialect: Option<InlineAsmDialect>,
         can_throw: bool,
     ) -> PointerValue {
-        #[cfg(any(feature = "llvm13-0"))]
+        #[cfg(any(feature = "llvm13-0", feature = "llvm14-0"))]
```
TheDan64, line 228: Isn't this cfg redundant since the fn is 13+?

src/values/metadata_value.rs
TheDan64: Could you link the ticket here in a non doc comment?

TheDan64 (approved):
Would like to see a couple small things cleaned up but looks good otherwise

src/values/metadata_value.rs
nbaksalyar: How does that look?

TheDan64:
Thanks!

PR #324, opened 2022-05-16
Fix warnings by removing unused imports

nbaksalyar:
## Description

Fix warnings by removing unused imports.
Also fixes the build error that's caused by an unused doc comment.

## Checklist

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1055,7 +1055,7 @@ impl<'ctx> ContextRef<'ctx> {
         }
     }
 
-    /// Gets a usable context object with a correct lifetime.
+    // /// Gets a usable context object with a correct lifetime.
     // FIXME: Not safe :(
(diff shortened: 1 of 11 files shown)


TheDan64:
I don't really understand why this fixes the linker error? Could you explain?

nbaksalyar:
> I don't really understand why this fixes the linker error? Could you explain?

Hmm, not sure about the linker error. The main fix is this line here:

```rust
     // /// Gets a usable context object with a correct lifetime.
```

It fixes the [rustdoc error](https://github.com/TheDan64/inkwell/runs/6443197385?check_suite_focus=true#step:4:106).

But while at it, I also fixed a bunch of warnings about unused imports.

TheDan64:
Oh I see, I missed that haha

TheDan64 (approved):
Thanks!

PR #327, opened 2022-05-20
print LLVM Types to LLVMString and implement Display trait

taquangtrung:
## Description

In this PR, I implemented the function `print_to_string` for all existing LLVM Types, which prints the type definition to LLVMString.

This function will be used to implement the `Display` trait for all existing LLVM Types in the folder https://github.com/TheDan64/inkwell/tree/master/src/types

## Related Issue

There is no issue discussing the `print_to_string` before, so maybe we can discuss directly in this PR, instead of creating a new issue?

## How This Has Been Tested

The function `print_to_string` is basically a wrapper function that expose the `print_to_string` of the corresponding underlying type of each LLVM Type, like below:

``` rust
    /// Print the definition of a `FloatType` to `LLVMString`.
    pub fn print_to_string(self) -> LLVMString {
        self.float_type.print_to_string()
    }
```

So, maybe unit test is not necessary?

## Option\<Breaking Changes\>

No breaking changes.

## Checklist

### DIFF
--- a/src/types/enums.rs
+++ b/src/types/enums.rs
@@ -2,12 +2,14 @@ use llvm_sys::core::LLVMGetTypeKind;
 use llvm_sys::prelude::LLVMTypeRef;
 use llvm_sys::LLVMTypeKind;
 
+use crate::support::LLVMString;
 use crate::types::traits::AsTypeRef;
 use crate::types::MetadataType;
 use crate::types::{ArrayType, FloatType, FunctionType, IntType, PointerType, StructType, VectorType, VoidType};
 use crate::values::{BasicValue, BasicValueEnum, IntValue};
 
 use std::convert::TryFrom;
+use std::fmt::{self, Display};
 
 macro_rules! enum_type_set {
     ($(#[$enum_attrs:meta])* $enum_name:ident: { $($(#[$variant_attrs:meta])* $args:ident,)+ }) => (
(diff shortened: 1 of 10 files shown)


src/types/enums.rs
```diff
@@ -507,3 +548,21 @@ impl<'ctx> From<BasicTypeEnum<'ctx>> for BasicMetadataTypeEnum<'ctx> {
...
+
+impl Display for BasicTypeEnum<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.print_to_string())
+    }
+}
+
+impl Display for BasicMetadataTypeEnum<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.print_to_string())
```
TheDan64, line 566: These should probably make mention of the wrapping type, ie `"BasicMetadataTypeEnum::{}", self.print_to_string())` or something like that?

taquangtrung: I think that should the task of the `Debug` trait, by using the marker `"{:?}"`.

For the `Display` trait, the Type names should be printed like how LLVM Types are printed in C++ API.

How do you think about that?

TheDan64: Hmm I suppose that's fair

TheDan64 (approved):
Thanks!

PR #328, opened 2022-06-01
implement print_to_string and Display trait for LLVM Values

taquangtrung:
Hi,

## Description

This PR implements the `Display` trait for the current LLVM Values. This feature is currently missing in the current code base.

## Related Issue

It is related to PR https://github.com/TheDan64/inkwell/pull/327, but implement for LLVM Value instead.

## How This Has Been Tested

The trait Display is implemented by simply calling the function `print_to_string` from the trait `AnyValue`, like below. 

``` rust
use super::AnyValue;
....
impl Display for IntValue<'_> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "{}", self.print_to_string())
    }
}
```
Test cases of the function `print_to_string` are already included in the `tests` directory.

So, maybe more unit tests are not necessary.

## Option\<Breaking Changes\>

No breaking changes.

## Checklist

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -791,7 +791,7 @@ impl<'ctx> Module<'ctx> {
         }
     }
 
-    /// Prints the content of the `Module` to a string.
+    /// Prints the content of the `Module` to an `LLVMString`.
     pub fn print_to_string(&self) -> LLVMString {
         unsafe { LLVMString::new(LLVMPrintModuleToString(self.module.get())) }
     }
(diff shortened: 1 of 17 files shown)


taquangtrung:
@TheDan64 : I have fixed the timeout issue of failed unit tests. 

It turned out that in the previous implementation there is a repeated recursive call to `print_to_string`, caused by my implementation for each LLVM Value and the trait `AnyValue`.

So, I fixed it by removing my implementation of `print_to_string`, and using the one implemented by `AnyValue` to implement the `Display` trait instead.

TheDan64 (approved):
Thanks!

PR #331, opened 2022-06-10
support set_name of InstructionValue, IntValue, etc

taquangtrung:
Hi,

## Description

In this PR, I expose the `set_name` function to several LLVM Values, like InstructionValule, IntValue, etc.

This feature is important when inkwell is used to optimize or transform LLVM bitcode, the users can call `set_name` to rename the optimized/transformed instructions, variables, etc.

## Related Issue

It's a new feature, so I didn't create any issues. Instead, we can discuss the details here if necessary.

## How This Has Been Tested

My implementation is just a wrapper to the function `set_name` in `Value` data structure, 
which is already well tested. So, I didn't write any new unit tests.

## Option\<Breaking Changes\>

No breaking change.

## Checklist


src/values/instruction_value.rs
```diff
@@ -149,6 +149,20 @@ impl<'ctx> InstructionValue<'ctx> {
...
 
+    /// Get name of the `InstructionValue`.
+    pub fn get_name(&self) -> &CStr {
+        self.instruction_value.get_name()
+    }
+
+    /// Set name of the `InstructionValue`.
+    pub fn set_name(&self, name: &str) {
+        if self.get_type().is_void_type() {
+            panic!("Cannot set name of the VoidType instruction: {:?}", self);
```
TheDan64, line 160: Would rather see this return a Result than panic

src/values/instruction_value.rs
```diff
@@ -141,6 +141,20 @@ impl<'ctx> InstructionValue<'ctx> {
         }
     }
 
+    /// Get name of the `InstructionValue`.
+    pub fn get_name(&self) -> &CStr {
+        self.instruction_value.get_name()
```
TheDan64, line 146: I suspect this might return a null ptr if the instruction is a void type

taquangtrung: I tried to check if a `&CStr` is a null pointer, but it seems a null ptr is also a valid `&CStr` type?

What I'm currently doing is to implement another function:

```
    /// Get name of the `InstructionValue` or return a default name.
    pub fn get_name_or_default(&self) -> String {
        match self.get_name().to_str() {
            Ok(name) => name.to_string(),
            _ => "<empty-instruction-name>".to_string(),
        }
    }
```

How about we change the output of the existing function `get_name` to `Option<String>`, like below?

```
    pub fn get_name(&self) -> Option<String> {
         match self.instruction_value.get_name().to_str() {
            Ok(name) => Some(name)
            _ => None
        }
    }
```

In this case, we don't need to care about the instruction type.

TheDan64: We should return an `Option<&CStr>`, that way making an allocation is up to the end user not inkwell

taquangtrung: All fixed as suggested!

TheDan64 (approved):
Thanks!

PR #333, opened 2022-06-17
Added support for and fixes null sections

### LINKED ISSUE #220: `GlobalValue::get_section` can segfault if there is no section on the global

TheDan64:
## Description

Adds support for nullable sections & checks for null ptrs

## Related Issue

Fixes #220 

## How This Has Been Tested

Unit tests updated

## Option\<Breaking Changes\>

Input/Output types changed to Options

## Checklist


TheDan64:
I'm fairly convinced that this just never worked in LLVM 3.x... From what I can tell 3.x would always deref the section ptr without checking if it was set before hand

TheDan64:
Blocked until a decision on #338 is made

Issue #343, opened 2022-08-21
Segmentation Fault when printing `FunctionValue` which lives longer than `Module`


ytoml:
**Describe the Bug**
When trying to print the content of `FuncitonValue` that originates in already dropped `Module`, it leads to segmentation fault.

**To Reproduce**
```rust
use inkwell::context::Context;

fn main() {
    let ctx = Context::create();

    // fn_value lives longer than module.
    let fn_value = 
    {
        let module = ctx.create_module("test");
        let fn_type = ctx.f64_type().fn_type(&[], false);
        let fn_value = module.add_function("test_fn", fn_type, None);

        // This is OK, it prints "define double @test_fn()\n"
        println!("{fn_value}");
        fn_value
    };

    // Segmentation fault
    println!("{fn_value}");
}
```

**Expected Behavior**
Expected to print `"define double @test_fn()\n"` again.
(or print nothing because the module already dropped? Actually I'm not confident because I'm LLVM newbie...)

**LLVM Version (please complete the following information):**
 - LLVM Version: Homebrew LLVM 14.0.6
 - Inkwell Branch Used: [e.g. llvm14-0]

**Desktop (please complete the following information):**
 - OS: macOS 12.5 (with Apple M1, 2020 macbook pro)

**Additional Context**
Nothing special.

ytoml:
I met this bug when following `Kareidoscope` tutorial.
In my first implementation, I wrapped compiling procedure like:
```Rust
fn compile<'ctx>(
    ctx: &'ctx Context,
    mod_name: &str,
    function: &Function,
) -> compiler::Result<FunctionValue<'ctx>> {
    let builder = ctx.create_builder();
    let module = ctx.create_module(mod_name);
    let fpm = PassManager::create(&module);
    fpm.add_instruction_combining_pass();
    fpm.add_reassociate_pass();
    fpm.add_gvn_pass();
    fpm.add_cfg_simplification_pass();
    fpm.initialize();

    Compiler::default()
        .builder(&builder)
        .context(ctx)
        .module(&module)
        .function_pass_manager(&fpm)
        .compile(function)
}
```
and the `Module` instance drops this after this function but `FunctionValue` lives.
Then, printing the returned `FunctionValue` led to segmentation fault.

With `lldb`, I found it triggers `EXC_BAD_ACCESS(code=1, address=0x17)`, but I couldn't figure out further details.
It seems to touch freed memory address after dropping `Module`.

[...]

PR #344, opened 2022-08-24
Replace deprecated functions in LLVM14 with new ones

ytoml:
## Description
All functions deprecated since llvm14 are replaced with recommended alternatives.
Other deprecations (e.g. `LLVMSetCurrentDebugLocation` in `builder.rs`, deprecated since llvm9.0 ) are **untouched**.

## Related Issue
I thought it might fix SEGV in https://github.com/TheDan64/inkwell/issues/343, but actually it doesn't.
However, such deprecation removal is always recommended and seems to have no room to argument, I just have directly opened this issue 😃.

## How This Has Been Tested
I have run test on Ubuntu22.04 LTS (x86_64 VM instance).
Since this is change only related to `llvm14-0` feature, I tested with:
```shell
cargo test --target=x86_64-unknown-linux-gnu --features=llvm14-0 --all
```
All tests and doctests are successfully passed.

## Breaking Changes

It does not break existing APIs, but I added implementation of `AsTypeRef` for `CallableValue` in order to get function type because it's required by `LLVMBuildCall2` and `LLVMBuildInvoke2`.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -130,10 +130,7 @@ impl<'ctx> Builder<'ctx> {
 
         unsafe { InstructionValue::new(value) }
     }
-}
 
-#[llvm_versions(14.0..=latest)]
-impl<'ctx> Builder<'ctx> {
     /// Builds a function call instruction.
     /// [`FunctionValue`]s can be implicitly converted into a [`CallableValue`].
     /// See [`CallableValue`] for details on calling a [`PointerValue`] that points to a function.
@@ -170,7 +167,6 @@ impl<'ctx> Builder<'ctx> {
         F: Into<CallableValue<'ctx>>,
     {
         let callable_value = function.into();
-        let fn_ty_ref = callable_value.as_type_ref();
         let fn_val_ref = callable_value.as_value_ref();
 
         // LLVM gets upset when void return calls are named because they don't return anything
@@ -178,7 +174,10 @@ impl<'ctx> Builder<'ctx> {
 
         let c_string = to_c_str(name);
         let mut args: Vec<LLVMValueRef> = args.iter().map(|val| val.as_value_ref()).collect();
+
+        #[cfg(feature = "llvm14-0")]
         let value = unsafe {
+            let fn_ty_ref = callable_value.as_type_ref();
             LLVMBuildCall2(
                 self.builder,
                 fn_ty_ref,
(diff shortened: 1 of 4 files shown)


src/builder.rs
```diff
@@ -120,7 +130,185 @@ impl<'ctx> Builder<'ctx> {
...
+    /// builder.position_at_end(entry);
+    ///
+    /// let ret_val = builder.build_call(fn_value, &[i32_arg.into(), md_string.into()], "call")
+    ///     .try_as_basic_value()
+    ///     .left()
+    ///     .unwrap();
+    ///
+    /// builder.build_return(Some(&ret_val));
+    /// ```
+    pub fn build_call<F>(&self, function: F, args: &[BasicMetadataValueEnum<'ctx>], name: &str) -> CallSiteValue<'ctx>
```
TheDan64, line 168: Rather than duplicating the entire methods, could we just `llvm_versions`/`cfg` the new calls in the method itself?

ytoml: Actually I prefer that style too, but unfortunately compiler says `custom attributes cannot be applied to statements`...

ytoml: Ah, but we can do it with `#[cfg(not(feature="llvm14-0"))]`. I will fix them.

src/builder.rs
```diff
@@ -164,6 +174,21 @@ impl<'ctx> Builder<'ctx> {
 
         let c_string = to_c_str(name);
         let mut args: Vec<LLVMValueRef> = args.iter().map(|val| val.as_value_ref()).collect();
+
+        #[cfg(feature = "llvm14-0")]
```
TheDan64, line 178: I know it's really ugly but we should future proof these by inverting both this and below with the full expansion:

`#[cfg(not(any(feature = "llvm4-0", ..., feature = "llvm13-0"))]`

and

`#[cfg(any(feature = "llvm4-0", ..., feature = "llvm13-0")]`

respectively. That way future llvm versions will be automatically supported by this code. As you noted, rust doesn't support macro attributes on statements yet. Been waiting a long time for that feature :(

cc: https://github.com/rust-lang/rust/issues/54727

src/values/callable_value.rs
```diff
@@ -90,6 +93,18 @@ impl<'ctx> AsValueRef for CallableValue<'ctx> {
 
 impl<'ctx> AnyValue<'ctx> for CallableValue<'ctx> {}
 
+#[llvm_versions(14.0..=latest)]
```
TheDan64, line 96: I suppose there's no reason to feature gate this impl even if it's only used in 14

ytoml: Removed.

src/builder.rs
ytoml: That's right, I fixed it.

TheDan64 (approved):
Thanks!

ytoml:
Sorry I forgot remove `#[llvm_versions(14.0..=latest)]` from import section in `collable_value.rs` thus test fails.

Issue #282, opened 2021-11-18
Docs are no longer being updated


TheDan64:
**Describe the Bug**
Travis-CI is no longer working so we stopped using it. Docs were built with travis and must be regenerated with the new build system

PR #361, opened 2022-10-05
publish docs for each commit

ruifengx:
## Description

Modified GitHub Actions configuration to publish documentation on every push to `master`. It will only do so if the `tests` action succeeds, so we will not be publishing documentations for broken versions.

## Related Issue

Probably related to #2.

## How This Has Been Tested

The workflow runs correctly in my fork (see the log [here](https://github.com/ruifengx/inkwell/actions/runs/3187875220)), and here is the rendered documentation: <https://ruifengx.github.io/inkwell/>.

**EDIT:** note that I arranged the action to only build for `llvm14-0`. Please check if this is the desired behaviour. Also, the documentation looks very different from the old one, so please confirm it looks correct.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,9 +1,12 @@
-name: Test Suite
+name: Test Suite and Doc
 
 on: [push, pull_request]
 
 env:
   CARGO_TERM_COLOR: always
+  DOC_LLVM_FEATURE: llvm14-0
+  DOC_LLVM_VERSION: '14.0'
+  DOC_PATH: target/doc
 
 jobs:
   tests:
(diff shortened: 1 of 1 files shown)


TheDan64:
Ahh this is so great thank you!! I was having a hard time updating the old docs to the new ci format

TheDan64:
Do you think we can build the docs on nightly? There was one feature which added the different version cfgs to the docs. I'll try and find it

TheDan64:
https://github.com/TheDan64/inkwell/blob/master/internal_macros/src/lib.rs#L330 I think it should just work automatically if the docs get built on nightly

TheDan64:
Closes #282

ruifengx:
I just switched to the nightly channel to build the doc, but unfortunately [I don't see a difference](https://ruifengx.github.io/inkwell). (Though honestly I don't know which file to look, so it might be the case that it is already working.) Is `cargo +nightly doc --features llvm14-0` enough to get the desired effect? Or do I need to use something like `--all-features` (IMO not very likely)?

TheDan64:
Ah, I forgot you need the `nightly` feature flag

ruifengx:
Ah, [it works now](https://ruifengx.github.io/inkwell). Though it does not look very pretty IMO. Here is a screenshot.
Screenshot

TheDan64:
I think it looks better in light mode theme tbh haha. But I think it's much needed for those who aren't running the latest llvm version

TheDan64 (approved):
Thanks!

TheDan64:
hmm, I merged it but it didn't run: https://github.com/TheDan64/inkwell/actions/runs/3208253306
Is it because it's set to push only?

ruifengx:
Yes, I think so. It should run on the next push to master.

TheDan64:
Right but whenever PR's get merged into master, it doesn't run: https://github.com/TheDan64/inkwell/actions/runs/3209299174/jobs/5245894824

TheDan64:
There's not many direct pushes to master

TheDan64:
Oh, wait. This one shows up now: https://github.com/TheDan64/inkwell/actions/runs/3208253306/jobs/5243978472
I'm a little confused bc it wasn't there before lol. So that's good then

Issue #367, opened 2022-11-05
How to get the same opaque struct type?


gavrilikhin-d:
Each call to `context.opaque_struct_type("name")` creates new type. Is it possible to get type that is already present in module?

gavrilikhin-d:
Looks like llvm sys has `LLVMGetTypeByName2` which is lacking in inkwell

gavrilikhin-d:
I'll try to implement it myself and add a pull request

PR #369, opened 2022-11-05
Add method to get type by its name

gavrilikhin-d:
Added `context.get_type("name")` method

## Description

Added new method with documentation and tests to context. It uses `LLVMGetTypeByName2` from llvm_sys, which was missing in inkwell.

## Related Issue

Related to my question: #367 

## How This Has Been Tested

Test when type is found and when not found.

## Checklist

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -945,6 +945,10 @@ impl Context {
     /// # Example
     ///
     /// ```no_run
+    /// use inkwell::context::Context;
+	/// use inkwell::types::AnyTypeEnum;
+    ///
+    /// let context = Context::create();
     ///	let name = "opaque";
     ///	let opaque = context.opaque_struct_type(name);
     ///
@@ -1777,6 +1781,10 @@ impl<'ctx> ContextRef<'ctx> {
     /// # Example
     ///
     /// ```no_run
+    /// use inkwell::context::Context;
+	/// use inkwell::types::AnyTypeEnum;
+    ///
+    /// let context = Context::create();
     ///	let name = "opaque";
     ///	let opaque = context.opaque_struct_type(name);
     ///
--- a/src/context.rs
+++ b/src/context.rs
@@ -1768,6 +1768,24 @@ impl<'ctx> ContextRef<'ctx> {
         self.context.opaque_struct_type(name)
     }
 
+    /// Get type by its name, if it exists.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
+    ///	let name = "opaque";
+    ///	let opaque = context.opaque_struct_type(name);
+    ///
+    /// let got = context.get_type(name);
+    ///	assert_eq!(got, Some(AnyTypeEnum::from(opaque)));
+    ///
+    /// assert_eq!(context.get_type("non-existent"), None);
+    /// ```
+    #[inline]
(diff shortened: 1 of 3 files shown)


gavrilikhin-d:
How to test compilation with every llvm version?

TheDan64:
Unfortunately you'd need to have that llvm version installed to run it. You can test against the CI build, though...

src/context.rs
```diff
@@ -926,6 +961,24 @@ impl Context {
...
+    ///	let name = "opaque";
+    ///	let opaque = context.opaque_struct_type(name);
+    ///
+    /// let got = context.get_type(name);
+    ///	assert_eq!(got, Some(AnyTypeEnum::from(opaque)));
+    ///
+    /// assert_eq!(context.get_type("non-existent"), None);
+    /// ```
+    #[inline]
+    pub fn get_type<'ctx>(&self, name: &str) -> Option<AnyTypeEnum<'ctx>> {
```
TheDan64, line 978: We should add this to ContextRef as well

gavrilikhin-d: Done in 48af9c33817a881eaa797f954bbf02b2f258c9ba

src/context.rs
```diff
@@ -262,6 +262,17 @@ impl ContextImpl {
         unsafe { StructType::new(LLVMStructCreateNamed(self.0, c_string.as_ptr())) }
     }
```
gavrilikhin-d, line 264: There is a `LLVMGetTypeByName` function befor 12.0. Should I add support for it too?

gavrilikhin-d:
There is `LLVMGetTypeByName` function befor `llvm12-0`. Should I support it too?

gavrilikhin-d:
Also, is there a reason why inkwell supports such outdated llvm versions?

TheDan64:
If the api is similar enough then yeah, I suppose we can support both

gavrilikhin-d:
`AnyTypeEnum` in `context.rs` is available only with llvm higher than 12.0

TheDan64:
> AnyTypeEnum in context.rs is available only with llvm higher than 12.0

The use import? The enum exists in older versions, so you could extend it

gavrilikhin-d:
Lol, while looking through code for `Module`, i found `get_struct_type(&self, name: &str) -> Option<StructType<'ctx>>` which does exactly the same

gavrilikhin-d:
I think, adding this function to `Context` after `llvm12-0` would be good, though. But it needs to be renamed to `get_struct_type` too.

gavrilikhin-d:
Are there really only struct types, that can be get by name?

TheDan64:
I'm not sure. I think I recall trying it with other types and it only seemed to work for structs for some reason

TheDan64:
But feel free to write some tests and change it if you can repro non structs

gavrilikhin-d:
I found thatLLVM itself assumes that only structs are named types
```cpp
LLVMTypeRef LLVMGetTypeByName2(LLVMContextRef C, const char *Name) {
    return wrap(StructType::getTypeByName(*unwrap(C), Name));
}
```

gavrilikhin-d:
I tried with `"i8"` and other not-struct types - they may not be found by name

TheDan64 (approved):
Thanks!

Issue #366, opened 2022-11-04
Public constructors for wrapper types


samarsha:
**Is your feature request related to a problem? Please describe.**
It's difficult to interoperate between Inkwell and llvm-sys. This is a problem because Inkwell is not complete, so some things are difficult or impossible to do without calling LLVM APIs directly.

The get_ref() method from LLVMReference gets me only halfway there. If I use an LLVM pointer to create a new LLVM object, like an LLVMValueRef, it's not possible for me to put it back in an Inkwell wrapper. So I can't fill in gaps in Inkwell's functionality by unwrapping, using llvm-sys, and rewrapping. Everything I do with the value afterwards has to be through llvm-sys, even if I would be able to use Inkwell otherwise.

**Describe the solution you'd like**
All `pub(crate)` constructors like

```rust
impl<'ctx> AnyValueEnum<'ctx> {
    pub(crate) unsafe fn new(value: LLVMValueRef) -> Self { ... }
}
```

should be made `pub`.

**Describe possible drawbacks to your solution**
It weakens encapsulation, but I think it's worth it for pragmatism, and Inkwell has already started down that path with get_ref().

**Describe alternatives you've considered**
The only alternative I see is forking Inkwell or using 100% llvm-sys instead of a mixture of Inkwell and llvm-sys.

TheDan64:
It seems like the better solution is to add needed functionality from llvm sys into inkwell. Contributions are always welcome

TheDan64:
For context, inkwell was never intended to interoperate with llvm-sys, but to purely be a wrapper ontop of it

samarsha:
> It seems like the better solution is to add needed functionality from llvm sys into inkwell. Contributions are always welcome

It's definitely good if Inkwell becomes better so that I need llvm-sys less, but I don't agree that it's better for Inkwell to prevent me from using llvm-sys (or any other LLVM library) when needed. By keeping the constructors pub(crate), Inkwell is artificially making itself an all-or-nothing proposition: either I use Inkwell for everything, and wait for PRs for functionality that isn't supported yet, or I don't use Inkwell at all. LLVM changes rapidly and I think it's realistic to admit that Inkwell can't always keep up.

I also think that, even an ideal world where Inkwell always supported 100% of LLVM's functionality perfectly, allowing Inkwell to work with LLVM objects that didn't originate from Inkwell is still good. You might want to combine Inkwell with another LLVM library that works with LLVM objects directly, and doesn't know about Inkwell types.

[...]

PR #372, opened 2022-11-06
Public new

gavrilikhin-d:
Makes types' `new` method public

## Description

Look at the arguments provided in #366 .

TLDR why need this:
1. Frequent API changes in llvm
2. Interoperation with different wrappers

## Related Issue

Implements #366 

## How This Has Been Tested

Runned tests with `cargo test --all --target=aarch64-apple-darwin --features=llvm14-0`

## Checklist

### DIFF
--- a/src/types/array_type.rs
+++ b/src/types/array_type.rs
@@ -18,7 +18,11 @@ pub struct ArrayType<'ctx> {
 }
 
 impl<'ctx> ArrayType<'ctx> {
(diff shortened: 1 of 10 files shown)


gavrilikhin-d:
Should I add some safe wrapper around `new` that will check type and would panic on missmatch? For example, implement `From<LLVMTypeRef>` for types

TheDan64:
I don't think we want From impls because these methods should be unsafe

TheDan64 (approved):

PR #373, opened 2022-11-06
Add `llvm-latest` feature and enable it by default

gavrilikhin-d:
Adds default `llvm-latest` feature.

## Description

Without default specified llvm version feature, completion in IDE doesn't know about `llvm_sys` code.

## How This Has Been Tested

In VSCode. Tested for build with `cargo test --all --target=aarch64-apple-darwin --features=llvm14-0`

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ categories = ["development-tools::ffi"]
 edition = "2021"
 
 [features]
-default = ["target-all"]
+default = ["target-all", "llvm-latest"]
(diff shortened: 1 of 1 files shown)


gavrilikhin-d:
Hm, I see... There is a problem, that after adding llvm-version to default features, default features must be explicitly turned-off, if need to specify other features.

gavrilikhin-d:
Sadly, cargo doesn't have exclusive features support

TheDan64:
Indeed, which is why we have the macro to artificially make the features exclusive

PR #374, opened 2022-11-06
Move get/set section to derive macro

gavrilikhin-d:
Move repeated code for `get_section` and `set_section` for `GlobalValue` and `FunctionValue` to derive macro

## Description

Added new derive macro `Section`.

## Related Issue

First steps to fix test error inside `test_object_file` as in issue #345.

Later `cfg(target_os = "macos")` maybe used inside this single macro, to fix this problem.
I tried to implement this fix with replacing `"section_name"` to `",section_name"` , but now I have test errors instead of LLVM ERROR

## How This Has Been Tested

`cargo test --all  --target=aarch64-apple-darwin --features=llvm14-0`

## Checklist

### DIFF
--- a/src/values/fn_value.rs
+++ b/src/values/fn_value.rs
@@ -7,8 +7,8 @@ use llvm_sys::core::{
     LLVMCountBasicBlocks, LLVMCountParams, LLVMDeleteFunction, LLVMGetBasicBlocks, LLVMGetFirstBasicBlock,
     LLVMGetFirstParam, LLVMGetFunctionCallConv, LLVMGetGC, LLVMGetIntrinsicID, LLVMGetLastBasicBlock, LLVMGetLastParam,
     LLVMGetLinkage, LLVMGetNextFunction, LLVMGetNextParam, LLVMGetParam, LLVMGetParams, LLVMGetPreviousFunction,
-    LLVMGetSection, LLVMIsAFunction, LLVMIsConstant, LLVMSetFunctionCallConv, LLVMSetGC, LLVMSetLinkage,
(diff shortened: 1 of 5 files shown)


internal_macros/src/lib.rs
```diff
@@ -616,3 +616,42 @@ pub fn llvm_enum(attribute_args: TokenStream, attributee: TokenStream) -> TokenS
     };
     q.into()
 }
+
+/// This macro adds shared code to get/set section of global values
+#[proc_macro_derive(Section)]
+pub fn section_derive(input: TokenStream) -> TokenStream {
```
TheDan64, line 622: This is a neat idea but I'm a little concerned that sharing code through derive macros will make the code harder to maintain (for example, if there's a bug here, or if an LLVM fn is deprecated for a newer version, we have to release a new inkwell_internals ver before inkwell)

Our current paradigm is to add methods to the underlying `Value` type, and expose them in the wrapping `*Value` types - I think we should stick to that here

TheDan64 (changes requested):

internal_macros/src/lib.rs
TheDan64: For example: https://github.com/TheDan64/inkwell/blob/master/src/values/float_value.rs#L48

gavrilikhin-d: Done.

gavrilikhin-d:
Also fixed error with sections. Though build on macros is still failing, but without internal error.

gavrilikhin-d:
Looks like code is only tested on x86-64. I think that at least arm should be checked as well

TheDan64 (approved):
Thanks!

Issue #380, opened 2022-11-15
Add `fn_type` to enums


gavrilikhin-d:
**Is your feature request related to a problem? Please describe.**

My compilers dynamically lower types to inkwell's:
```rs
pub fn lower_type(&self, ty: &mir::Type) -> inkwell::types::BasicTypeEnum;
``

Later, it create function with such lowered return type
```rs
self.lower_type(ty).fn_type(params, false) // Not implemented
```
This most obvious approach doesn't work.

**Describe the solution you'd like**
I'd like `BasicTypeEnum`, `BasicMetadataTypeEnum`, `AnyTypeEnum` to implement:
```rs
pub fn fn_type(self, param_types: &[BasicMetadataTypeEnum<'ctx>], is_var_args: bool) -> FunctionType<'ctx>
```
Types from `BasicTypeEnum` may always be used as return type, yet I'm not sure about `MetadataType` and `FunctionType`.

Probably, `BasicMetadataTypeEnum` and `AnyTypeEnum` must change signature to:
```rs
pub fn fn_type(self, param_types: &[BasicMetadataTypeEnum<'ctx>], is_var_args: bool) -> Result<FunctionType<'ctx>, ()>
```

endorpersand:
`BasicTypeEnum` *does* have a fn_type method. However, it's implemented as a part of `BasicType`, so to use `BasicTypeEnum`'s fn_type method you need to import `BasicType` in your module.

As for `MetadataType` and `FunctionType`, the [LLVM language reference](https://llvm.org/docs/LangRef.html#function-type) seems to read that only `void` and first-class types are permitted, which excludes `MetadataType` and `FunctionType`.

PR #381, opened 2022-11-15
Add `fn_type` for type enums

gavrilikhin-d:
## Description

Adds implementation of `fn_type` for type enums, for convenience

## Related Issue

#380 

## How This Has Been Tested

Run doc tests and `cargo test --features llvm14-0`

## Checklist

### DIFF
--- a/src/types/enums.rs
+++ b/src/types/enums.rs
@@ -104,6 +104,32 @@ enum_type_set! {
 }
 
 impl<'ctx> BasicMetadataTypeEnum<'ctx> {
+    /// Creates a [`FunctionType`] with this [`BasicMetadataTypeEnum`] for its return type.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use inkwell::context::Context;
+    /// use inkwell::types::BasicMetadataTypeEnum;
+    ///
+    /// let context = Context::create();
+    /// let ty: BasicMetadataTypeEnum = context.i8_type().into();
+    /// let fn_type = ty.fn_type(&[], false);
+    /// ```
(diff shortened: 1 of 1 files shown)


src/types/enums.rs
```diff
@@ -104,6 +104,66 @@ enum_type_set! {
...
+    pub fn fn_type(self, param_types: &[BasicMetadataTypeEnum<'ctx>], is_var_args: bool) -> FunctionType<'ctx> {
+        use BasicMetadataTypeEnum::*;
+        match self {
+            ArrayType(ty) => ty.fn_type(param_types, is_var_args),
+            FloatType(ty) => ty.fn_type(param_types, is_var_args),
+            IntType(ty) => ty.fn_type(param_types, is_var_args),
+            PointerType(ty) => ty.fn_type(param_types, is_var_args),
+            StructType(ty) => ty.fn_type(param_types, is_var_args),
+            VectorType(ty) => ty.fn_type(param_types, is_var_args),
+            // TODO: are functions really able to return metadata?
```
TheDan64, line 129: We should verify this TODO before adding this function. If it's not valid, it should return a Result

gavrilikhin-d: Ok, I'll check it

gavrilikhin-d: I tried verification of module with function has metadata as return type. It works. But I never found anything about returning metadata on official documentation.

gavrilikhin-d: Any ideas what to do?

TheDan64: If it passed verification then I suppose that's fine

TheDan64:
This needs to be rebased and might need some slight tweaks to support LLVM 15? Not certain

PR #383, opened 2022-11-17
Support llvm-sys 150.x.x

hedgar2017:
## Updates

1. Support the opaque pointers and all related functions.
2. Guard the `LLVMConst*` function family.
3. Guard some optimization passes, including those for coroutines.
4. Update the `AtomicRMWBinOp` enum.
5. Add conditional compilation clauses for some low-hanging functions.
6. Fix the `AddressSpace` enum and make it more generic (breaking).
7. Add functions for parsing LLVM IR from the memory and files.
8. Add the function for parsing command line options.
9. The `bitcast` operation now returns `AnyType` to support function pointer casts.
10. Some minor fixes and improvements.
11. Updated some dependencies.

## Notes

1. I decided not to continue nesting in the functions which require opaque pointers, but instead implement another set of function with the suffix `_2`. It's not a perfect solution, but the alternative one would be worse from my point of view.
2. The PR wasn't yet sanitized with `clippy`, and some tests are probable missing. Let's discuss. Honestly I feel like the project hasn't been sanitized in multiple previous PRs as well.

## Related issues

https://github.com/TheDan64/inkwell/issues/359

Supersedes [this PR](https://github.com/TheDan64/inkwell/pull/365)

- [ x ] I have read the [Contributing Guide](https://github.com/TheDan64/inkwell/blob/master/.github/CONTRIBUTING.md)

PR #385, opened 2022-12-18
impl Default for AddressSpace to make future changes easier

folkertdev:
in #383 it came up that the current representation of an address space is not entirely correct. To make changes to that type easier, and because usually "Generic" is the address space that users want, implementing and using the default instance means that even if we change the representation of address spaces, this would not actually be noticeable by users.

## \<Breaking Changes\>

There specifically are no breaking changes in the PR

## Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -118,6 +118,12 @@ pub enum AddressSpace {
     Local = 5,
 }
 
+impl Default for AddressSpace {
+    fn default() -> Self {
+        AddressSpace::Generic
+    }
+}
+
 impl TryFrom<u32> for AddressSpace {
     type Error = ();
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,27 +107,31 @@ assert_unique_used_features! {"llvm4-0", "llvm5-0", "llvm6-0", "llvm7-0", "llvm8
 
 /// Defines the address space in which a global will be inserted.
 ///
-/// The default address space is zero. The first six address spaces are available as associated
-/// constants. To convert any higher number, use the `TryFrom` instance. An Address space is a
-/// 24-bit number.
+/// The default address space is zero. An address space can always be created from a `u16`:
+/// ```no_run
+/// inkwell::AddressSpace::from(1u16);
+/// ```
+///
+/// An Address space is a 24-bit number. To convert from a u32, use the `TryFrom` instance
+///
+/// ```no_run
+/// inkwell::AddressSpace::try_from(42u32).expect("fits in 24-bit unsigned int");
+/// ```
 ///
 /// # Remarks
 /// See also: https://llvm.org/doxygen/NVPTXBaseInfo_8h_source.html
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub struct AddressSpace(u32);
 
-impl AddressSpace {
-    pub const ZERO: Self = Self(0);
-    pub const ONE: Self = Self(1);
-    pub const TWO: Self = Self(2);
-    pub const THREE: Self = Self(3);
-    pub const FOUR: Self = Self(4);
-    pub const FIVE: Self = Self(5);
-}
-
 impl Default for AddressSpace {
     fn default() -> Self {
         AddressSpace(0)
+    }
+}
+
+impl From<u16> for AddressSpace {
+    fn from(val: u16) -> Self {
+        AddressSpace(val as u32)
     }
 }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -107,14 +107,40 @@ assert_unique_used_features! {"llvm4-0", "llvm5-0", "llvm6-0", "llvm7-0", "llvm8
 
 /// Defines the address space in which a global will be inserted.
 ///
+/// The default address space is zero. The first six address spaces are available as associated
+/// constants. To convert any higher number, use the `TryFrom` instance. An Address space is a
+/// 24-bit number.
+///
 /// # Remarks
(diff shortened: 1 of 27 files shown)


folkertdev:
cc @hedgar2017

hedgar2017:
@folkertdev I think we must remove this enum completely and use u32 instead.

hedgar2017:
@folkertdev each target has its own understanding and treatment of address spaces. For example for our virtual machine it's:
0 - stack
1 - the 1st heap
2 - the 2nd heap
3 - arbitrary memory page
4 - the space of constants
and so on.

Some targets may use a large number of address spaces, so someone will eventually be blocked by this enum anyway.

hedgar2017 (changes requested):
Let's remove the enum?
It's already casted to u32 everywhere.

folkertdev:
I understand that you want to change the representation. That is however breaking for pretty much every user of inkwell. so this PR would be a step towards making that process easier, because usually you want to use the default address space.

Anyway, I don't know exactly what the policy is on breaking changes, that is ultimately up to @TheDan64 to decide. This PR is safe to merge no matter what, since it is strictly additive.

hedgar2017:
@folkertdev I understand, but it's not a good reason to stockpile such blocking and misleading technical debt.

Concerning this patch, adding this default may be reasonable for your particular target, but for a lot of other targets the 0-th address space is not more default than others.

folkertdev:
well, https://llvm-swift.github.io/LLVMSwift/Structs/AddressSpace.html says

> An address space is an identifier for a target-specific range of address values. An address space is a fundamental part of the type of a pointer value and the type of operations that manipulate memory.
> 
> LLVM affords a default address space (numbered zero) and places a number of assumptions on pointer values within that address space:
> 
> - The pointer must have a fixed integral value
> - The null pointer has a bit-value of 0

so out of all the address spaces, I think `0` is objectively the most default. For common targets, it is also probably the only one you'll ever use

hedgar2017:
@folkertdev well, you're right, I indeed found this in the official doc.
It's up to @TheDan64 of course.

hedgar2017 (approved):

TheDan64:
Has anyone confirmed 5+ are valid values LLVM accepts for address spaces?

TheDan64:
I'm ok with converting to a u32 if so

folkertdev:
well, it turns out that the address space is really as 24-bit integer. in other words, this test passes 

```rust
#[test]
fn test_ptr_address_space() {

    let spaces = [0, 1, 2, 3, 4, 5, 6, 1 << 24 - 1];

    for index in spaces {
        let address_space = AddressSpace(index);

        let ptr = context.i32_type().ptr_type(address_space);
        assert_eq!(ptr.get_address_space(), address_space);
    }

    let address_space = AddressSpace(1 << 24);
    let ptr = context.i32_type().ptr_type(address_space);
    assert_ne!(ptr.get_address_space(), address_space);
}
```

Not sure what to do with that. We could pick `u16` internally, and assume that is good enough (really how many address spaces can you use in practice). Using `u32` is incorrect I think and using it would not align with Inkwell's goals.

hedgar2017:
@folkertdev thanks for checking! By the way, we can use `2^24 - 1` address spaces in practice. If you don't use them, it doesn't mean that nobody else needs it.

I propose we make some kind of `AddressSpace(u32)` wrapper with an assertion in the constructor that it's not bigger than `2^24 - 1`. This way we'll fully cover the functionality LLVM provides and still keep it safe and ugly-low-level-crash-free.

hedgar2017 (changes requested):
Oh, thanks a lot for this change.
Please encapsulate the inner value with an `address_space < 2^24` assertion and I believe it's ready for merging!

src/lib.rs
```diff
@@ -107,28 +107,39 @@ assert_unique_used_features! {"llvm4-0", "llvm5-0", "llvm6-0", "llvm7-0", "llvm8
...
-    Local = 5,
+pub struct AddressSpace(u32);
+
+impl AddressSpace {
+    pub const ZERO: Self = Self(0);
+    pub const ONE: Self = Self(1);
+    pub const TWO: Self = Self(2);
+    pub const THREE: Self = Self(3);
+    pub const FOUR: Self = Self(4);
+    pub const FIVE: Self = Self(5);
```
hedgar2017, line 125: These constants are probably redundant. Do you need them?

hedgar2017 (changes requested):

src/lib.rs
hedgar2017: Another alternative is not using magic numbers and having an enum in your own front-end with proper descriptive names for each address space.
`inkwell::AddressSpace::try_from(crate::AddressSpace::Stack)?` is not ugly.

hedgar2017 (changes requested):
I believe we should remove the constants, because their names are not descriptive like @folkertdev himself said a while ago.
The inkwell users must define their AddressSpace logic outside of this crate.

folkertdev:
I disagree. you would force functions upon your users that look like they can panic, but they actually can't. That is disappointing API design.

hedgar2017:
I didn't get what the constants had to do with panics.
No big deal for me, it's up to @TheDan64

folkertdev:
added an instance for `From<u16>` (that means `::from(3)` works and is not ambiguous. `u8` can always (and conveniently) be upcast to `u16`, and removed the constants.

hedgar2017:
@folkertdev has your commit passed through? I see the last one 21h ago.

folkertdev:
should be good now, I think this is ready.

hedgar2017 (approved):

TheDan64 (approved):
Thanks!

Issue #390, opened 2023-01-20
`Context::const_string` returns incorrect type


endorpersand:
**Describe the Bug**
`Context::const_string` returns a `VectorValue`, when it is supposed to return an `ArrayValue`.

**To Reproduce**
The example in the docs clearly shows that the function does not return a `VectorValue`:
```rs
use inkwell::context::Context;
use inkwell::values::AnyValue;

let context = Context::create();
let string = context.const_string(b"my_string", false);

assert_eq!(string.print_to_string().to_string(), "[9 x i8] c\"my_string\"");
```

**Expected Behavior**
`Context::const_string` should return an `ArrayValue`.

**LLVM Version (please complete the following information):**
 - LLVM Version: 14.0.6
 - Inkwell Branch Used: llvm14-0

**Desktop (please complete the following information):**
 - OS: macOS 13.0.1

TheDan64:
Good catch, would you like to file a PR?

PR #391, opened 2023-01-21
Treat `const_string` as returning `ArrayValue`

endorpersand:
## Description

This changes the type signature of `Context::const_string` to return ArrayValue instead of VectorValue.

This alone breaks the `test_values::test_string_values` test, 
so I also added the `is_const_string` and `get_string_constant` methods from VectorValue to ArrayValue.

I didn't remove those methods from VectorValue because I wanted to avoid making any breaking changes. I'm not too sure if they are needed, though.

## Related Issue

#390 

## How This Has Been Tested

- Doc tests
- `cargo test -F llvm14-0`

## Checklist

### DIFF
--- a/src/values/array_value.rs
+++ b/src/values/array_value.rs
@@ -1,4 +1,4 @@
-use llvm_sys::core::{LLVMIsAConstantArray, LLVMIsAConstantDataArray};
+use llvm_sys::core::{LLVMIsAConstantArray, LLVMIsAConstantDataArray, LLVMIsConstantString, LLVMGetAsString};
 use llvm_sys::prelude::LLVMValueRef;
 
 use std::ffi::CStr;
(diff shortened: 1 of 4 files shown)


src/values/array_value.rs
```diff
@@ -82,6 +82,51 @@ impl<'ctx> ArrayValue<'ctx> {
...
+    /// ```
+    // SubTypes: Impl only for ArrayValue<IntValue<i8>>
+    pub fn get_string_constant(&self) -> &CStr {
+        // REVIEW: Maybe need to check is_const_string?
+
+        let mut len = 0;
+        let ptr = unsafe { LLVMGetAsString(self.as_value_ref(), &mut len) };
+
+        if ptr.is_null() {
+            panic!("FIXME: Need to retun an Option");
```
TheDan64, line 125: Let's just return an option

TheDan64:
If they don't make sense on vector any more, we should remove them

endorpersand:
The thing about the VectorValue methods is that the LLVM doxygen pages seem to indicate that the functions *would* work on vectors ([ConstantDataSequential](https://llvm.org/doxygen/classllvm_1_1ConstantDataSequential.html)), it's just that `LLVMConstantStringInContext` creates an array. 

I've removed them, but it's possible that there is some usecase for them?

TheDan64 (approved):

Issue #396, opened 2023-02-01
MacOS x86_64 fail with Section not found error, seems setting custom section name is broken


ptitSeb:
**Describe the Bug**

We use inkwell on the wasmer project. After an update from the 0.1.0-beta4 to latest 0.1.0, the build on macOS x86_64 is failing.
Wasmer create a custom section named `__TEXT,wasmer_function` and it seems this section is not found durring the link.

This is probably due to this commit https://github.com/TheDan64/inkwell/commit/3917e3207941c1a90f28df4fe08ffdd26de71575

It should be noted that only the x86_64 macOS platform is impacted. All other platform are fine, including ARM64 macOS.

**To Reproduce**
No easy way to reproduce. The log can be found here: https://github.com/wasmerio/wasmer/actions/runs/4053796572/jobs/6993591076 but it's quite a big log.

TheDan64:
Any idea what'd be the correct fix? I don't have access to a macOS machine, so I have no way to test

ptitSeb:
I would limit the new code of the commit to only macos aarch64 (instead of just macos), as the old code was working fine on macos x86_64.

TheDan64:
Ok, you're welcome to make that change and test it out if you'd like

ptitSeb:
Mmmm, I retested and in fact, it break Wasmer both on x86_64 and aarch64. I think the issue is that, in the case of Wasmer, we add `__TEXT,wasmer_function`, so a section name that already contains a `,`, and in that case it should not be added to the name.
I'll work on a solution that test if the section name already have a `,` in it or not

PR #398, opened 2023-02-07
Fix macOS section handling that already have a comma in ther name

ptitSeb:
## Description

Following the commit [#3917e32](https://github.com/TheDan64/inkwell/commit/3917e3207941c1a90f28df4fe08ffdd26de71575) section have a `,` being added to their name on macOS. On some certain cases, that `,` should not be added because the name is already fully qualified, like `__TEXT,wasmer_function`.
This PR check if section name already contains a comma before adding it.

## Related Issue

address issue #396 

## How This Has Been Tested

Tested with latest Wasmer on macOS , both x86_64 and aarch64 arch.

## Checklist

### DIFF
--- a/src/values/mod.rs
+++ b/src/values/mod.rs
@@ -186,7 +186,14 @@ impl<'ctx> Value<'ctx> {
 
         // On MacOS we need to remove ',' before section name
         if cfg!(target_os = "macos") {
-            Some(unsafe { CStr::from_ptr(ptr.add(1)) })
+            let name = unsafe { CStr::from_ptr(ptr) };
+            let name_string = name.to_string_lossy();
+            let mut chars = name_string.chars();
+            if Some(',') == chars.next() {
+                Some(unsafe { CStr::from_ptr(ptr.add(1)) })
+            } else {
+                Some(name)
+            }
         } else {
             Some(unsafe { CStr::from_ptr(ptr) })
         }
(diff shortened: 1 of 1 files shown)


src/values/mod.rs
```diff
@@ -195,7 +202,7 @@ impl<'ctx> Value<'ctx> {
     /// Sets the section of the global value
     fn set_section(self, section: Option<&str>) {
         #[cfg(target_os = "macos")]
-        let section = section.map(|s| format!(",{}", s));
+        let section = section.map(|s| if s.contains(",") { format!("{}", s) } else { format!(",{}", s) });
```
TheDan64, line 205: Should this be starts_with?

ptitSeb: No. the idea is if a section name is just "mystuff", it needs to be set as ",mystuff", but if it's "XXX,mystuff", then don't touch it.
Section are expected to be in the format "<segment name>,<section name>", and it seems a no segment-name is acceptable, that why the previous commit added the ",".

syrusakbary:
@TheDan64 is there any blocker for merging the PR? We would love to use it

TheDan64:
I can take a look at merging this and releasing a v0.1.1 this weekend

TheDan64 (approved):

BigBadE:
Does this PR even build? This shadows "section" to be an Option<String> instead of an Option<&str>, which can't be passed to to_c_str.

Issue #408, opened 2023-04-15
Upgrade Syn to 2.0


Pat-Lafon:
**Is your feature request related to a problem? Please describe.**

I would like upgrade Inkwell's dependency of Syn to the next major release, 2.0. As other projects have updated, it's quite unfortunate to have 2 versions of Syn in my project dependency tree.

**Describe the solution you'd like**

I would like to upgrade the Syn version to 2.0 and fix any of the breaking changes. I'll open a pr but the template says I should have an issue first.

**Describe possible drawbacks to your solution**
How might this solution hinder other components or users if implemented?

If some projects are still on Syn 1.0, then they will have multiple versions of Syn in their dependency tree.

Syn still supports a msrv of 1.56 but if this were to increase then that might affect Inkwells minimum version. I'm not 100% sure about how/when this could occur but I'll note it here.

**Describe alternatives you've considered**

One could not upgrade but with major crates like clap doing so I'm not sure why one wouldn't.

**Additional context**

PR #409, opened 2023-04-15
Syn2

Pat-Lafon:
## Description

Updates inkwell's `internal_macros` to use Syn 2.0. This involves addressing any breaking changes.

I also addressed some clippy lints in `internal_macros`. Given the contributor guidelines, I was surprised to not see a clippy check in the workflow.

## Related Issue

#408

## How This Has Been Tested

I will be running tests via github workflow.

## Option\<Breaking Changes\>

I don't believe updating the internal Syn dependency is a breaking change but feel free to correct me if otherwise.

## Checklist

### DIFF
--- a/internal_macros/Cargo.toml
+++ b/internal_macros/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.7.0"
 [dependencies]
 proc-macro2 = "1.0"
 quote = "1.0"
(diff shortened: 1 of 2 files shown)


TheDan64 (approved):
Good catch, thanks!

PR #426, opened 2023-06-26
add test for inlinealways function attribute

folkertdev:
## Description

Improve test for attributes on functions. Attributes on the `Function` location were not tested at all. The `alignstack` attribute was applied to the return type, but this attribute is not actually valid in that position. A `verify()` on the module called this issue out. 

## Checklist

### DIFF
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -17,9 +17,68 @@ use crate::types::AnyTypeEnum;
 // REVIEW: Should Attributes have a 'ctx lifetime?
 /// Functions, function parameters, and return types can have `Attribute`s to indicate
 /// how they should be treated by optimizations and code generation.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy)]
 pub struct Attribute {
     pub(crate) attribute: LLVMAttributeRef,
+}
+
+impl std::fmt::Debug for Attribute {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        if self.is_string() {
+            return f
+                .debug_struct("Attribute::String")
+                .field("ptr", &self.attribute)
+                .field("kind_id", &self.get_string_kind_id())
+                .field("value", &self.get_string_value())
+                .finish();
+        }
+
+        if self.is_enum() {
+            return f
+                .debug_struct("Attribute::Enum")
+                .field("ptr", &self.attribute)
+                .field("kind_id", &self.get_enum_kind_id())
+                .field("value", &self.get_enum_value())
+                .finish();
+        }
+
+        if self.is_type() {
+            return f
+                .debug_struct("Attribute::Type")
+                .field("ptr", &self.attribute)
+                .field("kind_id", &self.get_enum_kind_id())
+                .field("value", &self.get_type_value())
+                .finish();
+        }
+
+        unreachable!(
+            "attribute at {:?} is not a string, enum or type attribute",
+            self.attribute
+        );
+    }
+}
+
+impl Eq for Attribute {}
+
+impl PartialEq<Self> for Attribute {
+    fn eq(&self, other: &Self) -> bool {
+        if self.is_enum() && other.is_enum() {
+            return self.get_enum_kind_id() == other.get_enum_kind_id()
+                && self.get_enum_value() == other.get_enum_value();
+        }
+
+        if self.is_string() && other.is_string() {
+            return self.get_string_kind_id() == other.get_string_kind_id()
+                && self.get_string_value() == other.get_string_value();
+        }
+
+        if self.is_type() && other.is_type() {
+            return self.get_enum_kind_id() == other.get_enum_kind_id()
+                && self.get_type_value() == other.get_type_value();
+        }
+
+        self.attribute == other.attribute
+    }
 }
 
 impl Attribute {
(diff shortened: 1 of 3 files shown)


tests/all/test_attributes.rs
```diff
@@ -78,56 +80,78 @@ fn test_attributes_on_function_values() {
...
-    let void_type = context.void_type();
-    let i32_type = context.i32_type();
-    let fn_type = void_type.fn_type(&[i32_type.into()], false);
+    let i32_ptr_type = context.i32_type().ptr_type(AddressSpace::default());
+    let fn_type = i32_ptr_type.fn_type(&[i32_ptr_type.into()], false);
     let fn_value = module.add_function("my_fn", fn_type, None);
     let entry_bb = context.append_basic_block(fn_value, "entry");
     let string_attribute = context.create_string_attribute("my_key", "my_val");
-    let alignstack_attribute = Attribute::get_named_enum_kind_id("alignstack");
-    let enum_attribute = context.create_enum_attribute(alignstack_attribute, 1);
```
folkertdev, line 88: `alignstack` is not a valid attibute on a function return type. At least for llvm 15, any value beside 0 (1 on line 88 here) leads to issues. The attribute is still applied when that value is 0. I think the value is only useful for some attributes (e.g. `dereferencable(n)`, where `n` is the enum value).

to apply the `noalias` attribute instead, the function return type must  be a pointer type.

src/attributes.rs
```diff
@@ -17,11 +17,70 @@ use crate::types::AnyTypeEnum;
 // REVIEW: Should Attributes have a 'ctx lifetime?
 /// Functions, function parameters, and return types can have `Attribute`s to indicate
 /// how they should be treated by optimizations and code generation.
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy)]
 pub struct Attribute {
     pub(crate) attribute: LLVMAttributeRef,
 }
 
+impl std::fmt::Debug for Attribute {
```
folkertdev, line 25: I used the Debug and PartialEq instance during debugging. I can remove them again, but then when something fails you only see an address, which is not helpful at all.

folkertdev:
is CI supposed to work for LLVM 16? it seems to fail before it even tries to build/run tests, like LLVM 16 is just not installed on the CI machine.

TheDan64:
Yes, 16 was working before. Not sure what changed

folkertdev:
well I'm not sure what changed then. The branch is rebased on current `TheDan64:master` so it should be using whatever master is using.

TheDan64:
@FawazTirmizi do you think you'd be able to look into LLVM16 suddenly failing in CI?

folkertdev:
gently bumping this

TheDan64:
Hi, sorry for the delay. The llvm16 build will have to get fixed before this can be merged

TheDan64:
Yeah, I'll rerun it but I'm not confident it'll fix it

TheDan64:
Been looking into this but it's not really clear why it fails for the latest version but not the others..

TheDan64:
Been wondering if llvm-sys changed something in 160

folkertdev:
I added a call to `llvm-config` to the CI script, and it prints

```
Run llvm-config --version --bindir --libdir
llvm-config: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by llvm-config)
llvm-config: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by llvm-config)
llvm-config: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by llvm-config)
llvm-config: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by llvm-config)
```

that makes sense actually, the LLVM install action installs `clang+llvm-16.0.4-x86_64-linux-gnu-ubuntu-22.04.tar.xz`. So I think the VM that we run this on uses a version of ubuntu that is too old for the LLVM build that is used.

folkertdev:
and it just looks like for llvm `16.04` ubuntu 22.04 is the only one for which a pre-built tar exists. For earlier point releases it was ubuntu 20.04 still. 

I'm not sure what the best approach is here. I suspect we can configure this in github CI using something like this maybe
```diff
+        # only use ubuntu-22.04 for llvm 16 and higher
+        include:
+          - os: ubuntu-22.04
+            llvm-version: ["16.0", "16-0"]
+        exclude:
+          - os: ubuntu-20.04
+            llvm-version: ["16.0", "16-0"]
```

not pretty, but might do the trick. I have no idea if this is just a fluke or if llvm (in a point release ?!) decided do bump its minimum supported ubuntu version?

an alternative is to force an earlier version of llvm. `16.01` is the latest one that has a 20.04 build

TheDan64:
ugh, that's really annoying if they bumped the ubuntu version in a point release of all times

folkertdev:
for point releases, the builds they provide are really spotty. The later point releases have no binaries for ubuntu at all...

so what is the best option here

1. use 22.04 for llvm 16 only (I assume we want to keep testing with an older OS where possible)
2. pin the llvm 16 version to 16.0.1, so ubuntu 20.04 can be used for everyone

TheDan64:
If all the older versions work on u22 then we should just update to that. I'm assuming that isn't the case, so we probably want to make newer versions start using u22 but keep the older ones on u20

folkertdev:
allright, got it to work. Turns out there is a way to do this sort of thing compactly

TheDan64:
Oh wow, thank you for looking into it!

TheDan64 (approved):
Thank you!

Issue #425, opened 2023-06-19
Segfaults when builder is called without setting its position first


jdupak:
**Describe the Bug**
Program compiles successfully but segfaults, if builder position is not set to some BB.

**To Reproduce**

```rust
        let main = module.add_function("main", ctx.i64_type().fn_type(&[], false), None);
        let entry = ctx.append_basic_block(main, "entry");
        // builder.position_at_end(entry); without this is segfaults
        let exit_code = builder.build_alloca(ctx.i64_type(), "exit_code");
        let exit_bb = ctx.append_basic_block(main, "exit");
```

**Expected Behavior**
This should at least produce a "controled" runtime error.

**LLVM Version (please complete the following information):**
 - LLVM Version: 11, 14
 - Inkwell Branch Used: crates.io release 0.2.0

**Desktop (please complete the following information):**
 - OS: Manjaro

TheDan64:
I wanna say this is a known issue but I can't find a ticket for it lol.

Not sure the LLVM api has a safe way for us to verify if a position has been set

marcospb19:
I was about to open an issue, but found this one.

```rust
use inkwell::context::Context;

fn main() {
    let context = Context::create();
    let builder = context.create_builder();

    // [1] 70215 segmentation fault (core dumped)
    builder.build_alloca(context.i64_type(), &"n");
}
```

```toml
inkwell = { version = "0.2.0", features = ["llvm15-0"] }
```

EricLBuehler:
> if builder position is not set to some BB.

Perhaps Inkwell can store a state of BB, so that it has a runtime error? Or better yet, a type-state of some sort?

TheDan64:
Yeah, I've thought about using type state for this before but we'd have to figure out which methods are valid for each state and there are a lot of builder methods

EricLBuehler:
I agree, it would be complicated to implement type state.

To implement type state based safety, perhaps a builer pattern can be implemented for the `Builder` struct. With this idea, the only way a "buildable"  `Builder` can be created is by calling one of the `position_*` methods on a "non-buildable" `Builder`? This would return a "buildable"  `Builder` and would essentially be a `build` method for the builder pattern.

[...]

PR #436, opened 2023-08-30
Add runtime checks for builder position

EricLBuehler:
I added an `assert_eq` to check that the builder has a position set in every `build_*` method, which improves safety.

## Description

I added an `enum` to keep track of the state, which was added to the `Builder` struct. The state on construction is `NotSet`, and the `position_*` methods are used to set the state to `Set`. I factored out the raw `assert_eq` into a macro to improve code readability.

## Related Issue

This resolves the problem discussed in #425.

## How This Has Been Tested

I ran the tests locally on my WSL2 Ubuntu 22.04, as well as the Github Actions on my fork, all of which passed.

## Checklist


EricLBuehler:
Note that this PR fixes a safety bug with Inkwell as described in #425.

Cargo.toml
```diff
@@ -140,6 +140,7 @@ llvm-sys-160 = { package = "llvm-sys", version = "160.1.0", optional = true }
 once_cell = "1.16"
 parking_lot = "0.12"
 static-alloc = { version = "0.2", optional = true }
+trc = "1.2.2"
```
TheDan64, line 143: What's the purpose of trc if Builder isn't usable in a multi threaded context and isn't clone?

src/builder.rs
```diff
@@ -102,6 +117,7 @@ impl<'ctx> Builder<'ctx> {
     /// builder.build_return(Some(&i32_arg));
     /// ```
     pub fn build_return(&self, value: Option<&dyn BasicValue<'ctx>>) -> InstructionValue<'ctx> {
+        check_positioned!(self);
```
TheDan64, line 120: Rather than asserting, I think we should return a Result to indicate an error has occurred

TheDan64: It'd be good to start using the `thiserror` crate. Could add a `BuilderError` type

EricLBuehler: I agree - that would make the check more visible. However, it would be an API change, and users may need to update large amounts of their code to either handle an `Err` or extract the `Ok` value.

Of course, maybe this is a good thing, but how often do programmers make the mistake of not setting a position?

Cargo.toml
EricLBuehler: I used `Trc` because I needed a layer of indirection to modify the `PositionState` even if there is only a `&self`. Because I needed to use `Trc::as_ptr` and an unsafe block, I wanted to be sure about the intricacies of the implementation so that safety is upheld.

src/builder.rs
TheDan64: Being a pre-1.0 crate, we aren't making any API stability guarantees. We should go ahead and do this. If people want the old functionality, unwrapping will achieve the same result

Cargo.toml
TheDan64: It seems like we could just avoid the unsafety with `Cell` or `AtomicBool`

EricLBuehler: True, I will implement that.

src/builder.rs
```diff
@@ -60,6 +61,26 @@ enum PositionState {
     Set,
 }
 
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum BuilderError {
+    #[error("Builder position is not set")]
+    PositonError,
```
TheDan64, line 68: typo: Position

TheDan64: Maybe `UnsetPosition` would be a better name? It's sorta redundant to have Error in the variant name

EricLBuehler: True. Changing that now - and I will also run `typos` over Inkwell. Perhaps adding `typos` can be added to CI?

TheDan64 (approved):
This looks great; thank you for all of the great work and cleanup!

EricLBuehler:
Glad to help!

src/builder.rs
TheDan64: I didn't know about `typos` before - could you open a ticket please?

EricLBuehler: Yes, will open it now.

src/builder.rs
```diff
@@ -51,11 +52,47 @@ use crate::values::{
...
+    Set,
+}
+
+/// Errors that can be generated by the Builder. All `build_*` methods return a `Result<_, BuilderError>`, which must be handled.
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum BuilderError {
+    #[error("Builder position is not set")]
+    UnsetPosition,
+    #[error("GEP error: does not point to a struct or index is out of bounds")]
+    GEPError,
```
marcospb19, line 70: Does this solve #213?

marcospb19, line 70: This is called `GEPError`, in my head it goes like "any GEP error goes into this variant", but there are actually 2 other GEP error variants.

I'd say this one should have a different name or be broken down into more variants.

WDYT? Is this worth an issue?

EricLBuehler: I don't think so. The `GEPError` I added just replaces an `Err<'static str>` that was returned before #436 was merged. Could a `BuilderError` or something similar could be returned for #214?

EricLBuehler: I removed the `GEPError` on my fork. Turns out, `GEPPointee` and `GEPIndex` are all that were needed.

EricLBuehler: Should I reopen another PR?

marcospb19: Oh nice, I'd say yes! But we can wait for TheDan to be 100% sure.

Issue #419, opened 2023-05-13
Consider replacing parking_lot::RwLock with the lock from the standard library


yorickpeterse:
This type is used at https://github.com/TheDan64/inkwell/blob/41d335bccac4b1c2554a81aa7024e2b6018e6e48/src/targets.rs#L141. For such a simple scenario, the RwLock type from the standard library should be more than sufficient. In addition, it can reduce the number of dependencies (including indirect ones) by up to 8 dependencies, depending a bit on which ones are already in your dependency graph (or not). This in turn can speed up compile times a bit, but it's mostly because using parking_lot for _just_ this scenario is overkill.

If replacing parking_lot is fine I'll happily set up a pull request :smile:

TheDan64:
The reason for using parking lot here is that we don't care about poisoning, and wouldn't want another thread to crash (it seems idiomatic to just unwrap the std Mutex) on poison, nor do we really want to add extra err handling for that case

yorickpeterse:
@TheDan64 The only place this lock is used in various `Target` method. The lock itself is private, and only acquired briefly when calling LLVM's C functions. As far as I can tell from looking at the code, there's no way Inkwell can actually run into a panic while holding a lock. As such I don't see a way of the lock ever getting poisoned.

Even if it gets poisoned it might not matter, as I believe you can basically just ignore that using `PoisonError::into_error`, meaning you'd end up with something like this:

```rust
let mut guard = lock.write().unwrap_or_else(|e| e.into_inner());
*guard = ...;
```

So to summarise: if poisoning is the sole reason for using parking_lot, then I think there are easy enough ways to work around that, without the need for extra dependencies :smiley:

TheDan64:
We also use it for the global Context

yorickpeterse:
@TheDan64 In that case the snippet I shared should be sufficient to deal with the rare case that a process panics, but the developer wants different threads using Inkwell to continue running.

TheDan64:
Forgot to mention - feel free to PR

yorickpeterse:
@TheDan64 I'll take a look at it this week :smile:

PR #442, opened 2023-09-08
Replace parking_lot with the standard library

yorickpeterse:
## Description

Historically parking_lot's locking types performed better, but since a few Rust releases ago the types provided by the standard library perform well enough. Replacing parking_lot also reduces the number of dependencies.

This fixes https://github.com/TheDan64/inkwell/issues/419.

## Related Issue

https://github.com/TheDan64/inkwell/issues/419

## How This Has Been Tested

By running `cargo test --features llvm16-0` and the equivalent `cargo check` command.

## Checklist

### DIFF
--- a/src/targets.rs
+++ b/src/targets.rs
@@ -17,7 +17,7 @@ use llvm_sys::target_machine::{
     LLVMTargetRef,
 };
 use once_cell::sync::Lazy;
-use parking_lot::RwLock;
+use std::sync::RwLock;
 
 use crate::context::AsContextRef;
 use crate::data_layout::DataLayout;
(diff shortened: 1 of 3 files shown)


src/targets.rs
```diff
@@ -167,32 +167,32 @@ impl Target {
         };
 
         if config.base {
-            let _guard = TARGET_LOCK.write();
+            let _guard = TARGET_LOCK.write().unwrap_or_else(|e| e.into_inner());
```
yorickpeterse, line 170: I suppose this technically could be left as-is (i.e. without the unwrap), but I'm not a fan of just ignoring `Result` types.

TheDan64 (approved):

PR #445, opened 2023-09-29
Add `try_load_library_permanently`

gavrilikhin-d:
## Checklist

- [ ] I have read the [Contributing Guide](https://github.com/TheDan64/inkwell/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/src/support/mod.rs
+++ b/src/support/mod.rs
@@ -13,6 +13,7 @@ use std::error::Error;
 use std::ffi::{CStr, CString};
 use std::fmt::{self, Debug, Display, Formatter};
 use std::ops::Deref;
+use std::path::Path;
 
 /// An owned LLVM String. Also known as a LLVM Message
 #[derive(Eq)]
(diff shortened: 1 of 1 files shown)


src/support/mod.rs
```diff
@@ -137,12 +138,49 @@ pub fn get_llvm_version() -> (u32, u32, u32) {
     return (major, minor, patch);
 }
 
+/// Permanently load the dynamic library with the given `filename`.
+///
+/// It is safe to call this function multiple times for the same library.
+///
+/// Returns `true` if there was an error while loading the library.
+#[deprecated(note = "Use `try_load_library_permanently` instead")]
```
TheDan64, line 146: Could you please add that it will be removed in a future version?

src/support/mod.rs
```diff
@@ -137,12 +138,49 @@ pub fn get_llvm_version() -> (u32, u32, u32) {
...
 pub fn load_library_permanently(filename: &str) -> bool {
     let filename = to_c_str(filename);
 
     unsafe { LLVMLoadLibraryPermanently(filename.as_ptr()) == 1 }
 }
 
+/// Possible errors that can occur when loading a library
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
```
TheDan64, line 154: Could you please derive thiserror::Error

src/support/mod.rs
```diff
@@ -137,12 +138,49 @@ pub fn get_llvm_version() -> (u32, u32, u32) {
...
+pub enum LoadLibraryError {
+    /// The given path could not be converted to a [`&str`]
+    UnicodeError,
+    /// The given path could not be loaded as a library
+    LoadingError,
+}
+
+/// Permanently load the dynamic library at the given `path`.
+///
+/// It is safe to call this function multiple times for the same library.
```
TheDan64, line 164: Did you get this from the docs somewhere? How are we sure it's safe?

TheDan64 (changes requested):

src/support/mod.rs
gavrilikhin-d: I'd suggest to have an update cycle `load_library_permanently` -> `try_load_library_permanently` -> `load_library_permanently`, instead of removing it.

In the end, `load_library_permanently` should have the same signature as `try_load_library_permanently`

src/support/mod.rs
gavrilikhin-d: https://llvm.org/doxygen/classllvm_1_1sys_1_1DynamicLibrary.html#a53d32d3b3baefdec31d3d94b0586d437

gavrilikhin-d: Or for C wrapper: https://llvm.org/doxygen/group__LLVMCCore.html#ga6fc1331c1a6d2cc6f0fda94f4a4636f9

src/support/mod.rs
TheDan64: We should maybe just change it then, I'm not really concerned about breaking changes since we're pre-1.0

src/support/mod.rs
```diff
@@ -137,10 +138,62 @@ pub fn get_llvm_version() -> (u32, u32, u32) {
...
+    assert_eq!(
+        load_library_permanently(Path::new("missing.dll")),
+        Err(LoadLibraryError::LoadingError)
+    );
+}
+
+/// Permanently loads all the symbols visible inside the current program
+pub fn load_visible_symbols() {
+    unsafe { LLVMLoadLibraryPermanently(std::ptr::null()) };
+}
```
gavrilikhin-d, line 177: I didn't find it anywhere in docs, but people from StackOverflow (and tests) say that passing `nullptr` as filename results in loading all the symbols visible inside program

src/support/mod.rs
```diff
@@ -137,10 +138,62 @@ pub fn get_llvm_version() -> (u32, u32, u32) {
...
+
+/// Search through all previously loaded dynamic libraries for `symbol`.
+///
+/// Returns an address of the symbol, if found
+pub fn search_for_address_of_symbol(symbol: &str) -> Option<usize> {
```
gavrilikhin-d, line 182: We may use this to protect from running non-linked functions in jit

TheDan64 (approved):
Thanks!

Issue #454, opened 2023-12-23
Adding get_allocated_type, is_terminator and is_conditional to InstructionValue


tfzee:
**Is your feature request related to a problem? Please describe.**
Missing Functions from Inkwell

**Describe the solution you'd like**
Adding safe wrappers for them

**Describe possible drawbacks to your solution**
-

**Describe alternatives you've considered**
-

**Additional context**
-

tfzee:
This would be my straightforward solution for this.
 https://github.com/TheDan64/inkwell/pull/455

PR #455, opened 2023-12-23
Adding get_allocated_type, is_terminator and is_conditional to InstructionValue

tfzee:
Adding the 3 Functions and the corresponding tests.

## Description
-

## Related Issue

https://github.com/TheDan64/inkwell/issues/454

## How This Has Been Tested

Only tested on LLVM-16 

## Checklist

### DIFF
--- a/src/values/instruction_value.rs
+++ b/src/values/instruction_value.rs
@@ -227,6 +227,16 @@ impl<'ctx> InstructionValue<'ctx> {
         unsafe { !LLVMIsATerminatorInst(self.as_value_ref()).is_null() }
     }
 
+    // SubTypes: Only apply to terminators
+    /// Returns whether a terminator is conditional or not
+    pub fn is_conditional(self) -> bool {
+        if self.is_terminator() {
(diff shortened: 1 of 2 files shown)


src/values/instruction_value.rs
```diff
@@ -282,6 +298,15 @@ impl<'ctx> InstructionValue<'ctx> {
         Ok(())
     }
 
+    // SubTypes: Only apply to alloca instruction
+    /// Returns the type that is allocated by the alloca instruction.
+    pub fn get_allocated_type(self) -> Result<AnyTypeEnum<'ctx>, &'static str> {
```
TheDan64, line 303: Wouldn't this only be a BasicTypeEnum? I would imagine AnyType has types that cannot be allocated?

tfzee: Oh yeah that's correct! I incorrectly assumed it would be the same return type as get_type.

TheDan64 (approved):
Thanks!

PR #463, opened 2024-01-21
Expose some LLVMValueRef constructors

SUPERCILEX:
I only did some of them so llvm-plugin can move off its fork. Example use:
https://github.com/jamesmth/llvm-plugin-rs/blob/7fe29ca68b70538321ba63afc0dd5fc856809488/llvm-plugin/src/analysis.rs#L116


TheDan64 (approved):
Thanks!

PR #464, opened 2024-01-21
Add set_field_at_index for llvm-plugin

SUPERCILEX:
This is the last change needed to allow llvm-plugin to move off its fork:
https://github.com/jamesmth/llvm-plugin-rs/blob/7fe29ca68b70538321ba63afc0dd5fc856809488/examples/strings.rs#L103

### DIFF
--- a/src/values/struct_value.rs
+++ b/src/values/struct_value.rs
@@ -1,4 +1,4 @@
-use llvm_sys::core::{LLVMGetNumOperands, LLVMGetOperand};
+use llvm_sys::core::{LLVMGetNumOperands, LLVMGetOperand, LLVMSetOperand};
 
 use llvm_sys::prelude::LLVMValueRef;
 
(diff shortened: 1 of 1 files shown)


src/values/struct_value.rs
```diff
@@ -52,6 +52,17 @@ impl<'ctx> StructValue<'ctx> {
         unsafe { Some(BasicValueEnum::new(LLVMGetOperand(self.as_value_ref(), index))) }
     }
 
+    /// Sets the value of a field belonging to this `StructValue`.
+    pub fn set_field_at_index<BV: BasicValue<'ctx>>(self, index: u32, val: BV) -> bool {
+        if index >= self.count_fields() {
+            return false;
+        }
+
+        unsafe { LLVMSetOperand(self.as_value_ref(), index, val.as_value_ref()) }
```
TheDan64, line 61: What happens if the field and value are of different types? Is that UB? Or just a IR type error?

SUPERCILEX: Ah, yep it segfaults. I added type checking.

TheDan64 (approved):
Thanks!

PR #469, opened 2024-02-06
Add get_name to values

### LINKED ISSUE #306: Add a get_name implementation to BasicValue / BasicValueEnum
**Is your feature request related to a problem? Please describe.**

Hi @TheDan64, I've been working through Kaleidoscope's tutorial with inkwell and realized that I couldn't simply grab a parameter's name. e.g. I wanted to do something like the below
```
let the_fn: FunctionValue = ...
...
for param in the_fn.get_param_iter() {
  // I want to use param.get_name() here!
  self.named_values.insert(param.get_name(), param.as_any_value_enum());
}
```
Because function parameter's should be float valued in Kaleidoscope, I can simply check the param type with `is_float_value()` and then cast it with `into_float_value()` if that check succeeds, then I can use `FloatValue`'s `get_name()`. i.e.,

```
let param_as_float: FloatValue = if param.is_float_value() {
    Ok(param.into_float_value())
} else {
    Err( ... )
}?;

let param_name = param_as_float
    .get_name()
    .to_str()
    ...
```

So it seems you can work around it, it just doesn't feel as ergonomic.

**Describe the solution you'd like**
I haven't thought about solutions here. I did see an [old thread ](https://github.com/TheDan64/inkwell/issues/16#issuecomment-628995954)where `get_name()` was mentioned, so thought I'd re-raise it since it's been a few years.

SUPERCILEX:
Closes #306

### DIFF
--- a/src/values/enums.rs
+++ b/src/values/enums.rs
@@ -10,6 +10,7 @@ use crate::values::{
 };
 
 use std::convert::TryFrom;
+use std::ffi::CStr;
 use std::fmt::{self, Display};
 
 use super::AnyValue;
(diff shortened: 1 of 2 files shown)


TheDan64 (approved):
Thanks!

Issue #476, opened 2024-02-20
Support for `llvm::orc2::LLJit`


JCBurnside:
**Is your feature request related to a problem? Please describe.**
The current [`ExecutionEngine`](https://thedan64.github.io/inkwell/inkwell/execution_engine/struct.ExecutionEngine.html) is using an older engine that has issues with global mappings.

**Describe the solution you'd like**
I would like a jit engine using [LLJIT](https://llvm.org/docs/ORCv2.html#lljit-and-lllazyjit).  the needed c-api functions are anything in `llvm_sys::orc2::lljit`

**Describe possible drawbacks to your solution**
might be more difficult to call functions in an idiomatic way. 

**Describe alternatives you've considered**
using the old system and using llvm_sys directly.

**Additional context**
small sample of how to crate an LLJIT instances https://gist.github.com/JCBurnside/e93d6b95d1caa3746b7fe72073094cc2
sample of using LLJIT (line 95 is the part I am having issues translating) https://github.com/llvm/llvm-project/blob/main/llvm/examples/HowToUseLLJIT/HowToUseLLJIT.cpp

JCBurnside:
turns out it's just a transmute to get the returned address into a usable function.  also updated the gist with a fix for a double free.  `LLVMOrcCreateLLJIT` takes ownership of the builder.

PR #477, opened 2024-02-21
Implementation of orc2 based jit engine.

JCBurnside:
## Description

It supports creating the LLJIT engine off a module (consuming said module), adding global mappings off name and global values.
Documentation needs a bit of work (thus starting this as a draft pr) but rather simple implementation and tried to mirror `JITExecutionEngine` though it is constructed differently.

## Related Issue
#476

## How This Has Been Tested

I have tested with a local project referencing the local fork and the samples attached to `LLJITExecutionEngine::get_function` and `LLJITExecutionEngine::add_global_mapping`

## Checklist

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -45,6 +45,7 @@ use crate::data_layout::DataLayout;
 use crate::debug_info::{DICompileUnit, DWARFEmissionKind, DWARFSourceLanguage, DebugInfoBuilder};
 use crate::execution_engine::ExecutionEngine;
 use crate::memory_buffer::MemoryBuffer;
+use crate::orc2::LLJITExecutionEngine;
 #[llvm_versions(13.0..=latest)]
 use crate::passes::PassBuilderOptions;
 use crate::support::{to_c_str, LLVMString};
@@ -169,6 +170,7 @@ pub struct Module<'ctx> {
     data_layout: RefCell<Option<DataLayout>>,
     pub(crate) module: Cell<LLVMModuleRef>,
     pub(crate) owned_by_ee: RefCell<Option<ExecutionEngine<'ctx>>>,
+    pub(crate) owned_by_lljit : RefCell<Option<LLJITExecutionEngine<'ctx>>>,
     _marker: PhantomData<&'ctx Context>,
 }
 
@@ -184,6 +186,7 @@ impl<'ctx> Module<'ctx> {
         Module {
             module: Cell::new(module),
             owned_by_ee: RefCell::new(None),
+            owned_by_lljit : RefCell::new(None),
             data_layout: RefCell::new(Some(Module::get_borrowed_data_layout(module))),
             _marker: PhantomData,
         }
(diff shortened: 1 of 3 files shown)


JCBurnside:
alright I feel I have commented everything.  Feel free to let me know anything i missed.

JCBurnside:
hmmm sees i need to find out what llvm version LLJIT was added.  And then rebase to master.  I'll get right on that!

src/orc2.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// ```
+    ///
+    /// # Safety
+    ///
+    /// It is the caller's responsibility to ensure they call the function with
+    /// the correct signature and calling convention.
+    ///
+    /// The `JitFunction` wrapper ensures a function won't accidentally outlive the
+    /// execution engine it came from, but adding functions after calling this
+    /// method *may* invalidate the function pointer.
```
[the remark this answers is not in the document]
gavrilikhin-d: Just like you can't store references to `Vec` elements and push into it

src/module.rs
```diff
@@ -184,6 +186,7 @@ impl<'ctx> Module<'ctx> {
         Module {
             module: Cell::new(module),
             owned_by_ee: RefCell::new(None),
+            owned_by_lljit : RefCell::new(None),
```
TheDan64, line 189: I wonder if we should just have one owned by field with an enum in the option?

src/orc2.rs
```diff
@@ -0,0 +1,357 @@
...
+        let name = to_c_str(name.as_ref());
+        self.add_global_mapping_impl(name, addr)
+    }
+
+    /// A grouped version of [Self::add_global_mapping]
+    pub fn add_global_mappings(&self, mappings:&[(&str,usize)]) -> Result<(),LLVMString> {
+        self.add_global_mappings_impl(mappings.iter().map(|(name,addr)| (to_c_str(name),*addr)))
+    }
+
+    fn add_global_mapping_impl(&self,name:Cow<'_,CStr>, addr:usize) -> Result<(),LLVMString> {
```
TheDan64, line 155: please format these changes

ErisianArchitect:
It seems that JCBurnside may have abandoned this pull request. Would it be alright if I try to implement the Orc/Orc2 JIT engine?

TheDan64:
@ErisianArchitect yes that would be fine

ErisianArchitect:
@TheDan64 I'm currently working on support for Orc (not yet Orc2) for LLVM 8 through 11. I had a few things I wanted to discuss.

1.) When you add a module with AddCompiledIR functions, it accepts a symbol resolver function and a pointer to a context that is passed to that symbol resolver context. I was thinking that I could create a local symbol resolver the resolves symbols per-module, and a global symbol resolver that resolves symbols per-engine, and the local resolver could fallback to the global resolver if the symbol is not found in the local resolver.

2.) The Orc JIT engine symbol resolver for the AddCompiledIR functions uses mangled symbols for function names. These mangled symbols are derived from the JIT Stack. I was thinking that the local/global symbol resolvers could internally store a HashMap where the mangled symbol is the key and the value is the function pointer.

3.) It would be nice for the OrcEngine to do some cleanup when dropped, so I was thinking it would be a good idea to add the [scopeguard](https://docs.rs/scopeguard/latest/scopeguard/) dependency for the on Drop mechanism. It would be easier than programming it manually.

I probably have some other thoughts, but that's all I've come up with so far. I'm still designing the API.

ErisianArchitect:
The Orc V1 API is nearing completion (I didn't end up needing scopeguard after all). But there's still a major issue. When you add a module to the Orc JIT stack with eager compilation  (LLVM v11) using `LLVMOrcAddEagerlyCompiledIR`, you cannot locate the symbol addresses using `LLVMOrcGetSymbolAddress`/`LLVMOrcGetSymbolAddressIn`. From my painful dive into the reason why (which was very unsuccessful, for the most part), I believe that the reason why symbols can't be found is because Orc JIT doesn't re-export symbols or retain information about them with eager compilation. This would be a limitation with the Orc V1 API (and maybe possibly is just a problem with my local LLVM static libraries). If this is an actual limitation of Orc V1, that means that eager compilation is essentially useless. I don't know if the symbols can be looked up by modules added afterward, but my guess is that they can't. My guess is that the module becomes completely useless after it's passed to the jit stack. I'm pretty sure it's not a bug in my code, because I can switch out `LLVMOrcAddEagerlyCompiledIR` with `LLVMOrcAddLazilyCompiledIR` and it works just fine.

I even tried emitting the code to an object memory buffer and then adding it to the jit stack as an object file, but the symbols still weren't resolvable.

I'm not sure whether to leave that part of the API in, or if I should exclude it. Either way, the issues should be resolved in Orc V2, which I plan on starting work on either today or tomorrow.

**Update**: I did some more digging, and it seems that this is a known limitation on Windows: https://stackoverflow.com/questions/49866755/rust-llvm-orc-jit-cannot-find-symbol-address

Issue #242, opened 2021-04-15
ExecutionEngine is not thread-safe


programmerjake:
**Describe the Bug**
The `ExecutionEngine` class in LLVM *is* thread safe and intended to be used from multiple threads. The `ExecutionEngine` wrapper in inkwell is not -- it uses `Rc<...>` internally.

https://github.com/llvm/llvm-project/blob/6ce1067f2ddd8623b163175ee9488673b9ced8d2/llvm/include/llvm/ExecutionEngine/ExecutionEngine.h#L159

**LLVM Version (please complete the following information):**
 - LLVM Version: 11.1.0
 - Inkwell Branch Used: llvm11-0

TheDan64:
We don't currently support using Inkwell types across multiple threads as it would be a pretty big undertaking to get right. Thanks for pointing it out, though. https://github.com/TheDan64/inkwell/issues/28 is related

I'm not even sure this is possible to do safely in rust as the EE needs to have a lifetime associated with the parent Context that it is derived from, and you wouldn't be able to send that `EE<'ctx>` to another thread because of that lifetime, I think?

programmerjake:
Hmm, maybe it could be done the other way around: Contexts would have the lifetime of the ExecutionEngine, since the ExecutionEngine handles multiple Contexts.

Alternative idea: the safe ExecutionEngine wrapper could hold on to `Vec<Arc<ContextInsides>>` for all the contexts it needs to keep live, and a `Context` would be more like `struct Context(Option<ContextInsides>)` where the `ExecutionEngine` would take out the insides and prevent using the context while LLVM could be using the Context from another thread internally.

BadBastion:
I have vested interest in this usecase and would be more than happy to take a crack at it. 

I believe `ExecutionEngine` doesn't directly depend on the lifetime of any specific `Context`, it is specifically the function pointers returned by execution engine that require both `Context` and `ExecutionEngine` to outlive the function pointer. 

The most direct way to represent this in rust would be changing `JitFunction` to hold an `Arc` to both `Context` &  `ExecutionEngine` objects. 

Unfortunately I don't see any way this isn't a breaking change to the API. Even if it is just returning `<'ctx>` lifetimes.

TheDan64:
Feel free to take a crack!

Its worth mentioning though, that EEs can be used multi threaded today with scoped threads!

Just wanted to throw that out there for anyone looking for a more immediate solution

BadBastion:
[...]

Issue #28, opened 2017-11-02
Support for multithreading


TheDan64:
IIRC LLVM supports multithreading. We should support this at some point, if possible. It will likely have it's own set of obstacles. Any `Rc` types will likely need to be changed to `Arc` and our existing memory model may or may not need to be further reworked. This is for the far future, so I am marking it as v0.3.0 for now.

novacrazy:
What's the current plan for this?

TheDan64:
There are currently no plans to support multithreaded LLVM. It's still something that we should explore in the future, but single-threaded LLVM is proving to be difficult and time consuming enough as it is to get right

novacrazy:
Do you have any rough idea as to how you would go about it? I'm working on a language right now and may be able to help if there is time.

TheDan64:
No clue. I think the first step is information. I have no idea what LLVM's multithreading mode does concretely. Does `Context` suddenly become `Sync`? (Ie is it being locked by a mutex or something?). Furthermore, it's not clear how we would toggle this functionality. LLVM does provide a runtime check to see if multithreaded is enabled, but we would likely need this information at compile time to mark the type as `Sync`...

JCapucho:
It seems that it is a compile time gate in llvm https://llvm.org/doxygen/group__LLVMCCoreThreading.html

haochending:
Is there a work-around at the moment to reuse the jit compiled function from a different thread?

TheDan64:
I don't believe that is thread-safe to do

PR #487, opened 2024-04-06
[DRAFT] thread safe-ish execution engine

BadBastion:
Draft for conversation, this is currently demonstration of what **isn't** a safe change. The main thing that is wrong with this is passing the `'ctx` across threads. 

The `JitFunction<'ctx, F>` type is quite a challenging interface for inkwell due to the limited ways to fulfill this lifetime with safe rust. The idiomatic rust would be to change to `JitFunction<F>` and an Arc to both **Execution Engine** & Context. 

This seems like the most elegant solution from the perspective of thread safe JIT functions, but incurs cascading breaking changes to the rest of the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ struct CodeGen<'ctx> {
     context: &'ctx Context,
     module: Module<'ctx>,
     builder: Builder<'ctx>,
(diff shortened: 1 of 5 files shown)


BadBastion:
Relevant to #242 and #28

PR #493, opened 2024-04-30
Update `llvm_versions` macro syntax to match Rust ranges

DaniPopes:
## Description

Refactor the `llvm_versions` internal macro implementation and update its syntax to more closely match normal Rust syntax.

This means removing the `latest` special case in favor of half-open ranges (`x..`) and allowing to specify <minimum LLVM version> up to x (`..x`). This is also useful to have 2 different implementations of a function that changed at version `y` by having two items with `llvm_version(..y)` and `llvm_version(y..)`, just like when slicing a string or a vector.

Since currently all supported versions are `<major>.0.0`, I've also allowed version specifiers to be just integers which represent just the major version.

This refactoring is in preparation to add a new `cfg_attr`-style macro for conditional attributes, for example `deprecated`, and to migrate remaining manual `#[cfg(any(...))]` to this macro.

The enum macro implementation is untouched, it was just moved to its own module.

## How This Has Been Tested

Compiling a few selected LLVM versions.

## Option\<Breaking Changes\>

Only to the `llvm_versions` internal macro.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ edition = "2021"
 
 [features]
 default = ["target-all"]
-
 # Please update internal_macros::FEATURE_VERSIONS when adding a new LLVM version
 llvm4-0 = ["llvm-sys-40"]
 llvm5-0 = ["llvm-sys-50"]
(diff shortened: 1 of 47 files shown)


TheDan64 (approved):
Thanks!

Issue #497, opened 2024-05-18
Inkwell cannot be built on MacOS


BigBadE:
**Describe the Bug**
values/mod.rs line 228 expects section to be a &str, but a conditional compilation block uses format! to modify section, making it a String. This was introduced by #398.

**To Reproduce**
Build on MacOS

**Expected Behavior**
Build success

**LLVM Version (please complete the following information):**
 - LLVM Version: [e.g. 6.0]
18.0 
- Inkwell Branch Used: [e.g. llvm6-0]
Master

**Desktop (please complete the following information):**
 - OS: [e.g. Ubuntu 16.04, Windows 10]
MacOS 13.3 and 13.4 (Aarch64 and x86)

**Additional Context**
I'm opening a PR with a fix after a MacOS machine runs a test on the fixed version

PR #498, opened 2024-05-18
Fix compilation on MacOS

BigBadE:
## Description

Simply fixes the conditional compilation code inside values/mod.rs to correctly use the &str type instead of String

## Related Issue
#497 

## How This Has Been Tested
Code was built on a MacOS x86_64 and MacOS Aarch64 machine (13.0 and 13.4) using github actions

On a side note, it may be wise to setup actions to automatically build the project to prevent issues like this.

There hasn't been any testing of the specific section function, but any issues after this change will have existed beforehand, all this change does it fix the compiler error.

## Checklist


TheDan64 (approved):
Thank you! Would love to see CI for macos!

PR #518, opened 2024-07-19
expose functions for working with LLVMContextRef for interoperability with other libraries

gavrilikhin-d:
## Description

This makes some functions in Context and ContextRef `pub` instead of `pub(crate)`. 
Also adds getter for raw LLVMContextRef

## How This Has Been Tested

Tested with llvm18-0 in my code that uses `melior`, `mlir-sys` and `inkwell` together.

## Checklist

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -419,11 +419,9 @@ unsafe impl Send for Context {}
 impl Context {
     /// Get raw [`LLVMContextRef`].
     ///
-    /// # Safety
-    ///
     /// This function is exposed only for interoperability with other LLVM IR libraries.
-    /// It's not intended to be used by most users, hence marked as unsafe.
-    pub unsafe fn raw(&self) -> LLVMContextRef {
+    /// It's not intended to be used by most users.
+    pub fn raw(&self) -> LLVMContextRef {
         self.context.0
     }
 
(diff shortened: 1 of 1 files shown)


gavrilikhin-d:
Obviously, tests are failing not due to this PR :/

src/context.rs
```diff
@@ -417,7 +417,24 @@ pub struct Context {
...
 
 impl Context {
-    pub(crate) unsafe fn new(context: LLVMContextRef) -> Self {
+    /// Get raw [`LLVMContextRef`].
+    ///
+    /// # Safety
+    ///
+    /// This function is exposed only for interoperability with other LLVM IR libraries.
+    /// It's not intended to be used by most users, hence marked as unsafe.
+    pub unsafe fn raw(&self) -> LLVMContextRef {
```
TheDan64, line 426: This doesn't need to be unsafe

src/context.rs
```diff
@@ -1321,7 +1338,23 @@ pub struct ContextRef<'ctx> {
...
 
 impl<'ctx> ContextRef<'ctx> {
-    pub(crate) unsafe fn new(context: LLVMContextRef) -> Self {
+    /// Get raw [`LLVMContextRef`].
+    ///
+    /// # Safety
+    ///
+    /// This function is exposed only for interoperability with other LLVM IR libraries.
+    /// It's not intended to be used by most users, hence marked as unsafe.
+    pub unsafe fn raw(&self) -> LLVMContextRef {
```
TheDan64, line 1347: Same here; calling the raw method alone can't produce undefined behavior since it just returns a pointer. Unsafety is in how it might be used

gavrilikhin-d:
Yes, I understand that those are safe functions. I intentionally marked
them as unsafe, because they shouldn't be used by sole inkwell users. And
those getters will be used in unsafe context either way, so it doesn't
actually matter

TheDan64:
We should not mark the raw methods as unsafe since there's no unsafety that can result from using them alone. See the std lib for example which has safe methods which return raw ptrs: https://doc.rust-lang.org/std/vec/struct.Vec.html#method.into_raw_parts

gavrilikhin-d:
Fixed

TheDan64 (approved):
Thanks!

Issue #541, opened 2024-10-09
Scalable vector type


my4ng:
Scalable [vector](https://llvm.org/docs/LangRef.html#t-vector) was added to LLVM 11, to support vector extensions with compile-time-unknown vector length such as arm's SVE and RISC-V's [RVV](https://llvm.org/docs/RISCV/RISCVVectorExtension.html). llvm_sys has the `LLVMScalableVectorType`, but inkwell currently has no support for it.

~~I am currently working on a solution that reuses the `VectorType` to support both fixed and scalable kinds, in a non-breaking manner.~~

EDIT: A separate scalable vector type and value are required to prevent mixing them with the fixed kind.

PR #542, opened 2024-10-09
Add scalable vector

my4ng:
## Description

This adds scalable vector support through the new `ScalableVectorType` and `ScalableVectorValue`.

## Related Issue

#541 

## How This Has Been Tested

Tests mirroring the fixed vector kind have been added for type, value and builder. 

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -51,7 +51,7 @@ use crate::values::CallableValue;
 use crate::values::{
     AggregateValue, AggregateValueEnum, AsValueRef, BasicMetadataValueEnum, BasicValue, BasicValueEnum, CallSiteValue,
     FloatMathValue, FunctionValue, GlobalValue, InstructionOpcode, InstructionValue, IntMathValue, IntValue, PhiValue,
-    PointerMathValue, PointerValue, StructValue, VectorValue,
+    PointerMathValue, PointerValue, StructValue, VectorBaseValue,
 };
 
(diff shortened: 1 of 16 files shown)


my4ng:
It appears that a separate `ScalableVectorType` is required to ensure type safety, rather than reusing `VectorType`, to ensure fixed and scalable values do not mix in operations.

my4ng:
It has been tested against the workflow. 

The `ScalableVectorType/Value` are available across all versions; however, the `scalable_vec_type` functions are gated by `#[llvm_versions(12..)]`, therefore it is not possible to construct a type or value in version less than 12. The doctests in them are ignored since it fails without the aforementioned functions. The alternative approach of hiding the type and value in versions less than 12 is a *lot* more involved and thus not taken.

TheDan64 (approved):
Thank you!

my4ng:
@TheDan64 sorry I have been away and forgot to fix a replace in the fixed vector type doc comments: https://github.com/TheDan64/inkwell/blob/8bd9f087c7fedd3897016568184aa8516d6a0abf/src/types/float_type.rs#L67-L80

Do you mind fixing it in master? Thanks!

TheDan64:
@my4ng fix what exactly?

my4ng:
@TheDan64 It should be vector type rather than scalar vector type for the doc comment linked as it is the normal vec_type function.

Issue #489, opened 2024-04-15
Add set_opaque_pointers to Context


Derppening:
**Is your feature request related to a problem? Please describe.**
#468 introduced opaque pointers, and in the PR opaque pointers are enabled by default with no opt-out mechanism.

**Describe the solution you'd like**
Add `Context::set_opaque_pointers` by using `LLVMContextSetOpaquePointers` in LLVM's C bindings, which is present in [LLVM 15](https://github.com/llvm/llvm-project/blob/4ba6a9c9f65bbc8bd06e3652cb20fd4dfc846137/llvm/include/llvm-c/Core.h#L562), [LLVM 16](https://github.com/llvm/llvm-project/blob/llvmorg-16.0.0/llvm/include/llvm-c/Core.h#L574).

**Describe possible drawbacks to your solution**
Since opaque pointers are enabled by default since LLVM 15, there should be no drawbacks as the default already utilizes opaque pointers; Users will need to explicitly opt-in to the legacy behavior by invoking the function.

**Describe alternatives you've considered**
None. Forcing LLVM 15/16 users to use opaque pointers may be an alternative as it is recommended for users to migrate over anyways, but there may be users that use LLVM 15 where the status of opaque pointers is ([source](https://llvm.org/docs/OpaquePointers.html#version-support)):

> **LLVM 15**: Opaque pointers are enabled by default. Typed pointers are still supported.

PR #570, opened 2025-02-17
Implement feature for disabling opaque pointers on LLVM 15 and 16

Derppening:
## Description

This pull request implements support for using typed pointers in LLVM 15. This is implemented by several steps:

- Add feature `typed-pointers` to distinguish APIs for typed vs opaque pointers
- Implicitly set the use of typed pointers during `Context` creation
- Add `typed-pointers` feature dependency on all LLVM versions that only have typed pointer support (i.e. LLVM 4-14)
- Updated `#[cfg(...)]` blocks to use `feature = "typed-pointers"` instead of version guards for pointer-related APIs
- Add version guards in `lib.rs` to ensure that the `typed-pointers` feature can only be enabled in LLVM 15-16

Note that previous defaults are still present - LLVM 15 and 16 still uses opaque pointers by default, and will only switch to typed pointers if the `typed-pointers` feature is enabled.

Some points of note or for discussion with regards to the current implementation:

- The current implementation uses a Cargo feature to guard typed vs opaque pointer API usage. Are there any known use-cases where a user of this library requires both?
  - If so, how should we handle APIs that have previously been version-guarded? E.g. `Builder::build_load`, where the typed pointer version of this API calls `LLVMBuildLoad` while the opaque pointer version calls `LLVMBuildLoad2`.
- LLVM technically supports both typed and opaque pointers from LLVM ~~13-16~~ 15-16, although only LLVM 15 officially supports both. Should I also expand the scope of this PR to include ~~opt-in into opaque pointers in LLVM 13-14, as well as~~ opt-out out of opaque pointers in LLVM 16?
  - If so, how should this look like?
  - **Edit:** I just realized that `LLVMContextSetOpaquePointers` is only available in LLVM 15-16, so please disregard my message regarding LLVM 13-14.

## Related Issue

#489

## How This Has Been Tested

This PR is tested using:

```
$ cargo test --workspace --features llvm15-0-prefer-dynamic,no-libffi-linking
$ cargo test --workspace --features llvm15-0-prefer-dynamic,no-libffi-linking,typed-pointers
$ cargo test --workspace --features llvm16-0-prefer-dynamic,no-libffi-linking
$ cargo test --workspace --features llvm16-0-prefer-dynamic,no-libffi-linking,typed-pointers
```

AFAICT all tests (including doctests) pass with LLVM 15 and 16 under both opaque pointer and typed pointer modes.

## Checklist


Derppening:
v2: Added the ability to disable opaque pointers in LLVM 16 as well. PR description is also updated to reflect this fact.

TheDan64 (approved):
Thanks for the detailed work!

TheDan64:
@qartik I was hoping LLVM 19 support would be unblocked before the next release, but if you need a new version published badly, I'm happy to do a release sooner than that

qartik:
@TheDan64 that will be very nice. We've had trouble trying to upgrade to LLVM 15/16 which still support typed pointers as inkwell had dropped support. This PR seems to allow use of typed pointers in those versions. It will help us in the migration to opaque pointers over time. Thanks!

TheDan64:
Yup, planning to get to it this weekend most likely

TheDan64:
Just blocked on last issue, #587

TheDan64:
@qartik inkwell 0.6.0 is now published to crates.io

PR #572, opened 2025-02-21
CallSiteValue::get_called_fn_value: return None on indirect calls

### LINKED ISSUE #571: `CallSiteValue::get_called_fn_value()` fails it assert when the callee is a `PointerValue`
**Describe the Bug**

```sh
thread 'main' panicked at /home/ju/.cargo/registry/src/index.crates.io-6f17d22bba15001f/inkwell-0.5.0/src/values/fn_value.rs:48:9:
assertion failed: !LLVMIsAFunction(value).is_null()
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

`CallSiteValue::get_called_fn_value()` fails it assert when the callee is a `PointerValue`.

Is there a way to tell if the callee is a `PointerValue` or a `FunctionValue` ?

**To Reproduce**

```rust
const LLVM_IR: &str = r#"
; ModuleID = 'main.c'
source_filename = "main.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128"
target triple = "x86_64-pc-linux-gnu"

; Function Attrs: noinline nounwind optnone uwtable
define dso_local void @dummy_fn() #0 {
  ret void
}

; Function Attrs: noinline nounwind optnone uwtable
define dso_local i32 @main() #0 {
  %1 = alloca i32, align 4
  %2 = alloca ptr, align 8
  store i32 0, ptr %1, align 4
  call void @dummy_fn()
  ; The following 3 lines cause a panic
  store ptr @dummy_fn, ptr %2, align 8
  %3 = load ptr, ptr %2, align 8
  call void %3()
  ret i32 0
}

attributes #0 = { noinline nounwind optnone uwtable "frame-pointer"="all" "min-legal-vector-width"="0" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cmov,+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" }

!llvm.module.flags = !{!0, !1, !2, !3, !4}
!llvm.ident = !{!5}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{i32 8, !"PIC Level", i32 2}
!2 = !{i32 7, !"PIE Level", i32 2}
!3 = !{i32 7, !"uwtable", i32 2}
!4 = !{i32 7, !"frame-pointer", i32 2}
!5 = !{!"Ubuntu clang version 18.1.8 (++20240731024944+3b5b5c1ec4a3-1~exp1~20240731145000.144)"}
"#;

fn main() {
    let memory_buffer = inkwell::memory_buffer::MemoryBuffer::create_from_memory_range_copy(
        LLVM_IR.as_bytes(),
        "name",
    );

    let binding = inkwell::context::Context::create();
    let module = binding.create_module_from_ir(memory_buffer).unwrap();

    for f in module.get_functions() {
        for bb in f.get_basic_blocks() {
            for i in bb.get_instructions() {
                if let Ok(call_site_value) = inkwell::values::CallSiteValue::try_from(i) {
                    eprintln!(
                        "Found a call site: {:?}",
                        call_site_value.get_called_fn_value()
                    );
                }
            }
        }
    }
}

```

**Expected Behavior**
A primitive should be provided to distinguish calls to `PointerValue` from calls to `FunctionValue`.

 - LLVM Version: 18.0
 - Inkwell Branch Used: llvm18-0

 - OS: ubuntu 22.04 (popOS)

**Additional Context**

airwoodix:
## Description

Fixes #571.

- `CallSiteValue::get_called_fn_value` now matches the behavior of [`CallBase::getCalledFunction`](https://llvm.org/doxygen/classllvm_1_1CallBase.html#a2b1ae7cf1bafdd43d1fee4c6ad0a2913) and returns `None` when the underlying instruction is an indirect call or a type inconsistency is detected (the latter only for LLVM >= 8).
- The new `CallSiteValue::get_called_fn_type` corresponds to [`CallBase::getFunctionType`](https://llvm.org/doxygen/classllvm_1_1CallBase.html#ac3c35bd078a268a207f607d0f57dadba) and returns the type of the function called by the underlying call/invoke instruction. This is always well defined, whether the call is indirect or not. This new function only exists for LLVM >=8 because it depends on `LLVMGetCalledFunctionType`.

## Related Issue

#571 

## How This Has Been Tested

Non-regression on #571 is tested by a dedicated test which uses a simplified version of the issue's reproduction code. For simplicity, this test is restricted to LLVM >= 15 (the test's IR snippet uses opaque pointers).

Doctests of `CallSiteValue::get_called_fn_value` and the new `CallSiteValue::get_called_fn_type` cover the usual-case, happy paths.

## Breaking Changes

`CallSiteValue::get_called_fn_value` now returns an `Option<FunctionValue>` instead of a `FunctionValue` to model the fact that it may not be possible to retrieve the underlying function value (when the call is indirect).

## Discussion

`FunctionValue::new` differs from most other type constructors (e.g. `FunctionType::new`) since it checks for the safety requirements and returns `None` if they are not met:

- is it thus obsolete to have it marked `unsafe`?
- is it the intention to have most type constructors behave this way, instead of being `unsafe`?

## Checklist

### DIFF
--- a/src/values/call_site_value.rs
+++ b/src/values/call_site_value.rs
@@ -1,8 +1,10 @@
 use std::fmt::{self, Display};
 
 use either::Either;
+#[llvm_versions(8..)]
+use llvm_sys::core::LLVMGetCalledFunctionType;
 use llvm_sys::core::{
-    LLVMGetInstructionCallConv, LLVMGetTypeKind, LLVMIsTailCall, LLVMSetInstrParamAlignment,
+    LLVMGetCalledValue, LLVMGetInstructionCallConv, LLVMGetTypeKind, LLVMIsTailCall, LLVMSetInstrParamAlignment,
     LLVMSetInstructionCallConv, LLVMSetTailCall, LLVMTypeOf,
 };
 #[llvm_versions(18..)]
@@ -11,6 +13,7 @@ use llvm_sys::prelude::LLVMValueRef;
 use llvm_sys::LLVMTypeKind;
 
 use crate::attributes::{Attribute, AttributeLoc};
+use crate::types::FunctionType;
 #[llvm_versions(18..)]
 use crate::values::operand_bundle::OperandBundleIter;
 use crate::values::{AsValueRef, BasicValueEnum, FunctionValue, InstructionValue, Value};
@@ -201,9 +204,12 @@ impl<'ctx> CallSiteValue<'ctx> {
 
     /// Gets the `FunctionValue` this `CallSiteValue` is based on.
     ///
+    /// Returns [`None`] if the call this value bases on is indirect or the retrieved function
+    /// value doesn't have the same type as the underlying call instruction.
+    ///
     /// # Example
     ///
-    /// ```no_run
+    /// ```
     /// use inkwell::context::Context;
     ///
     /// let context = Context::create();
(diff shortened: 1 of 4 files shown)


src/values/call_site_value.rs
```diff
@@ -220,12 +226,62 @@ impl<'ctx> CallSiteValue<'ctx> {
...
+    #[llvm_versions(..8)]
+    #[inline]
+    fn get_called_fn_value_check_type_consistency(&self, fn_value: Option<FunctionValue<'ctx>>) -> Option<FunctionValue<'ctx>> {
+	fn_value
+    }
+
+    #[llvm_versions(8..)]
+    #[inline]
+    fn get_called_fn_value_check_type_consistency(&self, fn_value: Option<FunctionValue<'ctx>>) -> Option<FunctionValue<'ctx>> {
+	fn_value.filter(|fn_value| fn_value.get_type() == self.get_called_fn_type())
```
TheDan64, line 252: Please use cargo fmt on your changes... it appears to be mixing in tabs and looks off

airwoodix: Thanks for the feedback! I fixed formatting issues related to the changes in this PR. Other changes applied by `rustfmt` are in #575.

TheDan64 (approved):
Thanks!

TheDan64:
> is it thus obsolete to have it marked unsafe?
No, I don't think so - a user could still provide an invalid LLVMValue to us I think. So it's still up to the user to validate that value is correct

> is it the intention to have most type constructors behave this way, instead of being unsafe?
See above

PR #573, opened 2025-02-22
FunctionType::get_param_types: support metadata type

### LINKED ISSUE #546: Inkwell causing a crash when `metadata` is encountered as the type of a function argument
**Describe the Bug**
As part of a project that ingests LLVM IR I am probing function definitions (and calls) for the types of their arguments. As my ingested IR comes from Rust (for now), it is not uncommon for me to see calls to `@llvm.dbg.declare` and the accompanying declaration.

```ll
declare void @llvm.dbg.declare(metadata, metadata, metadata)
```

Given the `func: FunctionValue` corresponding to this declaration, calling `func.get_type().get_param_types()` will cause a crash in `BasicTypeEnum::new` with the message "Unsupported basic type: Metadata". 

**To Reproduce**

In an environment with the below things available, run `test2::run`.

`lib.rs`:

```rust

#[cfg(test)]
mod test2 {
    use std::path::Path;

    use inkwell::{context::Context, memory_buffer::MemoryBuffer};

    #[test]
    fn run() {
        let test_input = r"add_min.ll";
        let source_buf = MemoryBuffer::create_from_file(Path::new(test_input)).unwrap();

        let mut ctx = Context::create();
        let module = ctx.create_module_from_ir(source_buf).unwrap();

        for function in module.get_functions() {
            let fun_ty = function.get_type();

            // This crashes
            fun_ty.get_param_types();
        }
    }
}
```

`add_min.ll`:

```ll
declare void @llvm.dbg.declare(metadata, metadata, metadata) #1
```

**Expected Behavior**

I would have initially expected for `metadata` to be a supported type, but the fact that it isn't indicates that there is likely quite some complexity to making it so. For my use case, it would be sufficient to _know_ that the type was metadata and avoid the crash.

Ideally I would like to avoid just banning known-problematic functions from my analysis, but that is a stopgap I am considering for the moment.

 - LLVM Version: 18.0
 - Inkwell Branch Used: [0.5.0](https://github.com/TheDan64/inkwell/releases/tag/0.5.0) with the `llvm18-0` feature

 - OS: macOS Sequoia 15.0.1

**Additional Context**

If there is a better way to detect usages of `metadata` in type position that doesn't eventually result in a crashing call to `BasicTypeEnum::new`, I would be happy to hear about it.

airwoodix:
## Description

Fixes #546 as suggested in the issue.

`FunctionType::get_param_types()` now returns `Vec<BasicMetadataTypeEnum>` instead of `Vec<BasicTypeEnum>`.

The conversion is provided by implementing `BasicMetadataTypeEnum::new`, which delegates to `BasicTypeEnum::new` for non-metadata types. This only works for LLVM >= 6, because of the same gate on `MetadataType::new`.

## Related Issue

#546

## How This Has Been Tested

- dedicated non-regression test for the original issue.
- enabled running the doctest on `FunctionType::get_param_types()`.

## Breaking Changes

Signature change for `FunctionType::get_param_types()`.

## Checklist

### DIFF
--- a/src/types/enums.rs
+++ b/src/types/enums.rs
@@ -111,6 +111,28 @@ enum_type_set! {
 }
 
 impl<'ctx> BasicMetadataTypeEnum<'ctx> {
+    /// Create [`BasicMetadataTypeEnum`] from [`LLVMTypeRef`].
+    ///
+    /// # Safety
+    ///
+    /// Undefined behavior if the referenced type cannot be represented as [`BasicMetadataTypeEnum`],
+    /// or the underlying pointer is null.
+    ///
+    /// Before LLVM 6, [`BasicMetadataTypeEnum::MetadataType`] variants cannot be created
+    /// with this function. Attempting to do results in undefined behavior.
+    #[llvm_versions(6..)]
+    pub unsafe fn new(type_: LLVMTypeRef) -> Self {
+        match LLVMGetTypeKind(type_) {
+            LLVMTypeKind::LLVMMetadataTypeKind => Self::MetadataType(MetadataType::new(type_)),
+            _ => BasicTypeEnum::new(type_).into(),
+        }
+    }
+
+    #[llvm_versions(..6)]
+    pub unsafe fn new(type_: LLVMTypeRef) -> Self {
+        BasicTypeEnum::new(type_).into()
+    }
+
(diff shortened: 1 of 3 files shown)


TheDan64 (approved):
Thanks!

PR #593, opened 2025-08-14
Removes uneccesary less-than 64-bit check on 32-bit value.

ErisianArchitect:
## Description

The is_alignment_ok function in builder.rs uses log2 to determine the number of bits that are number needs. It compares the result with `64`. This is unnecessary as the `align` parameter is only 32-bits, so this comparison will always return `true`.

If `align` were 64-bit, the check would still be unnecessary because you can't have 2^64 as a 64-bit value. If `align` were 128-bit, then you could perform this check using `align.trailing_zeros()`, which is faster than converting to f64 and using log2, which would be inaccurate for an 128-bit integer anyway.

## How This Has Been Tested

No tests needed. It's logically sound.

## Checklist

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -3696,7 +3696,7 @@ fn is_alignment_ok(align: u32) -> bool {
     // This replicates the assertions LLVM runs.
     //
     // See https://github.com/TheDan64/inkwell/issues/168
-    align > 0 && align.is_power_of_two() && (align as f64).log2() < 64.0
+    align > 0 && align.is_power_of_two()
 }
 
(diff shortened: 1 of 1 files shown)


ErisianArchitect:
I'm also working on some changes that will allow you to set the alignment during operations where the alignment could be specified. So for example, `build_aligned_alloca` where you can specify the alignment.
I also added an AlignmentError enum that can be used in place of the `&'static str` in BuilderError::AlignmentError.
I just didn't want to make these changes all at once.
If you're against any of these changes, let me know.

PR #594, opened 2025-08-14
Adds AlignmentError enum to simplify BuilderError::AlignmentError.

ErisianArchitect:
## Description

It's harder to determine the type of alignment error from a static string, especially when different strings are used in different AlignmentError sources. I (hopefully) improved this API by adding the `AlignmentError` enum, which is used in `BuilderError::AlignmentError` in place of the static string.

I discovered in the code that there were only two error types: non power of two, and unsized type, so those are the two variants that I added.

## How This Has Been Tested

Sorry, I'm not entirely sure how to test this. It doesn't seem like a change that needs tests.

## Checklist


ErisianArchitect:
Ahh, hang on. It seems that InstructionValue::set_alignment also returns `&'static str`, which could be changed to AlignmentError, but I realized that this would break backwards compatibility.

PR #596, opened 2025-08-15
Removed unnecessary checks from is_alignment_ok

ErisianArchitect:
## Description

`u32::is_power_of_two` will return `false` for `0`, so I removed the `0` check from `is_alignment_ok`.

## How This Has Been Tested

I was unable to get the tests module build on my machine because it seems that there are some OS specific errors that no one noticed before. I plan on looking into that next.

But I did test it, nonetheless, and the test can be reproduced with just three lines (well, 1-3 lines, depending on what you want to test):
```rust
assert!(!(0_u32.is_power_of_two()));
assert!(!(3_u32.is_power_of_two()));
assert!(32_u32.is_power_of_two());
```

## Checklist

- [ x] I have read the [Contributing Guide](https://github.com/TheDan64/inkwell/blob/master/.github/CONTRIBUTING.md)

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -3696,7 +3696,8 @@ fn is_alignment_ok(align: u32) -> bool {
     // This replicates the assertions LLVM runs.
     //
     // See https://github.com/TheDan64/inkwell/issues/168
-    align > 0 && align.is_power_of_two()
+    // is_power_of_two returns false for 0.
+    align.is_power_of_two()
 }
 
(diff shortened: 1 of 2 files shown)


ErisianArchitect:
I figured I would also mention, this is the source code for `is_power_of_two`:

```rust
pub const fn is_power_of_two(self) -> bool {
    self.count_ones() == 1
}
```

TheDan64 (approved):
Makes sense, thanks!

Issue #597, opened 2025-08-16
tests fail to build on Windows due to usage of libc POSIX-only functionality.


ErisianArchitect:
**Describe the Bug**
It's not so much of as bug as it is code that won't compile.

In `test_target.rs`, there is an `os` definition [here](https://github.com/TheDan64/inkwell/blob/279ef78b3507a431e48767748c2335aef26b7d9f/tests/all/test_targets.rs#L196). There is no option for `windows`, and so it will become a compiler error and fail to build because the compiler doesn't know the type. This can easily be fixed, and I would have made a pull request, but I wasn't entirely sure the best way to fix it since I'm not entirely sure what the expected value of default_triple is. This also wouldn't work on any OS besides Linux/Mac.

Similarly, `test_execution_engine.rs` has three functions that don't work on Windows.

`MockMemoryManager::new` [here](https://github.com/TheDan64/inkwell/blob/279ef78b3507a431e48767748c2335aef26b7d9f/tests/all/test_execution_engine.rs#L375):
The problem here is that it is using POSIX-only functionality in the libc library. I wasn't entirely sure what the alternative was, so I decided it would be better as a bug report.

`MockMemoryManager::finalize_memory` [here](https://github.com/TheDan64/inkwell/blob/279ef78b3507a431e48767748c2335aef26b7d9f/tests/all/test_execution_engine.rs#L458):
Same problem as `MockMemoryManager::new`, it's using POSIX-only functionality.

And lastly, `MockMemoryManager::destroy` [here](https://github.com/TheDan64/inkwell/blob/279ef78b3507a431e48767748c2335aef26b7d9f/tests/all/test_execution_engine.rs#L479):
Same problem with libc POSIX-only functionality.

**Desktop (please complete the following information):**
 - OS: Windows 11

ErisianArchitect:
Sorry about the missing label, I'm not sure how to add it.

Issue #595, opened 2025-08-15
Specific error types for errors instead of &'static str


ErisianArchitect:
**Is your feature request related to a problem? Please describe.**
Without specific error types, it can be difficult if not impossible to determine what the cause of the error is without reading the string, which makes debugging a little harder to do.

**Describe the solution you'd like**
These `&'static str` errors could be replaced with error types specific to those errors.

**Describe possible drawbacks to your solution**
This would not be a backwards compatible change.

---

I don't mind attempting to contribute this to inkwell, but I don't want to change anything if someone thinks it's a bad idea.

TheDan64:
Yes, I agree. This is mostly tech debt from the days before `thiserror` and rust having any sort of good error type guidlines

ErisianArchitect:
I can make a rather opinionated attempt at making this change. I'm not sure what you would prefer in terms of errors, but I can give it a shot in my next pull request.

ErisianArchitect:
https://github.com/TheDan64/inkwell/pull/598

I've submitted a pull request.

PR #598, opened 2025-08-16
More specific errors.

ErisianArchitect:
## Description

Many of the result types returned `&'static str`, or otherwise a wrapper around `&'static str`. I tried to find as many instances of this as I could to make the errors more specific. I may have missed some things.

## Related Issue

https://github.com/TheDan64/inkwell/issues/595

## How This Has Been Tested

Unfortunately, I'm unable to test it because the tests environment won't compile on my machine due to POSIX-only functionality.

https://github.com/TheDan64/inkwell/issues/597

## Option\<Breaking Changes\>

I felt that it was a good idea for errors to be more specific so that it would be easier to determine the source of the error.

## Checklist


ErisianArchitect:
Sorry, I missed a typo.

ErisianArchitect:
It seems that there are some tests failing due to my change to `is_alignment_ok`, removing the ability for `0` alignment. As far as I'm aware, `0` alignment should be invalid, so I think perhaps these tests should be removed.

ErisianArchitect:
I looked into it some more, and I found [this issue](https://github.com/llvm/llvm-project/issues/53371) with this comment by one of the main contributors:
> The "A value of 0 or" part of the docs is indeed incorrect. Zero is not a legal alignment.

Since 0 is not a legal alignment, I'm going to make the assumption that it's okay to remove these tests that check for an alignment of 0.

ErisianArchitect:
I decided to do a test to see what would happen if you tried to set the alignment to 0 just to make sure that it really wasn't a legal value, and this was the result when I tried to build the module:
```
thread 'main' panicked at src\bin\defer_test\main.rs:567:25:
Module verification failed.: "huge alignment values are unsupported\n  %defer_stack = alloca inalloca swifterror { [5 x ptr], i64 }, align 9223372036854775808\n"
```

So it seems that setting the alignment to 0 can cause undesirable outcomes, so I think it's a good thing that I removed the ability for alignments to be 0.

**Edit**: This is on inkwell 0.6.0 with LLVM 18.1.

src/builder.rs
```diff
@@ -77,23 +78,35 @@ enum PositionState {
...
     #[error("Builder position is not set")]
     UnsetPosition,
     #[error("Alignment error")]
-    AlignmentError(&'static str),
+    AlignmentError(#[from] crate::error::AlignmentError),
     #[error("Aggregate extract index out of range")]
     ExtractOutOfRange,
-    #[error("Bitwidth of a value is incorrect")]
-    BitwidthError(&'static str),
+    #[error("The bitwidth of value must be a power of 2 and greater than 8.")]
```
TheDan64, line 100: Isn't this >= 8?

TheDan64 (approved):
Looks good; just have one question to clarify before merging

ErisianArchitect:
Any update on this?

Issue #609, opened 2025-10-18
Replace usage of `Either` with custom made enums.


ErisianArchitect:
**Is your feature request related to a problem? Please describe.**
`Either` can make code harder to read, so I propose that most, if not all, usages of `Either` in inkwell are replaced with equivalent enums.

**Describe possible drawbacks to your solution**
We lose the benefit of being able to use `Either` later on down the road, and it would force us to create new enums.

**Additional context**
I've already begun working on this PR, I just wanted to make an issue before hand.

ErisianArchitect:
Oh, and I'd just like to ask in the case that you're alright with these changes: How thoroughly do you want me to replace Either? Because I can make basic enums to replace it, or I can make the enums more equivalent to Either, but that would require a lot more work. I was thinking of making a derive macro for these `Either` replacement enums so that they can be created more easily. I'd like to hear your thoughts on the matter.

PR #607, opened 2025-10-02
Add debug information utility functions and support Metadata on Operands

jaynus:
## Description

Adds a few utility functions around debug information and changes instruction operand handling to support debug metadata, with some fixes to error messaging around value types too.

### Changes
- Changes `InstructionValue::get_operand` family of functions (including `get_operands`)  to return a new `OperandValue`. They currently had `Either<BasicValue, BasicBlock>` which is insufficient for operands - they can be metadata values as well.
- Added Helper functions for the transitional DbgRecord changes.  
`Module::is_new_debug_format` - this corresponds to `LLVMIsNewDbgInfoFormat`
`Module::set_new_debug_format` - this corresponds to `LLVMSetIsNewDbgInfoFormat`
- Added `as_metadata_value` to the suite of `DI*` types, as all these types are metadata values that can be fetched via instruction operands (old) or `DbgRecord` (new)
- Better output to the panics in the enum variants, so you can see what kind was unimplemente
- Updated tests to all pass for the new `get_operand` call.

## How This Has Been Tested

This was tested with llvm 20.1 locally in a personal project. Ex usage such as:
```
    module.set_new_debug_format(false);
    module.get_functions().for_each(|function| {
        function.get_basic_blocks().iter().for_each(|block| {
            block.get_instructions().skip(1).for_each(|instr| {
                    (0..instr.get_num_operands()).for_each(|i| {
                            if let Some(OperandValue::BasicMetadataValue(metadata)) =
                                instr.get_operand(i)
                            {
                                ... Handle debug metadata operands here....
                            }
                        });
                    });
                }
            });
        });
```

## Breaking Changes

Changes `InstructionValue::get_operand` and `OperandIter` to use `OperandValue` instead of `Either<BasicValue, BasicBlock>`. This was required because operands on a value instruction can be more than just BasicValue or BasicBlock, they can also be `MetadataValue` in the case of debug intrinsics. 

- All `left()` calls must now be `into_basic_value()` or `try_into()` or `into()`
- All `right()` calls must now be `into_basic_block()` or `try_into()` or `into()`

## Checklist

- [x] I have read the [Contributing Guide](https://github.com/TheDan64/inkwell/blob/master/.github/CONTRIBUTING.md)

PR #610, opened 2025-10-18
Removes Either dependency.

ErisianArchitect:
## Description

`Either` tends to make code a little harder to read, and a little harder to write, so I decided it would be a good idea to remove the dependency entirely and instead rely on tailored enums to do the job.

## Related Issue

https://github.com/TheDan64/inkwell/issues/609

## How This Has Been Tested

All tests passed for all supported versions of LLVM on Windows.

## Option\<Breaking Changes\>

Breaking changes for readability.

## Checklist

(I ran cargo clippy this time, hopefully github actions doesn't complain anyway)

### DIFF
--- a/src/values/instruction_value.rs
+++ b/src/values/instruction_value.rs
@@ -1,22 +1,25 @@
 #[llvm_versions(14..)]
 use llvm_sys::core::LLVMGetGEPSourceElementType;
 use llvm_sys::core::{
-    LLVMGetAlignment, LLVMGetAllocatedType, LLVMGetFCmpPredicate, LLVMGetICmpPredicate, LLVMGetInstructionOpcode,
-    LLVMGetInstructionParent, LLVMGetMetadata, LLVMGetNextInstruction, LLVMGetNumOperands, LLVMGetOperand,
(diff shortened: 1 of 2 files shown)


ErisianArchitect:
Oh, shoot, I just realized that I forgot to write documentation for the new enums. Let me do that real quick before merge.

ErisianArchitect:
Oh, and `CallableValue` had `Either` inside it, so I figured it would be a good idea to replace the `Either` with an enum rather than replacing `CallableValue` with an enum. If you think it would be better to replace `CallableValue` itself with an enum, let me know. I figured there was a reason that it hid away the `Either` before.

ErisianArchitect:
I should probably mention: This PR conflicts with #607. I didn't realize that until after I made the PR. For some reason I thought that it had already been merged, so I didn't think there would be a conflict.

ErisianArchitect:
Any update on this? I'd like to start work on the next PR.

TheDan64:
I won't be able to look at this until the end of the week

ErisianArchitect:
Alright, I've got a side project I'm working on anyway, so no rush.

TheDan64:
Thanks!

Issue #633, opened 2025-11-24
internal error: entered unreachable code: The given type is not a basic type. (when looking at the operands of a call)


Yoric:
**Describe the Bug**
I'm trying to use inkwell to inspect some LLVM IR.

**To Reproduce**

```rust
use std::path::Path;

use inkwell::{context::Context, memory_buffer::MemoryBuffer, values::InstructionOpcode};

fn main() {
     // Read the IR file into a memory buffer
    let memory_buffer = MemoryBuffer::create_from_file(Path::new("/tmp/test.ll")).unwrap();

    let context = Context::create();
    // Parse the IR into a module
    let module = context
        .create_module_from_ir(memory_buffer).unwrap();

    let main = module.get_function("main").unwrap();
    for bb in main.get_basic_blocks() {
        for instruction in bb.get_instructions() {
            eprintln!("Instruction: {:?}", instruction);
            if instruction.get_opcode() == InstructionOpcode::Call {
                eprintln!("Looking at {} operands", instruction.get_num_operands());
                for (i, operand) in instruction.get_operands().enumerate() {
                    eprintln!("Operand {i}: {:?}", operand);
                }
            }
        }
    }
}
```

I have applied this to the LLVM IR for the following file:

```c
#include <fcntl.h>
#include <unistd.h>

int main() {
    int fd = open("/tmp/test.txt", O_RDONLY);
    if (fd >= 0) {
        char buffer[256];
        read(fd, buffer, sizeof(buffer));
        close(fd);
    }
    return 0;
}
```

([corresponding .ll](https://github.com/user-attachments/files/23728084/test.ll.gz))

In the first `call`, this panics with

```
internal error: entered unreachable code: The given type is not a basic type.
```

**Expected Behavior**
I expected to be able to look at all the operands of the `call`.

**LLVM Version (please complete the following information):**
 - LLVM Version: [e.g. 6.0] llvm18-1
 - Inkwell Branch Used: 0.7.1

**Desktop (please complete the following information):**
 - OS: [e.g. Ubuntu 16.04, Windows 10] macOS 12.7.6.

Yoric:
Looking at it more in depth, the error shows up at the following IR:

```asm
  call void @llvm.dbg.declare(metadata i32* %2, metadata !16, metadata !DIExpression()), !dbg !17
```

the panic is raised because the arguments are `metadata`, which cannot be converted into `BasicValueEnum`. If I skip `llvm.dbg.*`, other functions work nicely.

[...]

PR #635, opened 2025-11-26
[Feature] We should be able to walk through the operands of a call wi…

Yoric:
…th metadata arguments

## Description

- We expand `Operand` to accept a new variant `Metadata`. This is probably a breaking change, so there may be a better way to do this.
- We expand `Operand::get_operand_unchecked` to create these variants when the operand is a metadata.
- Slightly expands the error messages in case of panic in `AnyValueEnum::new` and `BasicValueEnum::new`.
- Replaces the `unreachable!` of `BasicValueEnum::new` with a `panic!`, since it turns out to be reachable.
- Some testing.

There are still at least two unknowns in this version of the MR:

1. Right now, we return `None` for operands that are marked as `LLVMMetadataTypeKind` but that convert to `null`. Not sure that this is the right behavior.
2. So far, I've only ever used inkwell to inspect IR, not to produce it, so I'm not particularly confident about my test.

## Related Issue

issue #633 

## How This Has Been Tested

New test `test_metadata_as_operand`.

Tests launched on macOS 12.7.6  with `cargo test --features llvm18-1`.

The following tests do not pass on my machine, but as far as I can tell, it's not related:
- `test_object_file::test_symbol_iterator` (`assertion failed: has_symbol_a`)
- `test_targets::test_default_triple` (`unknown OS`)

Waiting to see if it fares better in CI.

## Option\<Breaking Changes\>

I didn't find any way to implement this change that did not introduce at least one variant to at least one `enum`.

## Checklist

### DIFF
--- a/src/values/basic_value_use.rs
+++ b/src/values/basic_value_use.rs
@@ -7,8 +7,6 @@ use crate::basic_block::BasicBlock;
 use crate::values::{AnyValueEnum, BasicValueEnum, MetadataValue};
 
 /// Either [BasicValueEnum], a [BasicBlock] or a [Metadata].
-///
-/// Note that [Metadata] variants are only constructed for LLVM 17+.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Operand<'ctx> {
     /// Represents a [BasicValueEnum].
--- a/src/values/instruction_value.rs
+++ b/src/values/instruction_value.rs
@@ -3,20 +3,17 @@ use llvm_sys::core::LLVMGetGEPSourceElementType;
 use llvm_sys::core::{
     LLVMGetAlignment, LLVMGetAllocatedType, LLVMGetFCmpPredicate, LLVMGetICmpPredicate, LLVMGetIndices,
     LLVMGetInstructionOpcode, LLVMGetInstructionParent, LLVMGetMetadata, LLVMGetNextInstruction, LLVMGetNumIndices,
-    LLVMGetNumOperands, LLVMGetOperand, LLVMGetOperandUse, LLVMGetPreviousInstruction,
+    LLVMGetNumOperands, LLVMGetOperand, LLVMGetOperandUse, LLVMGetPreviousInstruction, LLVMGetTypeKind,
     LLVMGetVolatile, LLVMHasMetadata, LLVMInstructionClone, LLVMInstructionEraseFromParent,
     LLVMInstructionRemoveFromParent, LLVMIsAAllocaInst, LLVMIsABasicBlock, LLVMIsAGetElementPtrInst, LLVMIsALoadInst,
-    LLVMIsAStoreInst, LLVMIsATerminatorInst, LLVMIsConditional, LLVMIsTailCall,
-    LLVMSetAlignment, LLVMSetMetadata, LLVMSetOperand, LLVMSetVolatile, LLVMValueAsBasicBlock,
+    LLVMIsAStoreInst, LLVMIsATerminatorInst, LLVMIsConditional, LLVMIsTailCall, LLVMSetAlignment, LLVMSetMetadata,
+    LLVMSetOperand, LLVMSetVolatile, LLVMTypeOf, LLVMValueAsBasicBlock,
 };
 #[llvm_versions(10..)]
 use llvm_sys::core::{LLVMGetAtomicRMWBinOp, LLVMIsAAtomicCmpXchgInst, LLVMIsAAtomicRMWInst};
-#[llvm_versions(17..)]
-use llvm_sys::core::{LLVMGetTypeKind, LLVMIsAValueAsMetadata, LLVMTypeOf};
 use llvm_sys::core::{LLVMGetOrdering, LLVMSetOrdering};
 use llvm_sys::prelude::LLVMValueRef;
 use llvm_sys::LLVMOpcode;
-#[llvm_versions(17..)]
 use llvm_sys::LLVMTypeKind;
 
 use std::{ffi::CStr, fmt, fmt::Display};
@@ -694,35 +691,12 @@ impl<'ctx> InstructionValue<'ctx> {
             return Some(Operand::Block(bb.expect("BasicBlock should always be valid")));
         }
 
-        match self.try_ingest_metadata(operand) {
-            // This is indeed metadata.
-            Ok(Some(metadata)) => return Some(Operand::Metadata(metadata)),
-            // This is null metadata, which we can't convert to an operand.
-            Err(_) => return None,
-            // Not metadata at all.
-            Ok(None) => {}
+        if unsafe { LLVMGetTypeKind(LLVMTypeOf(operand)) == LLVMTypeKind::LLVMMetadataTypeKind } {
+            return Some(Operand::Metadata(MetadataValue::new(operand)));
         }
 
         Some(Operand::Value(unsafe { BasicValueEnum::new(operand) }))
     }
-
-    #[llvm_versions(17..)] // LLVMIsAValueAsMetadata was introduced un 17.0.
-    unsafe fn try_ingest_metadata(self, operand: LLVMValueRef) -> Result<Option<MetadataValue<'ctx>>, ()> {
-        assert!(!operand.is_null());
-        if unsafe { LLVMGetTypeKind(LLVMTypeOf(operand)) == LLVMTypeKind::LLVMMetadataTypeKind } {
-            // We may be dealing with null metadata, which would break memory invariants.
-            if LLVMIsAValueAsMetadata(operand).is_null() {
-                return Err(());
-            }
-            return Ok(Some(MetadataValue::new(operand)));
-        }
-        Ok(None)
-    }
-    #[llvm_versions(..17)] // LLVMIsAValueAsMetadata was introduced un 17.0.
-    unsafe fn try_ingest_metadata(self, _: LLVMValueRef) -> Result<Option<MetadataValue<'ctx>>, ()> {
-        Ok(None)
-    }
-
 
     /// Get an instruction value operand iterator.
     pub fn get_operands(self) -> OperandIter<'ctx> {
--- a/src/values/enums.rs
+++ b/src/values/enums.rs
@@ -461,6 +461,7 @@ impl<'ctx> AggregateValueEnum<'ctx> {
 
 impl<'ctx> BasicMetadataValueEnum<'ctx> {
     pub(crate) unsafe fn new(value: LLVMValueRef) -> Self {
+        assert!(!value.is_null());
         match LLVMGetTypeKind(LLVMTypeOf(value)) {
             LLVMTypeKind::LLVMFloatTypeKind
             | LLVMTypeKind::LLVMFP128TypeKind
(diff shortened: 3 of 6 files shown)


src/values/instruction_value.rs
```diff
@@ -684,14 +684,21 @@ impl<'ctx> InstructionValue<'ctx> {
...
+
+        if unsafe { LLVMGetTypeKind(LLVMTypeOf(operand)) == LLVMTypeKind::LLVMMetadataTypeKind } {
+            // We may be dealing with null metadata, which would break memory invariants.
+            if LLVMIsAValueAsMetadata(operand).is_null() {
+                return None;
+            }
+            return Some(Operand::Metadata(MetadataValue::new(operand)));
         }
+
+        Some(Operand::Value(unsafe { BasicValueEnum::new(operand) }))
```
ErisianArchitect: This function is a little hard to follow the way that it's written. Maybe it should be an if/else chain instead?

ErisianArchitect:
`Operand` has some ease-of-use functions for the variants.

```rust
impl<'ctx> Operand<'ctx> {
    /// Determines if the [Operand] is a [BasicValueEnum].
    #[inline]
    #[must_use]
    pub fn is_value(self) -> bool {
        matches!(self, Self::Value(_))
    }

    /// Determines if the [Operand] is a [BasicBlock].
    #[inline]
    #[must_use]
    pub fn is_block(self) -> bool {
        matches!(self, Self::Block(_))
    }

    /// If the [Operand] is a [BasicValueEnum], map it into [Option::Some].
    #[inline]
    #[must_use]
    pub fn value(self) -> Option<BasicValueEnum<'ctx>> {
        match self {
            Self::Value(value) => Some(value),
            _ => None,
        }
    }

    /// If the [Operand] is a [BasicBlock], map it into [Option::Some].
    #[inline]
    #[must_use]
    pub fn block(self) -> Option<BasicBlock<'ctx>> {
        match self {
            Self::Block(block) => Some(block),
            _ => None,
        }
    }

    /// Expect [BasicValueEnum], panic with the message if it is not.
    #[inline]
    #[must_use]
    #[track_caller]
    pub fn expect_value(self, msg: &str) -> BasicValueEnum<'ctx> {
        match self {
            Self::Value(value) => value,
            _ => panic!("{msg}"),
        }
    }

    /// Expect [BasicBlock], panic with the message if it is not.
    #[inline]
    #[must_use]
    #[track_caller]
    pub fn expect_block(self, msg: &str) -> BasicBlock<'ctx> {
        match self {
            Self::Block(block) => block,
            _ => panic!("{msg}"),
        }
    }

    /// Unwrap [BasicValueEnum]. Will panic if it is not.
    #[inline]
    #[must_use]
    #[track_caller]
    pub fn unwrap_value(self) -> BasicValueEnum<'ctx> {
        self.expect_value("Called unwrap_value() on UsedValue::Block.")
    }

    /// Unwrap [BasicBlock]. Will panic if it is not.
    #[inline]
    #[must_use]
    #[track_caller]
    pub fn unwrap_block(self) -> BasicBlock<'ctx> {
        self.expect_block("Called unwrap_block() on UsedValue::Value.")
    }
}
```
Could you add one for the Metadata variant?

Yoric:
> Could you add one for the Metadata variant?

Done.

ErisianArchitect:
I just noticed an error that I made when I wrote the Operand enum code. I first called it `UsedValue`, then renamed it to `Operand`, but when I renamed it, I forgot to replaced `UsedValue` in the error messages with `Operand`. I'll have to open a PR to fix it.

TheDan64 (approved):
Thanks!

TheDan64:
The code does not work across all LLVM versions and needs to be adjusted accordingly

ErisianArchitect (changes requested):

src/values/instruction_value.rs
Yoric: I've had to rewrite it a bit to make it compatible with all LLVM versions. I believe that the new version should be a bit easier to read.

I'm not a big fan of if/else chains, but if you think it's better, I can convert the code.

ErisianArchitect (approved):

Yoric:
Apologies for the test failures, I'll need to install more versions of LLVM on my machine.

Given the speed of the beast, this will probably take the rest of the day 😓 .

ErisianArchitect:
If you need it, I have a couple python scripts that can help simplify the process of building versions of LLVM from source. It works for me on Windows, might also work on other OSes as well.

Yoric:
As far as I understand from looking at `llvm-sys`, the feature I depend on appeared with LLVM 17. So my tests now work with `llvm16-0` and `llvm18-1`, I _think_ that it should be sufficient.

Note that there are two tests broken on macOS with both versions, but this was the case already before my changes, afaict.

src/values/basic_value_use.rs
```diff
@@ -4,15 +4,19 @@ use llvm_sys::prelude::LLVMUseRef;
 use std::marker::PhantomData;
 
 use crate::basic_block::BasicBlock;
-use crate::values::{AnyValueEnum, BasicValueEnum};
+use crate::values::{AnyValueEnum, BasicValueEnum, MetadataValue};
 
-/// Either [BasicValueEnum] or [BasicBlock].
+/// Either [BasicValueEnum], a [BasicBlock] or a [Metadata].
+///
+/// Note that [Metadata] variants are only constructed for LLVM 17+.
```
ErisianArchitect: You should use cfg/cfg_attr here to make it so that the documentation depends on the version. It shouldn't mention `metadata` in versions that it doesn't appear in.

src/values/instruction_value.rs
```diff
@@ -684,16 +688,42 @@ impl<'ctx> InstructionValue<'ctx> {
...
+            Err(_) => return None,
+            // Not metadata at all.
+            Ok(None) => {}
         }
+
+        Some(Operand::Value(unsafe { BasicValueEnum::new(operand) }))
     }
 
+    #[llvm_versions(17..)] // LLVMIsAValueAsMetadata was introduced un 17.0.
+    unsafe fn try_ingest_metadata(self, operand: LLVMValueRef) -> Result<Option<MetadataValue<'ctx>>, ()> {
```
ErisianArchitect: There's really no reason to have this function. You can just cfg the match statement in get_operand_unchecked. This adds unnecessary complexity.

ErisianArchitect (commented):
I think there are some issues with exposure of MetadataValue to versions that don't actually have a way to construct a MetadataValue operand.

But I'm not entirely sure about the usage of the API.

src/values/instruction_value.rs
```diff
@@ -684,16 +688,42 @@ impl<'ctx> InstructionValue<'ctx> {
...
-        } else {
-            Some(Operand::Value(unsafe { BasicValueEnum::new(operand) }))
+            return Some(Operand::Block(bb.expect("BasicBlock should always be valid")));
+        }
+
+        match self.try_ingest_metadata(operand) {
+            // This is indeed metadata.
+            Ok(Some(metadata)) => return Some(Operand::Metadata(metadata)),
+            // This is null metadata, which we can't convert to an operand.
+            Err(_) => return None,
```
ErisianArchitect: This error should be passed on to the caller, which means that this function should return a Result.

ErisianArchitect (changes requested):

ErisianArchitect:
Okay, I did some digging. Creating MetadataValue Operands is valid prior to LLVM 17, so this conditional code is invalid. I may be wrong, but you should be able to just conditionally compile out
```rust
            if LLVMIsAValueAsMetadata(operand).is_null() {
                return Err(());
            }
```
In versions prior to LLVM 17, and it should work as expected. I believe there are also other Metadata types beyond ValueAsMetadata, so this could also cause issues in cases where a Metadata type is not ValueAsMetadata.

Yoric:
> Okay, I did some digging. Creating MetadataValue Operands is valid prior to LLVM 17, so this conditional code is invalid. I may be wrong, but you should be able to just conditionally compile out
> 
> ```rust
>             if LLVMIsAValueAsMetadata(operand).is_null() {
>                 return Err(());
>             }
> ```
> 
> In versions prior to LLVM 17, and it should work as expected. I believe there are also other Metadata types beyond ValueAsMetadata, so this could also cause issues in cases where a Metadata type is not ValueAsMetadata.

I'm not sure I follow. In my tests with LLVM 18, removing this check causes segfaults. Are you saying that it works with LLVM < 17?

ErisianArchitect:
I'm saying that the ability for an Operand to be a MetadataValue, as far as I'm aware, predates 17 all the way to around version 3. Removing the check might cause segfaults because it might be a Value type (which is checked last) rather than a metadata type. That's my best guess.
But I'm likely wrong.

If a Metadata Operand type is invalid prior to version 17, then none of the code involving Metadata Operand should be included in any version prior to 17. It all needs to be conditionally compiled out. But since Operands as MetadataValues predates version 17, I think that we need to go back to the drawing board here.

Although I should probably take a closer look.

src/values/instruction_value.rs
ErisianArchitect: If not breaking compatibility results in an uncaught error, what choice is there? As it's written now, it could cause subtle bugs that would be hard to trace.

Yoric: Good point.

Yoric:
Alright, I've taken a deeper look at where the segfaults come from in my tests.

As it turns out, they show up in `impl fmt::Debug for MetadataValue` because we have `null` operands, which break somewhere in `MetadataValue::get_node_values`. I've plugged that particular hole (and simplified my code along the way), but I'm not entirely confident that it solves all the issues.

ErisianArchitect:
It's going to be a pain, but I think this is something that is going to require a much closer look to figure out. I think that Dan would agree with me here.

Yoric:
@ErisianArchitect Do you think the current implementation is worth pursuing? Otherwise, any suggestion?

ErisianArchitect:
It's supported via llvm-sys, so it makes sense to have it. The problem with your code is that `LLVMIsAValueAsMetadata` was added in LLVM 17, but the ability to create Metadata values exists prior to LLVM 17.

That means that for versions prior to LLVM 17, this code:
```rust
        if unsafe { LLVMGetTypeKind(LLVMTypeOf(operand)) == LLVMTypeKind::LLVMMetadataTypeKind } {
            // We may be dealing with null metadata, which would break memory invariants.
            if LLVMIsAValueAsMetadata(operand).is_null() {
                return Err(());
            }
            return Ok(Some(MetadataValue::new(operand)));
        }
```
would compile not compile. But it WOULD compile if you removed that inner if-statement that calls `LLVMIsAValueAsMetadata`.
That's why I think this needs a closer look, because I wonder if it would be an error to attempt to return an Operand::Value when it would otherwise be Metadata.

Yoric:
Note that I've removed `LLVMIsAValueAsMetadata` in the latest version of the PR.

ErisianArchitect:
Ahh, I missed that. Does it work as expected for all three variants?

Yoric:
> Ahh, I missed that. Does it work as expected for all three variants?

Yeah, I believe that it does.

ErisianArchitect:
I looked into this, and the error from #633 is coming from `BasicValueEnum::new`. It has a default arm with `unreachable` in it, and the unreachable is getting hit because there are more match arms that haven't been filled out. I don't know which match arm is getting hit, but I don't think adding a panic there is going to be sufficient.

PR #636, opened 2025-11-28
Fixes type name in `expect` message for `Operand`

ErisianArchitect:
## Description

When I made the `Operand` type in `basic_value_use.rs`, I initially named it `UsedValue`, but then later renamed it to `Operand` before submitting the pull request. When I renamed it, I forgot to change the type name used in a couple of error messages. This PR fixes that name issue.

## How This Has Been Tested

Nothing to be tested, the only change is a string.

## Checklist


ErisianArchitect:
I forgot that this was just addressed in another PR. Sorry, I just woke up.

Issue #568, opened 2025-02-07
Add support for inbounds, nsw, nuw, exact flags


Benjins:
**Is your feature request related to a problem? Please describe.**
The instruction-specific flags (e.g.  `inbounds`, `nsw`, `nuw`, `exact`) do not seem to be exposed (I could not find any docs on them, and I don't see the relevant llvm-sys API calls in the code). Specifically, `add`/`sub`/`mul` instructions can have `nuw`/`nsw` flags, `udiv`/`sdiv`/`urem`/`srem` instructions can have the `exact` flag, and `getelementptr` instructions can have the `inbounds` flag. These should all now be exposed via llvm-sys, though some only in more recent LLVM versions

**Describe the solution you'd like**
This information should be exposed somehow. I see that the fast-math flags, disjoint, volatile, etc. are exposed on the `InstructionValue` struct. That seems like a logical place to expose these as well, unless there's another place that is considered more appropriate or there are concerns about `InstructionValue` taking on too much surface

**Additional context**
[LangRef - Binary arithmetic ops](https://releases.llvm.org/19.1.0/docs/LangRef.html#binary-operations)
[LangRef - GEP](https://releases.llvm.org/19.1.0/docs/LangRef.html#getelementptr-instruction)

There are likely other flags coming down the pipeline: [the `samesign` flag for `icmp` instructions](https://github.com/llvm/llvm-project/pull/111419), for example

If it's as simple as adding getter/setter methods to `InstructionValue`, I can take a crack at a PR. Figured it was worth opening an issue to see if there were other thoughts

TheDan64:
Sorry for missing this issue; if llvm-sys / the underlying C api support it then we should too

PR #649, opened 2026-01-11
Add `FastMathFlags` struct in place of u32

my4ng:
## Description

This adds the `FastMathFlags` struct created from `bitflags` and `llvm_sys` exposed constants. It avoids the raw u32 values, and has a clean and safer syntax to create and manipulate flags, e.g.

```rust
let flags = FastMathFlags::NoNans | FastMathFlags::NoInfs;

let fast_flags = FastMathFlags::all();

let none_flags = FastMathFlags::empty();

let is_no_nans = flags.contains(FathMathFlags::NoNans);
```

Additional items related to `InstructionValue` referenced in #568 will be added in later PRs.

## Related Issue

Related to #568.

## How This Has Been Tested

`cargo test -F llvm21-1`

## Breaking Changes

`get_fast_math_flags` returns `Option<FastMathFlags>` and `set_fast_math_flags` takes `FastMathFlags` in lieu of `Option<u32>` and `u32` respectively.

## Checklist

### DIFF
--- a/src/types/enums.rs
+++ b/src/types/enums.rs
@@ -117,9 +117,6 @@ impl<'ctx> BasicMetadataTypeEnum<'ctx> {
     ///
     /// Undefined behavior if the referenced type cannot be represented as [`BasicMetadataTypeEnum`],
     /// or the underlying pointer is null.
-    ///
-    /// Before LLVM 6, [`BasicMetadataTypeEnum::MetadataType`] variants cannot be created
-    /// with this function. Attempting to do results in undefined behavior.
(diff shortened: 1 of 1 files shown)


my4ng:
On the note of breaking changes, I think there is merit to replace silent no-ops in `set_fast_math_flags` and other methods with returning `Result<(), InstructionValueError>`. In general, I don't think it is a good idea that different-from-expectation instruction types should go undetected; panic on unwrapping the result allows subtle bugs to be caught early.

TheDan64:
Yes, I'd be fine with changing the fast math return types to Results

PR #652, opened 2026-01-17
Change instr method signatures

my4ng:
## Description

More breaking changes to normalise instr method signatures:
- `is_...` methods return `bool` when it tests for instruction value type, or a general property, e.g. `is_terminator`.
- `is_...` methods return `Result<bool, InstructionValueError>` when it tests for a property of a specific type, `Err(...)` if not the correct type or any other error, and `Ok(true/false)` otherwise, e.g. `is_conditional`.
- `get_...` methods return `Result<T, InstructionValueError>` for getter for a property of a specific type, `Err(...)` if not the correct type or any other error, and `Ok(T)` otherwise, e.g. `get_non_negative_flag`.
  - `get_metadata` is not in this format because it is not limited to any specific type.
- `set_...` methods return `Result<(), InstructionValueError>` for setter for a property of a specific type, `Err(...)` if not the correct type or any other error, and `Ok(T)` otherwise, e.g. `set_non_negative_flag`.

## Related Issue

#649.

## How This Has Been Tested

`cargo test -F llvm21-1`

## Breaking Changes

- `is_conditional` now returns `Result<bool, InstructionValueError>`. (also fixed doc comment)
- `is_tail_call` now returns `Result<bool, InstructionValueError>`.
- `get_tail_call_kind` now returns `Result<super::LLVMTailCallKind, InstructionValueError>`. (also added `set_tail_call_kind`)
- `get_non_negative_flag` now returns `Result<bool, InstructionValueError>`.
- `set_non_negative_flag` now returns `Result<(), InstructionValueError>`.
- `get_disjoint_flag` now returns `Result<bool, InstructionValueError>`.
- `set_disjoint_flag` now returns `Result<(), InstructionValueError>`.

## Checklist

### DIFF
--- a/src/values/instruction_value.rs
+++ b/src/values/instruction_value.rs
@@ -67,6 +67,8 @@ pub enum InstructionValueError {
     NotZextInst,
     #[error("Not an or instruction.")]
     NotOrInst,
+    #[error("Not a branch instruction.")]
+    NotBrInst,
     #[error("Atomic Error: {0}")]
     AtomicError(AtomicError),
     #[error("Metadata is expected to be a node.")]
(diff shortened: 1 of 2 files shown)


src/values/instruction_value.rs
```diff
@@ -259,38 +265,52 @@ impl<'ctx> InstructionValue<'ctx> {
         unsafe { !LLVMIsATerminatorInst(self.as_value_ref()).is_null() }
     }
 
-    // SubTypes: Only apply to terminators
-    /// Returns if a terminator is conditional or not
-    pub fn is_conditional(self) -> bool {
+    // SubTypes: Only apply to branch instructions
+    /// Returns `Some` with whether or not the branch is conditional,
+    /// otherwise `None` if not a branch instruction
+    pub fn is_conditional(self) -> Option<bool> {
```
TheDan64: Shouldn't these return Result? It seems like it would be an error to call it on the wrong type

my4ng: Yes and no. I have the getter/setter return `Result<_, _>` and `is_...` return `Option<_>` as in the second rule.

PR #655, opened 2026-01-25
Memory buffer v2

my4ng:
## Description

This fixes two issues with the previous design.

1. The `&CStr` input was overly restrictive. It is OK to have interior nul bytes, as shown in object file tests where in memory buffer read from binary ELF files, most bytes are indeed nul. It will now panic only if the last byte is not nul.

2. The input data length in creation is now correct at input.len() - 1, as the nul byte should be at one past the end of the actual data. Similarly, the get_size (and hence as_slice) have the length plused one to include the trailing nul byte. This should now align with the LLVM documentation.

It was an oversight of mine to not have noticed them in the C++ documentation and testing with actual binary files. Additional tests have been added to ensure compliance with LLVM.

## Related Issue

#654

## How This Has Been Tested

`cargo test -F llvm21-1`

## Option\<Breaking Changes\>

- Reverted `create_from_memory_range` and `create_from_memory_range_copy` to take &[u8] input once again.

## Checklist

### DIFF
--- a/src/memory_buffer.rs
+++ b/src/memory_buffer.rs
@@ -10,7 +10,6 @@ use llvm_sys::prelude::LLVMMemoryBufferRef;
 use crate::object_file::ObjectFile;
 use crate::support::{to_c_str, LLVMString};
 
-use std::ffi::CStr;
 use std::marker::PhantomData;
 use std::mem::forget;
 use std::path::Path;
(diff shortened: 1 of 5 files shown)


TheDan64 (approved):

ErisianArchitect:
This pull-request has a bug.

```rust
    pub fn create_from_memory_range(input: &'a [u8], name: &str) -> Self {
        assert_eq!(
// ----------------- BUG IS HERE ---------------------------
            input[input.len() - 1],
            b'\0',
            "input byte slice must terminate with a nul byte"
        );

        let name_c_string = to_c_str(name);

        let memory_buffer = unsafe {
            LLVMCreateMemoryBufferWithMemoryRange(
                input.as_ptr() as *const libc::c_char,
                input.len() - 1,
                name_c_string.as_ptr(),
                false as i32, // guaranteed to have nul-terminator by CStr
            )
        };

        unsafe { MemoryBuffer::new(memory_buffer) }
    }
```

PR #666, opened 2026-03-17
feat: port to 2024 edition

marxin:
The vast majority of the changes were created by `cargo clippy --fix` and `cargo fmt`, I recommend looking at the individual commits in the PR.


my4ng:
This does raise the MSRV to at least 1.85, is this desirable since currently the one listed in readme is 1.56?

TheDan64:
Good catch, we should update the readme

marxin:
Done is #667.

PR #667, opened 2026-03-19
docs: reflect MSRV and latest supported LLVM version

marxin:
I would remove the `Stable, Beta, or Nightly` part as it might be confusing for a library customer.


TheDan64 (approved):
Thanks!

Issue #505, opened 2024-06-16
Most (all?) passes have a maximum LLVM version number of 16.


nielsdos:
**Describe the Bug**

Most passes have a maximum LLVM version number of 16.
For example, the type based alias analysis pass has: https://github.com/TheDan64/inkwell/blob/5c9f7fcbb0a667f7391b94beb65f1a670ad13221/src/passes.rs#L1055-L1058
However, this pass is also available in LLVM 17 and 18.
Looking at the passes.rs file, it looks like most (or all?) passes define a maximum version number of 16, making them impossible to use with LLVM 17.

**To Reproduce**

```rust
let pm = PassManager::create(());
pm.add_type_based_alias_analysis_pass(); // Doesn't work
```

**Expected Behavior**

I expect these passes to be available for LLVM 17.

**LLVM Version (please complete the following information):**
 - LLVM Version: 17.0.6
 - Inkwell Branch Used: 0.4

**Desktop (please complete the following information):**
 - OS: Linux

**Additional Context**

N/A

TheDan64:
Not sure why this happened, thanks for reporting

vertexclique:
Are these removed in 18? Don't we need to bump all of them compatible up to 18?

nielsdos:
Most (all? perhaps some under a different name?) of these optimizations are still available in LLVM 18

vertexclique:
I searched for the new pass manager doc, I couldn't find it tbh. They made it into classes where some of these are bundled into as multiple passes. https://releases.llvm.org/18.1.1/docs/NewPassManager.html#just-tell-me-how-to-run-the-default-optimization-pipeline-with-the-new-pass-manager

vertexclique:
Btw, setting through PassManagerOptions on the master doesn't apply the underlying passes. We need to write the current pass list in 18 with new methods (whereever they are).

GoncaloBranquinho:
> I searched for the new pass manager doc, I couldn't find it tbh. They made it into classes where some of these are bundled into as multiple passes. https://releases.llvm.org/18.1.1/docs/NewPassManager.html#just-tell-me-how-to-run-the-default-optimization-pipeline-with-the-new-pass-manager

I was struggling to find documentation regarding the new pass manager, and I realized that Inkwell still uses the legacy pass manager for the `PassManager` struct (please correct me if I’m wrong), while the `PassBuilderOptions` struct is already designed for the new pass manager.

[...]

PR #670, opened 2026-03-21
Deprecate legacy PassManager in favor of PassBuilderOptions and run_passes

GoncaloBranquinho:
## Description

This PR deprecates the `PassManager` struct.

The legacy `PassManager<T>` is replaced by the new pass manager workflow:

- `PassBuilderOptions`
- `Module::run_passes`

This aligns Inkwell with LLVM's new pass manager infrastructure.

The `PassManager` struct  will be removed once LLVM 16 support is dropped.

## Related Issue

Issue #505 

## How This Has Been Tested

- Ran `cargo clippy --features llvm20-1`, which completed successfully and only reported warnings related to deprecated APIs,  which are expected due to ongoing transition to the new LLVM pass manager infrastructure.
- Ran `cargo test --features llvm20-1`, with all tests passing. (328 passed, 17 ignored, consistent with master branch)
-  Ran `cargo doc --no-deps --features llvm20-1`, documentation builds successfully with no errors.

## Checklist

### DIFF
--- a/src/passes.rs
+++ b/src/passes.rs
@@ -150,6 +150,7 @@ impl PassManagerBuilder {
     ///
     /// pass_manager_builder.populate_function_pass_manager(&fpm);
     /// ```
+    #[allow(deprecated)]
     pub fn populate_function_pass_manager(&self, pass_manager: &PassManager<FunctionValue>) {
         unsafe {
             LLVMPassManagerBuilderPopulateFunctionPassManager(self.pass_manager_builder, pass_manager.pass_manager)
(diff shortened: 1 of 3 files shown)


src/passes.rs
```diff
@@ -265,6 +265,10 @@ impl<'ctx> PassManagerSubType for FunctionValue<'ctx> {
 /// documentation for specific passes is directly from the [LLVM
 /// documentation](https://llvm.org/docs/Passes.html).
 #[derive(Debug)]
+#[deprecated(
+    since = "0.8.0",
```
TheDan64: Should be 0.9.0 as that's the next version

GoncaloBranquinho: I used #[allow(deprecated)] on the internal impl blocks in passes.rs and on the test file, since PassManager needs to exist across all  LLVM versions while LLVM 16 is still supported, and cfg didn't seem appropriate here. But please let me know if you'd prefer a different approach.

Issue #634, opened 2025-11-25
doc.rs fails to build the documentation


Yoric:
As of this day, https://docs.rs/crate/inkwell/latest or https://docs.rs/crate/inkwell/0.7.0 or https://docs.rs/crate/inkwell/0.7.1 all indicate ` docs.rs failed to build inkwell-XXX`.

The logs show error
```
error: One of the LLVM feature flags must be provided: llvm8-0 llvm9-0 llvm10-0 llvm11-0 llvm12-0 llvm13-0 llvm14-0 llvm15-0 llvm16-0 llvm17-0 llvm18-1 llvm19-1 llvm20-1 llvm21-1
```

TheDan64:
I don't think there's a way to fix this. docs.rs makes some assumptions about how docs should be generated which don't really apply to inkwell (ie we need a version feature flag)

Yoric:
> ie we need a version feature flag

Just a random idea: could there perhaps be a virtual package inkwell-doc that just somehow serves to build the doc, using the most recent version of llvm?

Alternatively, perhaps https://thedan64.github.io/inkwell/inkwell/ could be brought to a more recent version, then?

ErisianArchitect:
@TheDan64 I believe that this issue should also be closed.

TheDan64:
Yes, thanks!

PR #673, opened 2026-04-03
Fix docs.rs build

AdamGS:
## Description

This PR should fix the docs build in `docs.rs` by enabling a a non-linking feature of LLVM

## Related Issue

https://github.com/TheDan64/inkwell/issues/634

## How This Has Been Tested

Building with and without this feature, I've seen and done similar changes in other projects in the past (more commonly so that docs show all available features), the full spec of that toml table is [here](https://docs.rs/about/metadata).

When running locally, the docs build fails in the same way with default features but works with the feature enabled.

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -178,5 +178,8 @@ regex = "1"
 version = "0.62.2"
 features = ["Win32_System_SystemInformation", "Win32_System_Memory"]
 
+[package.metadata.docs.rs]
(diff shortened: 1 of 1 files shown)


TheDan64:
Thank you!!! Never knew how to fix docs.rs

TheDan64 (approved):

PR #677, opened 2026-04-08
Enhance Type Safety v1.1 via Const Generics

### LINKED ISSUE #8: Type Safety v2
Today, `build_int_add` looks approximately like `build_int_add(&self, left: &IntValue, right: &IntValue) -> IntValue`. This is great, because it'll stop you from trying to add a `FloatValue` and an `IntValue`, for instance which would not work in LLVM. But I wonder if we can take type checking a step further! What happens when left is a `u8` and right is a `u16`? Needs to be verified, but I believe this is also a LLVM error because it doesn't make much sense. How would adding two values of different sizes work (without casting)? 

Therefore, I wonder if we can add sub-type annotations using generics (here-on known as sub-types). For example, `IntType<u32>` and `IntValue<u32>` so that you could express something like: `build_int_add<T>(&self, left: &IntValue<T>, right: &IntValue<T>) -> IntValue<T>` which would ensure only the same sub-types are added together when needed. So, `IntValue<u8>` and `IntValue<u8>` would be able to be added together but it'd prevent `IntValue<u8>` and `IntValue<u16>` from being added together, requiring a cast first. And, in case you do want different sub-types to be valid input you would just specify separate type variables: `do_foo<L, R>(&self, left: &IntType<L>, right: &IntType<R>)`

In terms of implementation details, the sub-type should basically just be a marker and not take up any additional space (as PhantomData if needed)

Outstanding questions:
* How would custom & 80 bit width types be handled given that they don't map to rust types (yet?)? It may be permissible to disallow custom & 80 bit types since they are far less common. Or just find some workaround. Thankfully, 128 bit is almost here :)
* `usize` should probably be a valid sub-type for ints, but I wonder if there's any issue with it's variable width nature? LLVM *should* allow us to map it to the pointer size at runtime
* Signatures will quickly get verbose. Just imagine a simple `String` type: `StructType<(PointerType<IntType<u8>>, IntType<usize>, IntType<usize>)>` Will the compiler be able to hide most of this from the user so that they don't have to specify annotations themselves?
* Could the `num` crate's traits help with float and int sub-types?
* Should signed types be explicit(IE `IntType<i8>`) or just an irrelevant fact about an `IntType<u8>`? I'm guessing LLVM doesn't let you mix signed and unsigned addition without casting first, so I'm leaning towards keeping them explicit.

meftunca:
Resolves #8 (for Integers) using Zero-Cost Const Generics abstraction.

### DIFF
--- a/src/typed/value.rs
+++ b/src/typed/value.rs
@@ -1,5 +1,6 @@
 use std::convert::TryFrom;
-use crate::values::IntValue;
+use std::marker::PhantomData;
+use crate::values::{IntValue, FloatValue, PointerValue, StructValue};
 
 /// A strictly-typed integer value wrapped around `IntValue`.
 /// The const generic `WIDTH` guarantees the exact bit width of the integer at compile-time.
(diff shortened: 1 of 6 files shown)


src/typed/value.rs
```diff
@@ -0,0 +1,114 @@
...
+use crate::values::{IntValue, FloatValue, PointerValue, StructValue};
+
+/// A strictly-typed integer value wrapped around `IntValue`.
+/// The const generic `WIDTH` guarantees the exact bit width of the integer at compile-time.
+#[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
+pub struct TypedIntValue<'ctx, const WIDTH: u32> {
+    pub(crate) int_value: IntValue<'ctx>,
+}
+
+impl<'ctx, const WIDTH: u32> TypedIntValue<'ctx, WIDTH> {
```
ErisianArchitect, line 12: What's stopping someone from using an invalid width? Using a const generic does not solve that problem.

ErisianArchitect (commented):
It doesn't really make sense to use a u32 for W, it makes more sense to use the actual types such as `u8`, `u16`, `u32`, etc.

TheDan64:
Although this is a step in the direction I imagined, the PR does not seem well thought out and reasoned. Closing as AI spam

Issue #66, opened 2019-01-29
Can't use builder which belongs to a different context


TheDan64:
**Describe the Bug**
Somehow newer rust versions (late 2018+) expose unsafe behavior (maybe due to LTO?). Anyway, mixing and matching unrelated context and builder seems to produce a segfault.

**To Reproduce**
Uncomment this line and run test: https://github.com/TheDan64/inkwell/blob/b84df07e8c6ee7ebbff9af6c3f3033a8a40b63fc/tests/all/test_builder.rs#L441

**Expected Behavior**
Successful test and no segfault

**LLVM Version (please complete the following information):**
 - LLVM Version: 7.0, probably others
 - Inkwell Branch Used: 7.0, probably others

**Desktop (please complete the following information):**
 - OS: Ubuntu 16.04

**Additional Context**
Not totally sure what to do about this one. I'm not sure there's a way to ensure builder is totally only used with objects of the same context, even at runtime?

0xFE4415:
Could be doable with: https://docs.rs/generativity/latest/generativity/

TheDan64:
@0xFE4415 super cool! Thanks for sharing. If I understand correctly, it's like GhostCell but without the closure?

0xFE4415:
I tried using `generativity` for the first time this week 😅

If I understand this correctly, I think you'd still need something along the lines of:

```rust
impl Context {
    pub fn with_context<R>(&self, f: impl for<'id> FnOnce(ContextRef<'id>) -> R) -> R {
        make_guard!(guard);
        f(ContextRef {
            context: self.context,
            _brand: Brand::from(guard),
        })
    }
}
```

And on top of that, a custom wrapper around `Id`, because `BasicBlock` is `#[repr(transparent)]`, so the compiler complains that *"this field contains `Id<'_>`, which has private fields, so it could become non-zero-sized in the future."*

I also don't see how this would work without removing (or at least deprecating) the existing create_* APIs on Context, since I don't see a way to ensure lifetime uniqueness without the closure.

ErisianArchitect:
Maybe I'm not understanding the problem being solved here, but couldn't you prevent builders that came from different contexts from being used by having some sort of context ID that is passed on to the builder?

0xFE4415:
You could do that too. I think something along these lines was vibe-coded here: https://github.com/TheDan64/inkwell/pull/679/changes

TheDan64:
[...]

PR #679, opened 2026-04-09
Fix: LLVM segfaults from mixed Builder/Value contexts (#66)

meftunca:
### What does this PR do?
This PR addresses issue #66 by enforcing runtime safety bounds to prevent LLVM Segfaults caused by invalid disjoint Context bindings.
Additionally, it introduces a new `orc` module for LLJIT and ThreadSafeContext wrappers.

When developers write flow logic crossing disjoint `Context`s in the same scope, Rust's borrow checker can't identify the mismatch in local pointer lifetimes, resulting in an opaque C++ layer segfault.

This introduces a set of configurable debug assertion hooks (`check_val_context` and `check_type_context`) inside `Builder` that intercept common `BasicValue` endpoints and panic gracefully with a clear Rust trace in Debug mode, completely zero-cost in Release mode.

### Implementation details
- Tracks active `ContextRef<'ctx>` securely inside `Builder<'ctx>`.
- Hooks checked in control flows, calls, and standard instructions (e.g., `build_store`, `build_return`, `build_call`, `build_struct_gep`).
- Included Unit test coverage: `test_builder_context_mismatch_panics`.
- Adds basic `src/orc/` namespace with `LLJIT` wrapper bindings.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -120,20 +120,50 @@ pub enum BuilderError {
 pub struct Builder<'ctx> {
     builder: LLVMBuilderRef,
     positioned: Cell<PositionState>,
-    _marker: PhantomData<&'ctx ()>,
+    context: ContextRef<'ctx>,
 }
 
 #[allow(unused)] // only used in documentation
 use crate::context::Context;
+use crate::context::ContextRef;
 
 impl<'ctx> Builder<'ctx> {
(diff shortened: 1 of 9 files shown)


TheDan64:
While the builder changes are somewhat interesting, the jit code is completely unrelated and shouldn't be combined. Closing as vibe coded

Issue #516, opened 2024-07-18
Unsoundness in `MemoryBuffer::create_from_memory_range`


fuzzypixelz:
**Describe the Bug**

`MemoryBuffer::create_from_memory_range` calls `LLVMCreateMemoryBufferWithMemoryRange` with `RequiresNullTerminator` set to `false`, while an LLVM memory buffer should always be null-terminated:

https://github.com/llvm/llvm-project/blob/adacb5010f5ca6e923b3cf2d8ea47cbaab96099d/llvm/include/llvm/Support/MemoryBuffer.h#L41-L51

Please note that setting `RequiresNullTerminator` to `true` won't solve the issue; LLVM will simply throw an error. Instead, the function signature needs to be changed to require null-terminated buffers. There can be no zero-copy implementation of `MemoryBuffer::create_from_memory_range` for arbitrary `&[u8]`.

The `RequiresNullTerminator` argument is truly baffling to me. It seems to mean "if `true`, check if the buffer is null-terminated" and not "I don't need this memory buffer to be null terminated" because it is always assumed to be so (perhaps you want to patch it after construction):

https://github.com/llvm/llvm-project/blob/main/llvm/lib/Support/MemoryBuffer.cpp#L49-L57

**To Reproduce**

Create a memory buffer with `MemoryBuffer::create_from_memory_range` and attempt to use it in `Context::create_module_from_ir` for instance. The parser will behave oddly as it tries to read past the buffer.

**Expected Behavior**

`MemoryBuffer::create_from_memory_range` should always pass a null-terminated buffer to `LLVMCreateMemoryBufferWithMemoryRange` and set `RequiresNullTerminator` to `true`. 

As this requires a copy in the general case, it should take a type that represents a buffer `b` of length `N` where `b[N - 1] == b'\0'` and set the buffer length to `N-1` in `LLVMCreateMemoryBufferWithMemoryRange`.

Otherwise the documentation should be updated to ask the user to do the the necessary manipulation themselves before calling the function.

**LLVM Version (please complete the following information):**
 - LLVM Version: 17.0.6
 - Inkwell Branch Used: master

**Desktop (please complete the following information):**
 - OS: macOS 14.2.1

**Additional Context**

N/A

TheDan64:
Dupe of #185 I think

Issue #304, opened 2022-02-20
`add_function_inlining_pass` segfaults


fuchsnj:
I'm playing around with running the pass manager on my code. The default passes from the [kaleidoscope example](https://github.com/TheDan64/inkwell/blob/b6984c03d88107771710a11478eb1ecdc8a72759/examples/kaleidoscope/main.rs#L1237-L1244) all seem to work well. As soon as I add `add_function_inlining_pass` it segfaults immediately.

Code
```rust
    let llvm_context = Context::create();
    let module = llvm_context.create_module("module");
    module.verify().unwrap();

    let pm: PassManager<FunctionValue> = inkwell::passes::PassManager::create(&module);
    pm.add_function_inlining_pass();
```

Result
```
Segmentation fault (core dumped)
```

Version
```
inkwell = { git = "https://github.com/TheDan64/inkwell", branch = "master", features = ["llvm13-0"] }
```

OS is `Ubuntu 21.10`

folkertdev:
a guess: you make a `FunctionValue` passmanager, but `add_function_inlining_pass` is pass that only works on modules, not on individual functions. 

It would be great to have more type safety here, because for most passes it's actually unclear what they operate on. Many seem to work (that is, don't segfault) for either, but do they actually do something useful? it's unclear.

Issue #185, opened 2020-05-02
Undefined behavior with `MemoryBuffer::create_from_memory_range`


novacrazy:
## Describe the Bug:

This code with result in a panic from LLVM saying `expected top-level entity` from the LLVM IR.
```rust
pub static PRELUDE_LLVM_IR: &[u8] = include_bytes!("prelude.llvm");

#[inline(never)]
pub fn build_prelude<'ctx>(ctx: &JitContext<'ctx>) -> Module<'ctx> {
    let module = ctx
        .create_module_from_ir(MemoryBuffer::create_from_memory_range(
            PRELUDE_LLVM_IR, 
            "prelude"
        ))
        .unwrap();

    module
}
```

However, if I were to mark `build_prelude` as `#[inline(always)]`, suddenly it doesn't panic. `JitContext` is just a small wrapper around `inkwell::context::Context`.

## Solution

Use `MemoryBuffer::create_from_memory_range_copy`. Seems `create_from_memory_range` leaking memory causes it to become corrupted or something funky, but if it's inline then the value is kept on the stack long enough for LLVM to parse it. I don't know. I've also only just discovered this, so it's possible `create_from_memory_range_copy` may also have issues, but so far it at least fixes that odd behavior.

## Details

I build debug with `opt-level=3`, so perhaps that has something to do with it.

 - LLVM Version: 10.0
 - Inkwell Branch Used: master
 - OS: Windows 10

If you're curious, here is the current `prelude.llvm`. It's just some intrinsic stuff with fast call. 
<details>

```llvm
; ModuleID = 'prelude'
source_filename = "prelude"

declare <4 x float> @llvm.fma.v4f32(<4 x float>, <4 x float>, <4 x float>) #0

; Function Attrs: nounwind inlinehint
define <4 x float> @raygon.fma.v4f32(<4 x float> %0, <4 x float> %1, <4 x float> %2) #1 {
entry:
    %res = call fast <4 x float> @llvm.fma.v4f32(<4 x float> %0, <4 x float> %1, <4 x float> %2)
    ret <4 x float> %res
}

attributes #0 = { nounwind readnone speculatable willreturn }
attributes #1 = { nounwind inlinehint }
```
</details>

TheDan64:
Good find, thanks!

Kixiron:
[...]

PR #680, opened 2026-04-09
Fix: MemoryBuffer UB and PassManager Runtime Segfaults (#185, #304)

meftunca:
Addresses two important soundness and stability issues from Milestone 8:

1. **MemoryBuffer UB (#185, #516)**:
   - Made `MemoryBuffer::create_from_memory_range` and its `_copy` variant `pub unsafe fn`.
   - Removed the runtime panics and documented the requirement for the caller to provide a legally null-terminated slice, eliminating a false sense of safety.
   - Updated all internal test usages with `unsafe {}` blocks.

2. **PassManager Type Safety (#304)**:
   - Moved 15 module-only passes (e.g., `add_function_inlining_pass`, `add_argument_promotion_pass`, `add_always_inliner_pass`) from the generic `PassManager<T>` implementation into an explicit `impl<'ctx> PassManager<Module<'ctx>>`.
   - Adding a module level pass to a `FunctionValue` pass manager will now result in a clean Rust compiler error instead of an opaque runtime LLVM segmentation fault.

This eliminates immediate footguns and brings the API closer to true zero-cost abstractions for safe usage.

### DIFF
--- a/src/memory_buffer.rs
+++ b/src/memory_buffer.rs
@@ -65,16 +65,11 @@ impl MemoryBuffer<'static> {
 
     /// Create a memory buffer copied from a byte slice with a trailing nul byte.
     ///
-    /// # Panics
+    /// # Safety
     ///
-    /// Panics if the input byte slice does not terminate with a nul byte.
-    pub fn create_from_memory_range_copy(input: &[u8], name: &str) -> Self {
-        assert_eq!(
-            input[input.len() - 1],
-            b'\0',
-            "input byte slice must terminate with a nul byte"
-        );
-
+    /// The `input` byte slice must confidently terminate with a nul byte (`\0`),
+    /// otherwise LLVM will experience out-of-bounds undefined behavior/segfaults.
+    pub unsafe fn create_from_memory_range_copy(input: &[u8], name: &str) -> Self {
         let name_c_string = to_c_str(name);
 
         let memory_buffer = unsafe {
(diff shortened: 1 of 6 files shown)


src/memory_buffer.rs
```diff
@@ -65,16 +65,11 @@ impl MemoryBuffer<'static> {
...
-    pub fn create_from_memory_range_copy(input: &[u8], name: &str) -> Self {
-        assert_eq!(
-            input[input.len() - 1],
-            b'\0',
-            "input byte slice must terminate with a nul byte"
-        );
-
+    /// The `input` byte slice must confidently terminate with a nul byte (`\0`),
+    /// otherwise LLVM will experience out-of-bounds undefined behavior/segfaults.
+    pub unsafe fn create_from_memory_range_copy(input: &[u8], name: &str) -> Self {
```
ErisianArchitect, line 72: This function does not need to be unsafe.

src/memory_buffer.rs
```diff
@@ -106,16 +101,11 @@ impl<'a> MemoryBuffer<'a> {
...
-    pub fn create_from_memory_range(input: &'a [u8], name: &str) -> Self {
-        assert_eq!(
-            input[input.len() - 1],
-            b'\0',
-            "input byte slice must terminate with a nul byte"
-        );
-
+    /// The `input` byte slice must confidently terminate with a nul byte (`\0`),
+    /// otherwise LLVM will experience out-of-bounds undefined behavior/segfaults.
+    pub unsafe fn create_from_memory_range(input: &'a [u8], name: &str) -> Self {
```
ErisianArchitect, line 108: This also doesn't need to be marked as unsafe.

src/memory_buffer.rs
```diff
@@ -124,7 +114,7 @@ impl<'a> MemoryBuffer<'a> {
                 // decremented since technically the nul byte is one past the end of the input.
                 input.len() - 1,
                 name_c_string.as_ptr(),
-                false as i32, // guaranteed to have nul-terminator by CStr
+                false as i32, // required to be false since LLVM expects already nul-terminated arrays
```
ErisianArchitect, line 117: I don't know why you changed the comment here. Pointless change.

ErisianArchitect (commented):
I do not approve of these changes. And I don't think any of your PRs should be approved. They are all vibe coded, and I'd be willing to bet that you couldn't explain the rationale behind these changes without using AI.

TheDan64:
The PassManager cleanup might be a good one. But it's deprecated anyway and slated for removal. The rest of the changes don't make much sense. Rather than fixing the memory safety, you just opted to mark it unsafe. Closing as probable vibe code

Issue #320, opened 2022-04-28
When building with LTO, MCJIT is not linked in + segfault


DCNick3:
**Describe the Bug**

When building a project using inkwell with LTO, the rust compiler seems to optimize the JIT out and attempt to use JIT fails with error message "JIT has not been linked in".

This seems to happen because rustc is able to guarantee that `LLVMLinkInMCJIT` is never called and does not link the whole MCJIT library due to it not being used. Due to this a global constructor registering jit as an execution engine does not run and JIT stops being available.

This is easily fixed by adding a call to `inkwell::execution_engine::ExecutionEngine::link_in_mc_jit`, which prevents the omission of the MCJIT library

Another unfortunate thing is that panic due to a missing JIT seems to lead to a segfault somewhere in destructors, preventing the aforementioned message from being printed. Didn't investigate any further though

**To Reproduce**

You can use [this repo](https://github.com/DCNick3/inkwell-repro)

```
git clone https://github.com/DCNick3/inkwell-repro
cd inkwell-repro
cargo r --release
```

**Expected Behavior**
JIT is linked in and usable

Also, would be nice if the sefault didn't happen

**LLVM Version (please complete the following information):**
 - LLVM Version: 13.0.0
 - Inkwell Branch Used: master (currently at commit bff378bee02bcbb5bed35f47e9ed69e6642e9188)

**Desktop (please complete the following information):**
 - OS: Arch Linux rolling release

TheDan64:
Hmm yeah maybe the EE constructor would be a good place to call it?

DCNick3:
Yeah, probably the best place. Seems that there is `link_in_interpreter` which has similar purpose, but for interpreter.

thetayloredman:
Reproduced on 0.8.0. The panic appears to be because the flag for the JIT that "owns" a given Module is not properly set, resulting in what appears to be a double free.

TheDan64:
Fixed in #681

PR #681, opened 2026-04-10
fix missing LinkIn* calls

foxidokun:
## Description

LLVMLinkIn* are basically no-op, but they are required for proper llvm module registration via static objects. Without calling them, there is no guarantee that LLVM module will exist.

In my case, enabling `lto=true` eliminated them completely, leading to a strange error "JIT not linked in" and hours of debugging how I managed to unlink it :) While functions ExecutionEngine::link_in_* were provided, they were unsearchable by reasonable keywords and were missing even fom Inkwell's own examples.

## Related Issue

After I understood that the issue was related to LTO, I found #320. 

## How This Has Been Tested

The change is minimal, so no extensive testing required. I was able to build my pretty big compiler with this change, as well as Inkwell examples.

Also cargo test on llvm 20 & 22

## Option\<Breaking Changes\>

This patch hides ExecutionEngine::link_in_* functions from the user. I just personally don't like exposing such a utility, and it is unneeded after this patch.

## Checklist

### DIFF
--- a/src/execution_engine.rs
+++ b/src/execution_engine.rs
@@ -124,15 +124,15 @@ impl<'ctx> ExecutionEngine<'ctx> {
         **self.execution_engine_rc()
     }
 
-    /// This function probably doesn't need to be called, but is here due to
-    /// linking(?) requirements. Bad things happen if we don't provide it.
-    pub fn link_in_mc_jit() {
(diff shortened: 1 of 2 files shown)


TheDan64 (approved):
Thanks!

PR #683, opened 2026-04-15
Fixes get_current_debug_location error.

### LINKED ISSUE #674: `Builder::unset_current_debug_location` does not work
**Describe the Bug**
Immediately after a call to `Builder::unset_current_debug_location` on a fresh builder, `Builder::get_current_debug_location` returns Some. On a related note, `get_current_debug_location` returns Some immediately after the builder is constructed, which is unexpected if not necessarily a bug.

**To Reproduce**
```

fn main () {
    let c = Context::create();
    let b = c.create_builder();

    assert!(b.get_current_debug_location().is_some()); // unexpectedly passes
    b.unset_current_debug_location();
    assert!(b.get_current_debug_location().is_none()); // unexpectedly fails
}
```

**Expected Behavior**
`get_current_debug_location` should return None after a call to `unset_current_debug_location`.

 - LLVM Version: 21.18
 - Inkwell Branch Used: llvm21-1, Inkwell v0.8.0

 - OS: MacOS Tahoe 26.4

ErisianArchitect:
## Description

`Builder::get_current_debug_location` was using `LLVMGetCurrentDebugLocation`, which is deprecated since LLVM 9.0 (and is available in LLVM 8.0).

This changes the usage of `LLVMGetCurrentDebugLocation` to `LLVMGetCurrentDebugLocation2`.

## Related Issue

Fixes #674

## How This Has Been Tested

I've added a new test in `test_builder.rs` called `test_current_debug_location`, it simply tests for the specific case illustrated in the issue. Before my change, the test failed. After my change, the test passes.

## Checklist


ErisianArchitect:
If you want me to add the is_none check before the explicit unset, I can, but I don't think that it's necessary since the fix was to stop using the deprecated function (which defines returning null as an error, if I remember correctly).

TheDan64:
We should add it for thoroughness

ErisianArchitect:
Done.

TheDan64:
Sorry I missed it earlier, but we should also set and verify unset works

TheDan64:
and get

ErisianArchitect:
I'm not entirely sure of the API around the debug locations. I'll look into it in a little while.

ErisianArchitect:
Alright. I was going to ask if we didn't have to do that part, but I managed to find some other test code that works with debug locations, so I just yoinked it and added that code into the new test.

**Edit**: Oh, I should be more clear. I changed the code up, I didn't just copy it verbatim.

TheDan64 (approved):
Thank you!

sbourdeauducq:
Would it be possible to publish a minor release with this bugfix?

TheDan64:
Yes, we can plan to do one soon

PR #685, opened 2026-04-16
Fixes bug in write_bitcode_to_file in #6.

### LINKED ISSUE #6: write_bitcode_to_file has inconsistent success
Locally, it seems to pass most of the time but will occasionally fail with an OS error saying the file descriptor is bad. (Ubuntu at least). Seems to fail more often on Travis (at least one job per build) so I've commented out the test for now and made the method private. It would be great to have someone more knowledgeable about filesystems (and possibly LLVM) take a look. It could possibly be a LLVM bug since the wrapper is really straightforward and thin. `write_bitcode_to_path` doesn't have this problem (different FFI function call, though)

ErisianArchitect:
## Description

write_bitcode_to_file would fail if `should_close` was true because it would double-close the file descriptor.

## Related Issue

Closes #6 

## How This Has Been Tested

I also fixed the test in test_module.rs, and it passes on 21-1 (didn't test any others because I don't have it setup for other versions, guess I'll have to see if it fails in the CI.

## Option\<Breaking Changes\>

Removes `should_close` argument from `write_bitcode_to_file` because it isn't applicable in safe Rust as Rust will automatically close the File unless you use `std::mem::forget`. As far as I'm aware, there's no way to prevent Rust from automatically closing a file besides with `into_raw_fd`, which can't be called on a File reference.

## Checklist

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -815,7 +815,7 @@ impl<'ctx> Module<'ctx> {
         {
             use llvm_sys::bit_writer::LLVMWriteBitcodeToFD;
             use std::os::unix::io::AsRawFd;
-    
+
             // REVIEW: as_raw_fd docs suggest it only works in *nix
             // Also, should_close should maybe be hardcoded to true?
             unsafe {
--- a/src/module.rs
+++ b/src/module.rs
@@ -809,25 +809,24 @@ impl<'ctx> Module<'ctx> {
 
     // See GH issue #6
     /// `write_bitcode_to_path` should be preferred over this method, as it does not work on all operating systems.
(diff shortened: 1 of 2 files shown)


ErisianArchitect:
Oh, if you're wondering why I changed the ordering of the cfgs in write_bitcode_to_file, it's because I thought it was more readable with the `not` branch first. Aside from that, I really just think the whole function should be cfg'd out for non-unix systems.

TheDan64 (approved):
Thanks!

ErisianArchitect:
I bet you're glad that this ancient issue was finally resolved. Haha.

TheDan64:
yeah hahaha

PR #483, opened 2024-03-14
WIP: Support LLVM 18

xermicus:
## Description

Support [LLVM 18](https://releases.llvm.org/18.1.0/docs/ReleaseNotes.html#id26).

## Related Issue

#482

## Checklist

- [x] Introduce feature flags 
- [ ] Wait out
  - [x] next `llvm-sys` release supporting LLVM 18
  - [ ] LLVM 18 binary releases to run tests on CI
- [x] Guard removed API methods and make test suite pass
- [ ] Implement added functionality
  - [x] Getting and setting tail, musttail, and notail attributes on call instructions
  - [x] Implmented a safe wrapper around LLVMCreateTargetMachineWithOptions
  - [x] New flags: LLVMGetNNeg, LLVMSetNNeg, LLVMGetIsDisjoint, LLVMSetIsDisjoint, LLVMGetFastMathFlags, LLVMSetFastMathFlags, LLVMCanValueUseFastMathFlags
  - [ ] ~~Operand Bundles~~ We don't need this right now, can I leave this open for a future PR?

- [x] I have read the [Contributing Guide](https://github.com/TheDan64/inkwell/blob/master/.github/CONTRIBUTING.md)

PR #687, opened 2026-04-18
Swaps raw pointer types with NonNull for better safety and niches

### LINKED ISSUE #686: Swapping pointer values that should never be null with NonNull variants.
I was reading some code, and I noticed that `PointerValue` wraps `Value`, which wraps `LLVMValueRef`, which is a type alias for `*mut llvm_sys::LLVMValue`. So I was thinking that perhaps it would make sense to instead use `NonNull<LLVMValue>` (or other relevant pointer type) in these places where non-nullable pointers are used.

Then these types could have `repr(transparent)`, which would guarantee that they have the niche optimization that NonNull has, allowing `T` to be the same size as `Option<T>`, which is a very useful memory optimization.

I'm going to get started on this PR today. I just wanted to make an issue for it since there's a rule that pull requests need to be related to an issue.

ErisianArchitect:
## Description

Many of the types used `LLVM*Ref` types, which are type aliases for `*mut LLVM*`. These pointers are null-checked before creating new instances of these types, so I thought it would make sense to use `NonNull` instead so that there would be better safety guarantees, and it would enable these types to have at least one niche, allowing for this condition: `size_of::<T>() == size_of::<Option<T>>() && align_of::<T>() == align_of::<Option<T>>()`.

## Related Issue

Closes #686 

## How This Has Been Tested

I finally figured out clippy, so I have a bash function that I run that runs clippy, runs all the tests, then runs cargo fmt. So all checks should hopefully pass on the first try. Fingers crossed.

It's quite a big update. It was extremely repetitive and tedious, but I think I got everything.

## Option\<Breaking Changes\>

I'm pretty sure there are no breaking changes.

## Checklist

### DIFF
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -1,15 +1,20 @@
 //! `Attribute`s are optional modifiers to functions, function parameters, and return types.
 
-use llvm_sys::core::{
-    LLVMGetEnumAttributeKind, LLVMGetEnumAttributeKindForName, LLVMGetEnumAttributeValue, LLVMGetLastEnumAttributeKind,
-    LLVMGetStringAttributeKind, LLVMGetStringAttributeValue, LLVMIsEnumAttribute, LLVMIsStringAttribute,
-};
 #[llvm_versions(12..)]
 use llvm_sys::core::{LLVMGetTypeAttributeValue, LLVMIsTypeAttribute};
 use llvm_sys::prelude::LLVMAttributeRef;
+use llvm_sys::{
+    LLVMOpaqueAttributeRef,
+    core::{
+        LLVMGetEnumAttributeKind, LLVMGetEnumAttributeKindForName, LLVMGetEnumAttributeValue,
+        LLVMGetLastEnumAttributeKind, LLVMGetStringAttributeKind, LLVMGetStringAttributeValue, LLVMIsEnumAttribute,
+        LLVMIsStringAttribute,
+    },
+};
 
-use std::ffi::CStr;
+use std::{ffi::CStr, ptr::NonNull};
 
+use crate::support::assert_niche;
 #[llvm_versions(12..)]
 use crate::types::AnyTypeEnum;
 
@@ -17,17 +22,19 @@ use crate::types::AnyTypeEnum;
 // REVIEW: Should Attributes have a 'ctx lifetime?
 /// Functions, function parameters, and return types can have `Attribute`s to indicate
 /// how they should be treated by optimizations and code generation.
+#[repr(transparent)]
 #[derive(Clone, Copy)]
 pub struct Attribute {
-    pub(crate) attribute: LLVMAttributeRef,
+    pub(crate) attribute: NonNull<LLVMOpaqueAttributeRef>,
 }
+const _: () = assert_niche::<Attribute>();
 
 impl std::fmt::Debug for Attribute {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         if self.is_string() {
             return f
                 .debug_struct("Attribute::String")
-                .field("ptr", &self.attribute)
+                .field("ptr", &self.as_mut_ptr())
                 .field("kind_id", &self.get_string_kind_id())
                 .field("value", &self.get_string_value())
                 .finish();
@@ -36,7 +43,7 @@ impl std::fmt::Debug for Attribute {
         if self.is_enum() {
             return f
                 .debug_struct("Attribute::Enum")
-                .field("ptr", &self.attribute)
+                .field("ptr", &self.as_mut_ptr())
                 .field("kind_id", &self.get_enum_kind_id())
                 .field("value", &self.get_enum_value())
                 .finish();
@@ -45,7 +52,7 @@ impl std::fmt::Debug for Attribute {
         if self.is_type() {
             return f
                 .debug_struct("Attribute::Type")
-                .field("ptr", &self.attribute)
+                .field("ptr", &self.as_mut_ptr())
                 .field("kind_id", &self.get_enum_kind_id())
                 .field("value", &self.get_type_value())
                 .finish();
@@ -84,16 +91,22 @@ impl PartialEq<Self> for Attribute {
 }
 
 impl Attribute {
-    /// Creates a new `Attribute` from a raw pointer.
+    /// Creates a new [Attribute] from a raw pointer.
+    ///
+    /// # Safety
+    ///
+    /// `attribute` must be non-null and point to a valid value.
     pub unsafe fn new(attribute: LLVMAttributeRef) -> Self {
         debug_assert!(!attribute.is_null());
 
-        Attribute { attribute }
+        Attribute {
+            attribute: unsafe { NonNull::new_unchecked(attribute) },
+        }
     }
 
     /// Acquires the underlying raw pointer belonging to this `Attribute` type.
     pub fn as_mut_ptr(&self) -> LLVMAttributeRef {
-        self.attribute
+        self.attribute.as_ptr()
     }
 
     /// Determines whether or not an `Attribute` is an enum. This method will
@@ -111,7 +124,7 @@ impl Attribute {
     /// assert!(enum_attribute.is_enum());
     /// ```
     pub fn is_enum(self) -> bool {
-        unsafe { LLVMIsEnumAttribute(self.attribute) == 1 }
+        unsafe { LLVMIsEnumAttribute(self.as_mut_ptr()) == 1 }
     }
 
     /// Determines whether or not an `Attribute` is a string. This method will
@@ -129,7 +142,7 @@ impl Attribute {
     /// assert!(string_attribute.is_string());
     /// ```
     pub fn is_string(self) -> bool {
-        unsafe { LLVMIsStringAttribute(self.attribute) == 1 }
+        unsafe { LLVMIsStringAttribute(self.as_mut_ptr()) == 1 }
     }
 
     /// Determines whether or not an `Attribute` is a type attribute. This method will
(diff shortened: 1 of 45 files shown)


ErisianArchitect:
Third times the charm, I guess.

ErisianArchitect:
Oh, I forgot to mention, I also added a couple functions to support.

`const_assert` and `assert_niche`.

`assert_niche` is used to assert that types have a niche (same size and alignment as `Option<T>`).

`const_assert` is just used by `assert_niche`, but it could be used anywhere that a compile time assertions needs to be done.

src/values/operand_bundle.rs
```diff
@@ -1,21 +1,24 @@
...
-use std::cell::Cell;
 use std::ffi::CStr;
 use std::marker::PhantomData;
+use std::ptr::NonNull;
 
 /// One of an instruction's operand bundles.
+#[repr(transparent)]
 #[derive(Debug)]
 pub struct OperandBundle<'ctx> {
-    bundle: Cell<LLVMOperandBundleRef>,
```
ErisianArchitect, line 16: As far as I could tell, bundle did not need to be a `Cell`, so I changed that too.

src/targets.rs
```diff
@@ -1481,15 +1493,19 @@ impl TargetMachineOptions {
     /// - The only way to access it is via this private method.
     /// - Disposal is taken care of automatically in `Drop::drop`.
     unsafe fn inner(&mut self) -> LLVMTargetMachineOptionsRef {
-        unsafe { *self.0.get_or_insert_with(|| LLVMCreateTargetMachineOptions()) }
+        unsafe {
+            self.0
+                .get_or_insert_with(|| NonNull::new_unchecked(LLVMCreateTargetMachineOptions()))
```
ErisianArchitect, line 1498: This is honestly a little bit weird, but it works.

ErisianArchitect (commented):
I don't know if it counts for much, but I read through the entire codebase at least twice while working on this PR, and have now read through all of my changes.

ErisianArchitect:
Oh, and one last note: I didn't apply this optimization to `passes.rs`. It's not that I couldn't have for some reason, it's just that I'd already read through all the code twice by the time I got to that, and I didn't want to go back and read it again (because of all the conditionally compiled blocks that don't show errors in my environment). I did the best that I could. Considering all the checks passed, I'm hoping that I didn't make any mistakes. I hope this optimization is worth the time I just spent on this, because it was a lot more work than I thought it was going to be.

TheDan64:
Looks cool, it'll take me a while to review it all

src/support/mod.rs
```diff
@@ -105,14 +109,14 @@ impl Drop for LLVMString {
 #[derive(Eq)]
 pub(crate) enum LLVMStringOrRaw {
     Owned(LLVMString),
-    Borrowed(*const c_char),
+    Borrowed(NonNull<c_char>),
```
TheDan64, line 112: If LLVMString and NonNull are both niche, doesn't that mean `LLVMStringOrRaw` can be niche as well? Or does the enum miss out on it?

src/support/mod.rs
```diff
@@ -233,6 +237,45 @@ pub(crate) fn to_c_str(mut s: &str) -> Cow<'_, CStr> {
...
+/// struct Foo {
+///     bar: &'static (),
+/// }
+/// const _: () = const_assert(size_of::<Foo>() == size_of::<&'static ()>());
+/// ```
+#[track_caller]
+#[inline(always)]
+pub(crate) const fn const_assert(assertion: bool) {
+    if !assertion {
+        panic!("Assertion Failed");
```
TheDan64, line 256: Is it possible to be able to take a static string? Would be nice if the panic had more specific details

ErisianArchitect: Unfortunately, no, not at compile time. `panic` expects a format argument, and format args, but it can't perform formatting at compile time.

ErisianArchitect: The best recommendation is to include a comment with the assertion. It's marked with `#[track_caller]`, so the error will appear at the location of the assertion.

src/support/mod.rs
ErisianArchitect: No, unfortunately it misses out on the niche since it needs an extra value for the discriminant. It can't say `this variant and this value` or `that variant and that value` in a flattened way. But LLVMStringOrRaw will have plenty of niches anyway, 2^63, I believe.

ErisianArchitect:
> Looks cool, it'll take me a while to review it all

Yeah, sorry about that, haha. I really did not expect this change to affect this many lines.

src/targets.rs
TheDan64: Do you think The struct should just be initialized with LLVMCreateTargetMachineOptions? Or does it being Option make sense?

src/data_layout.rs
```diff
@@ -33,8 +33,8 @@ impl DataLayout {
 
     pub fn as_ptr(&self) -> *const ::libc::c_char {
         match self.data_layout {
-            LLVMStringOrRaw::Owned(ref llvm_string) => llvm_string.ptr,
-            LLVMStringOrRaw::Borrowed(ptr) => ptr,
+            LLVMStringOrRaw::Owned(ref llvm_string) => llvm_string.ptr.as_ptr(),
+            LLVMStringOrRaw::Borrowed(ptr) => ptr.as_ptr().cast_const(),
```
TheDan64, line 37: "cast_const" gives me C++ ptsd 😂

src/targets.rs
ErisianArchitect: I'm not sure. I can look into it. I'm not sure why it's an `Option` in the first place, I didn't want to focus on anything besides what I was working on. I didn't want to give you too much to review. I think the reason it's an `Option` is because it's nullable. But I'll find out for sure and get back to you.

src/data_layout.rs
ErisianArchitect: Yeah, I didn't like having to do that either. But `NonNull` only gives you a `mut` pointer, so what can ya do, right?

src/targets.rs
ErisianArchitect:
```rs
/// LLVM target machine options provide another way to create target machines,
/// used with [Target::create_target_machine_from_options].
///
/// The option structure exposes an additional setting (i.e., the target ABI)
/// and provides default values for unspecified settings.
#[llvm_versions(18..)]
#[derive(Default, Debug)]
pub struct TargetMachineOptions(Option<NonNull<LLVMOpaqueTargetMachineOptions>>);
```

I guess I didn't have to look far after all. I'm not entirely sure what this documentation is saying, but I'll figure it out. Maybe it doesn't need to be an `Option`, maybe it does. But I think it's a good idea to figure out that reason so we can include that information with the documentation.

ErisianArchitect: Okay, I'm back again. From the looks of it, it seems that the reason that it's an `Option` is for some sort of lazy initialization. I'm not sure why...

[After some more digging while writing this comment]

I narrowed it down to this commit: https://github.com/TheDan64/inkwell/pull/483/changes#diff-05702e78430359853e5b38404035fa394540368bee185a2cdc936a62b9cd65b0R1493

Unfortunately, he doesn't explain his changes.

But looking at the code, there doesn't appear to be any reason that `TargetMachineOptions` *needs* lazy initialization. I can remove it if you want.

ErisianArchitect: I asked the robot just for the hell of it, and the robot claims that `LLVMTargetMachineOptionsRef` should never be null, so maybe it doesn't represent a default value of any sort. I'll look deeper, because I don't trust anything the robot says.

TheDan64: Could you please open up a follow up issue? You're right that this PR is already big enough as is. Was mostly curious, it doesnt need immediate attention

src/attributes.rs
```diff
@@ -45,7 +52,7 @@ impl std::fmt::Debug for Attribute {
         if self.is_type() {
             return f
                 .debug_struct("Attribute::Type")
-                .field("ptr", &self.attribute)
+                .field("ptr", &self.as_mut_ptr())
```
TheDan64, line 55: I don't mind the new field, but elsewhere you called it `address`. We should pick one and use it consistently throughout

src/builder.rs
```diff
@@ -3663,20 +3875,17 @@ impl<'ctx> Builder<'ctx> {
...
         }
     }
 
     /// Get the debug info source location of the instruction currently pointed at by the builder,
     /// if available.
     pub fn get_current_debug_location(&self) -> Option<DILocation<'ctx>> {
         use llvm_sys::core::LLVMGetCurrentDebugLocation2;
-        let metadata_ref = unsafe { LLVMGetCurrentDebugLocation2(self.builder) };
-        if metadata_ref.is_null() {
-            return None;
```
TheDan64, line 3676: Is there a reason for the removed is null check? Now it never returns None, even if null

ErisianArchitect: Because the null check was moved into `NonNull::new`. There's a `?` after the call that causes the `None` result to be propagated.

src/attributes.rs
ErisianArchitect: I'm not sure what you mean. `attribute` was the name of the prior field. I just didn't change it. I don't think that I changed the name of fields anywhere.

src/builder.rs
TheDan64: Ah I missed that, neat!

src/attributes.rs
TheDan64: ah okay, sorry - I was reviewing in split view and misread

src/targets.rs
ErisianArchitect: Sure, I can do that.

ErisianArchitect:
Any update on this? I want to get started on Orc2 soon because I'm going to need it for my upcoming project. Feel free to reject this PR if you want, I won't be offended. I just want this one to be done before I move on to Orc2.

TheDan64:
Sorry, I forgot to finish reviewing. Will get to it soon

TheDan64 (approved):

ErisianArchitect:
Thanks! I'll start work on Orc2 soon.

PR #689, opened 2026-04-22
RFC: Encode zero kind_id semantic in return type

foxidokun:
## Description

LLVM returns id=0 for unknown attribute ids. This patch encode this semantic via zero-cost Option<NonZeroU32>, which will provide .unwrap() and other convenient methods.

## Reasoning

I'd like to have convenient .unwrap(), .ok_or() and other methods to validate that attribute exist. However, this patch is pure cosmetics, so feel free to drop it

## Considerations

- NonZeroU32 has kinda ugly conversion API
- Should we refactor functions that accept kind_id to NonZeroU32 too? Like `Context::create_enum_attribute`

## Checklist

### DIFF
--- a/src/attributes.rs
+++ b/src/attributes.rs
@@ -9,6 +9,7 @@ use llvm_sys::core::{LLVMGetTypeAttributeValue, LLVMIsTypeAttribute};
 use llvm_sys::prelude::LLVMAttributeRef;
 
 use std::ffi::CStr;
+use std::num::NonZeroU32;
 
 #[llvm_versions(12..)]
 use crate::types::AnyTypeEnum;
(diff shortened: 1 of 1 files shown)


TheDan64 (approved):
Change looks good but tests need to be updated to pass the suite

ErisianArchitect:
This is a breaking change, is it not? What's the need for Option<Nonzerou32> here?

TheDan64:
That's a good point that it's a breaking change and should have been documented as such. But we do make breaking changes from time to time since we don't have a stable API due to LLVM version changes. So I'm not necessarily opposed to the change... I do think that Option makes it clearer that the attribute doesn't exist rather than relying on implicit C semantics which an inkwell user may not be familiar with. Perhaps it should just be `Option<u32>` for simplicity?

TheDan64:
It seems like it'd be nicer to make the whole thing a bit more typesafe, with `KindId` transparent wrappers, and `const` attribute values for reference, then you could get one dynamically, or use the const and provide it to `create_enum_attribute` as OP mentioned

foxidokun:
> This is a breaking change, is it not?

Yes, it is, but not very big. Wasn't documented because it's just RFC

> Option<u32> for simplicity

Yes, this will be more usable, but will lose"zero-costness" (memory layout optimizations), but it's not a significant loss.

> the whole thing a bit more typesafe

I agree, I'll try to play with this idea of wrappers. TBH I'd really like to distinguish Attribute::get_named_enum_kind and Context::get_kind_id, they has similar names and  sometimes passing one's id to another works (sad experience...)

PR #696, opened 2026-06-15
Fixes #695: Underflow in create_from_memory_range.

### LINKED ISSUE #695: Integer underflow error in #655
[This pull-request has a bug.](https://github.com/TheDan64/inkwell/pull/655)

```rust
    pub fn create_from_memory_range(input: &'a [u8], name: &str) -> Self {
        assert_eq!(
// ----------------- BUG IS HERE ---------------------------
            input[input.len() - 1],
            b'\0',
            "input byte slice must terminate with a nul byte"
        );

        let name_c_string = to_c_str(name);

        let memory_buffer = unsafe {
            LLVMCreateMemoryBufferWithMemoryRange(
                input.as_ptr() as *const libc::c_char,
                input.len() - 1,
                name_c_string.as_ptr(),
                false as i32, // guaranteed to have nul-terminator by CStr
            )
        };

        unsafe { MemoryBuffer::new(memory_buffer) }
    }
```

ErisianArchitect:
## Description

There was an integer underflow in create_from_memory_range in `memory_buffer.rs`. While fixing the underflow error, I dug into the actual source code of LLVM to see what was going on, and discovered that there's absolutely no need for the input data to be null-terminated, so I removed that requirement entirely and provided my sources.

## Related Issue

#695 

## How This Has Been Tested

All tests, even doc tests passed on x86_64 Linux with llvm21-1.

```bash
function prepare-pr() {
    if [[ $1 != "--no-format" ]]; then
        cargo fmt || {
            local ret_code=$?
            failure "cargo fmt failed."
            return $ret_code
        }
    fi
    cargo clippy --no-default-features --features "llvm21-1 target-x86" || {
        local ret_code=$?
        failure "cargo clippy failed."
        return $ret_code
    }
    cargo test --no-default-features --features "llvm21-1 target-x86" --release || {
        local ret_code=$?
        failure "cargo test failed."
        return $ret_code
    }
    cargo test --no-default-features --features "llvm21-1 target-x86" --doc --release || {
        local ret_code=$?
        failure "cargo test --doc failed."
        return $ret_code
    }
    success "Ready for PR"
}
```

## Checklist

### DIFF
--- a/src/memory_buffer.rs
+++ b/src/memory_buffer.rs
@@ -113,7 +113,7 @@ impl<'a> MemoryBuffer<'a> {
                 input.len(),
                 name_c_string.as_ptr(),
                 /* If this is `true`, LLVM will expect a null-terminator. If it is false, the null-terminator is not
-                 * necesary.
+                 * necessary.
                  * https://llvm.org/doxygen/IR_2Core_8cpp_source.html#l04679
                  * ```
                  * LLVMMemoryBufferRef LLVMCreateMemoryBufferWithMemoryRange(
--- a/src/memory_buffer.rs
+++ b/src/memory_buffer.rs
@@ -66,25 +66,14 @@ impl MemoryBuffer<'static> {
         unsafe { Ok(Self::new(memory_buffer)) }
     }
 
-    /// Create a memory buffer copied from a byte slice with a trailing nul byte.
-    ///
-    /// # Panics
-    ///
-    /// Panics if the input byte slice does not terminate with a nul byte.
+    /// Create a memory buffer copied from a byte slice.
     pub fn create_from_memory_range_copy(input: &[u8], name: &str) -> Self {
-        assert_eq!(
-            input[input.len() - 1],
-            b'\0',
-            "input byte slice must terminate with a nul byte"
-        );
-
         let name_c_string = to_c_str(name);
 
         let memory_buffer = unsafe {
             LLVMCreateMemoryBufferWithMemoryRangeCopy(
                 input.as_ptr() as *const libc::c_char,
-                // decremented since technically the nul byte is one past the end of the input.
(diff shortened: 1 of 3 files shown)


ErisianArchitect:
Sorry it took me so long to get around to this, I was pretty busy.

src/memory_buffer.rs
```diff
@@ -107,27 +96,44 @@ impl<'a> MemoryBuffer<'a> {
...
+             * ```
+             */
             LLVMCreateMemoryBufferWithMemoryRange(
                 input.as_ptr() as *const libc::c_char,
-                // decremented since technically the nul byte is one past the end of the input.
-                input.len() - 1,
+                input.len(),
                 name_c_string.as_ptr(),
-                false as i32, // guaranteed to have nul-terminator by CStr
+                /* If this is `true`, LLVM will expect a null-terminator. If it is false, the null-terminator is not
```
TheDan64, line 115: Doesn't this mean we can get rid of `to_c_str` and just use the input name str?

TheDan64: if we set it to true

TheDan64: or is the the byte buffer and not the string

ErisianArchitect: No, that's only for input_data, not for the name string.

TheDan64: jk I kept reading lol

TheDan64 (approved):
Good catch!

src/memory_buffer.rs
ErisianArchitect: If it's not clear what the comment is for, I can make it more clear.

ErisianArchitect:
My spidey senses were tingling.

Issue #355, opened 2022-09-22
ContextRef in debug builder


ghaith:
**Describe the Bug**
I saw a problem when using the `DIGlobalVariableExpression::as_metadata_value` which still uses context, meaning I cannot just pass the context ref the module holds.
I assume the references to context can all be changed to ContextRef, but it might still break some code until this is done. 

If ContextRef is meant to be a reference to the Context that the module holds, it might be meaningful for it to implement Deref as well.

TheDan64:
ContextRef can't deref to Context or vice versa, since it doesn't actually use one. Will have to think about this some more

TheDan64:
Been thinking about this and the only approach I've really got is a new context trait that is implemented for both Context and ContextRef

TheDan64:
Either that or like `AsRef<PrivateContextType>`

TheDan64:
Working on a fix for this. Going with the first approach

Issue #335, opened 2022-06-27
Higher Kinded Types?


JCBurnside:
**Is your feature request related to a problem? Please describe.**
trying to make a functional language so I need higher kinded types (eg `int32 -> int32 -> int32`) thus I need to create a function that has a return type of another function.  as well why can't I simply create pointer and/or function off an `AnyTypeEnum`?

**Describe the solution you'd like**
Allow chaining of the `fn_type` method on `AnyTypeEnum`.  simply implemented as a match on self that would allow 

**Describe possible drawbacks to your solution**
I don't think it can hinder much?

**Describe alternatives you've considered**
not really any work arounds i've found.

JCBurnside:
Just realized i need to make it return a function pointer in order to pull it off.  Still would like to see the latter part of the suggestion.

TheDan64:
I'm not sure that it makes sense to add to AnyTimeEnum since not all types (now or in the future) may be valid. But perhaps we could make the method fallible or something. hmm

EpicVon2468:
I know this is an old issue, but I'd like to mention I've personally ran into this "problem" and ended up writing a wrapper trait to get around it.

<img width="939" height="724" alt="Image" src="https://github.com/user-attachments/assets/c0e625dc-6a54-47b3-9353-4e19656f0309" />

Would appreciate _something_ of this nature built-in (if possible)

TheDan64:
I would be open to an impl on AnyTypeEnum which is fallible for the case of an invalid variant

EpicVon2468:
For a fallible implementation, would it be better to make the function return `Option<FunctionType<'ctx>>`, or just panic?

~I'm not familiar with best practice for libraries~

TheDan64:
Result would be good, and a simple struct which denotes the error. Something like

```rust
use thiserror::Error;

#[derive(Error)]
#[error("This variant cannot be used with fn_type")]
struct NotFnTypeable;

fn fn_type(...) -> Result<FunctionType, NotFnTypeable> { ... }
```

EpicVon2468:
> #[error("This variant cannot be used with fn_type")]
> struct NotFnTypeable;

Should the error also contain the name / variant which was found, or?
I'm working on an implementation of this right now.

EpicVon2468:
(follow-up: should a similar wrapper be implemented for `array_type` ?)

TheDan64:
Sure, you can make the error struct more generic and use it for both

[...]

PR #697, opened 2026-06-18
Implementation for #355

EpicVon2468:
## Description

Implementation for the `AnyTypeEnum::fn_type` and `AnyTypeEnum::array_type` functions as described in #335

## Related Issue

https://github.com/TheDan64/inkwell/issues/335

## How This Has Been Tested

Works on my machine (Ubuntu 26.10, `raptorlake`).
There (probably) can't really be any issues with this breaking stuff.

## Checklist

### DIFF
--- a/src/types/enums.rs
+++ b/src/types/enums.rs
@@ -759,3 +759,120 @@ impl Display for BasicMetadataTypeEnum<'_> {
         write!(f, "{}", self.print_to_string())
     }
 }
+
+#[derive(thiserror::Error, Debug)]
+#[error("This variant of `AnyTypeEnum` cannot be used with [TODO: list function]")]
+pub struct InvalidVariant;
+
+impl<'ctx> AnyTypeEnum<'ctx> {
+    /// Creates a [`FunctionType`] with [`Self`] for its return type.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`InvalidVariant`] if [`Self`] cannot be used to create a function type.
+    ///
+    /// In this case, the only variant which cannot be used to create a function type is [`Self::FunctionType`].
+    ///
+    /// # Examples
+    ///
+    /// Using a valid variant:
+    ///
+    /// ```no_run
+    /// use inkwell::context::Context;
+    /// use inkwell::types::AnyTypeEnum;
+    ///
+    /// let context = Context::create();
+    /// let i8_type = context.i8_type();
+    /// let wrapped_type = AnyTypeEnum::from(i8_type);
+    /// let result = wrapped_type.fn_type(&[], false);
+    ///
+    /// // `AnyTypeEnum::IntType` can be used to create a function type.
+    /// assert!(result.is_ok());
+    /// let fn_type = result.unwrap();
+    /// ```
+    ///
+    /// Using an invalid variant:
+    ///
+    /// ```no_run
+    /// use inkwell::context::Context;
+    /// use inkwell::types::AnyTypeEnum;
+    ///
+    /// let context = Context::create();
+    /// let fn_type = context.i8_type().fn_type(&[], false);
+    /// let wrapped_type = AnyTypeEnum::from(fn_type);
+    /// let result = wrapped_type.fn_type(&[], false);
+    ///
+    /// // `AnyTypeEnum::FunctionType` cannot be used to create a function type.
+    /// assert!(result.is_err());
+    /// ```
+    pub fn fn_type(
+        self,
+        param_types: &[BasicMetadataTypeEnum<'ctx>],
+        is_var_args: bool,
+    ) -> Result<FunctionType<'ctx>, InvalidVariant> {
+        match self {
+            AnyTypeEnum::ArrayType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            AnyTypeEnum::FloatType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            AnyTypeEnum::IntType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            AnyTypeEnum::PointerType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            AnyTypeEnum::StructType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            AnyTypeEnum::VectorType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            AnyTypeEnum::ScalableVectorType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            AnyTypeEnum::VoidType(inner) => Ok(inner.fn_type(param_types, is_var_args)),
+            _ => Err(InvalidVariant),
+        }
+    }
+
+    /// Creates an [`ArrayType`] with [`Self`] for its element type.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`InvalidVariant`] if [`Self`] cannot be used to create an array type.
+    ///
+    /// In this case, the only variants which cannot be used to create an array type are [`Self::FunctionType`] and [`Self::VoidType`].
+    ///
+    /// # Examples
+    ///
+    /// Using a valid variant:
+    ///
+    /// ```no_run
+    /// use inkwell::context::Context;
+    /// use inkwell::types::AnyTypeEnum;
+    ///
+    /// let context = Context::create();
+    /// let i8_type = context.i8_type();
+    /// let wrapped_type = AnyTypeEnum::from(i8_type);
+    /// let result = wrapped_type.array_type(5);
+    ///
+    /// // `AnyTypeEnum::IntType` can be used to create an array type.
+    /// assert!(result.is_ok());
+    /// let array_type = result.unwrap();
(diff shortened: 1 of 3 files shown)


src/types/enums.rs
```diff
@@ -759,3 +759,134 @@ impl Display for BasicMetadataTypeEnum<'_> {
         write!(f, "{}", self.print_to_string())
     }
 }
+
+#[derive(thiserror::Error, Debug)]
+#[error("The variant `AnyTypeEnum::{}` cannot be used with `{}()`!", self.variant, self.function)]
+pub struct InvalidVariant {
+    // TODO: Shorten lifetime?
```
EpicVon2468: (Wasn't sure how else to display the variant and function name; If there's a better way let me know)

EpicVon2468:
Another follow-up: should any other methods have wrappers implemented? should other `*TypeEnum` types have similar or identical methods implemented?

src/types/enums.rs
ErisianArchitect: You should add a lifetime generic, that's more flexible.

EpicVon2468: I'll get on this once I'm home

ErisianArchitect: Oh, wait, what is this? Is this an error message? Why is it implemented with strings? This could very well be an enum.

ErisianArchitect: Also, this error type would clash with inkwell's internal error types. It would make working with errors unwieldy if users have to switch between incompatible error types.

EpicVon2468: Which internal error types would be referred to here?

EpicVon2468: > Oh, wait, what is this? Is this an error message? Why is it implemented with strings? This could very well be an enum.

Yeah admittedly I was probably a bit too tired for implementing this yesterday.  I'll try to fix it up now

EpicVon2468: Now uses enumeration variant, but if there's a different way to implement the error type then lmk

src/types/enums.rs
```diff
@@ -759,3 +759,123 @@ impl Display for BasicMetadataTypeEnum<'_> {
         write!(f, "{}", self.print_to_string())
     }
 }
+
+#[derive(thiserror::Error, Debug)]
+#[error("This variant of `AnyTypeEnum` cannot be used with `{self:?}()`!")]
+pub enum InvalidVariant {
```
TheDan64: I think each variant should have the AnyTypeEnum inside like a tuple. And move the error macro to individual variants. So you'd have something like:

```rust
#[error("The fn_type method cannot be called by the {0} variant")]
FnType(AnyTypeEnum<'ctx>),
```

which would produce a really nice and explanatory error message

src/types/enums.rs
ErisianArchitect: The error type I was referring to is in src/error.rs

src/types/enums.rs
EpicVon2468: Sounds good to me, I'll implement when I'm free

src/types/enums.rs
ErisianArchitect: Sorry, I typed the last comment on my phone and didn't feel like being detailed.

It would be a good idea to add your  new error type to the main error in `src/error.rs`.

https://github.com/TheDan64/inkwell/blob/39f778fc393ee6d31d595ae1bb1c524b9d799e57/src/error.rs#L21

You can add a new error variant to it that implements `#[from]` for your error type. This will enable users to use the `?` operator on your error type and it will be recognized by inkwell's top-level error type.

EpicVon2468:
Running into some slight issues around the lifetime of the error
Seems that `Error` requires `'static`, meaning this is seemingly the only way to meet the criterion of storing `AnyTypeEnum<'ctx>` as well as integrating with `src/error.rs`...
image

EpicVon2468:
Oopsy I broke it
Forgot to ran tests before committing so I found out rust catches that leak only after pushing

EpicVon2468:
Actually, forget regular compile errors, I'm getting ICEs when I try to run tests 😭
Probably user error on my end, would explain why no errors showed up in my IDE until compile-time

src/types/enums.rs
EpicVon2468: As per the other comments I have below: this is causing issues related to lifetimes; need to find the best way to fix

TheDan64:
If you're getting an ICE on stable, you should definitely open a rust bug is one doesn't already exist. Sorry this seemingly small change has been so painful lol

EpicVon2468:
Haha it's fine.  I run nightly for everything (My projects use nightly features, so it's easier to make it the default than to try and fix some weird config issues I was having) & the bug was just about incremental cache, `cargo clean` fixed it.

~yes I know running nightly for everything is a horrible idea, it's a price I'm willing to pay~

Although _I'm_ sorry if this has been a pain, normally ~I'd like to think~ I'm good at programming, but I tend to over-doubt everything I do for contributions to projects :p

A more on-topic note:
While it's technically possible to just add a lifetime to `inkwell::error::Error` to (possibly?) avoid the `'static` problem, I'm not sure how that would go compatibility wise, and would rather not be at fault for however many people updating only to find their code is broken because of lifetimes on the error type 😅

ErisianArchitect:
Hmm. That's a tough one. What I think is that it's not necessary to include the original value within the error variant. I don't think there's any precedent for doing that. Otherwise, perhaps we could just exclude it from the crate-level error type and just expect users to handle the error themselves. I don't know if this is a case where someone would want to retry something using the `AnyTypeEnum`. I would think that it would be a bug in your code, and perhaps we should nudge the programmer into handling the bug properly rather than doing something weird. Perhaps we could give a method that would tell you whether or not this transformation can even happen before attempting it. Just some thoughts. You're doing fine, by the way. Programming is tricky sometimes.

EpicVon2468:
Omitting the original value would fix lifetime issues, but it also means people would have to do more tracing of which value was being passed on their end.

I personally would've opted for using `Option` on these methods out of personal preference, but if we're doing `Result` then it's more a matter of choosing _how_ the programmer would have to handle the result.

ErisianArchitect:
Can you really not make a fn_type for a function type?

TheDan64:
Another option would be to add a variant_name method to the type, which returns a static str, and then just do:

```rust
#[derive(Error)]
#[error("An invalid method was called by variant {0}")]
struct InvalidVariant(&'static str);

impl From<AnyTypeEnum<'_>> for InvalidVariant {
    fn from(ate) -> Self {
        InvalidVariant(ate.variant_name())
    }
}
```

This nice thing about this approach is that we could eventually use it for other types like BasicTypeEnum as well. Although this lacks method info..

ErisianArchitect:
I think it would be better to just have a `*Kind` enum that specifies the kind rather than relying on strings that will ultimately take up unnecessary space in the binary.

EpicVon2468:
> Can you really not make a fn_type for a function type?

Counter-intuitively, a function type in LLVM does not appear to be a "type" similar to the rest (to my understanding).

From what I know, it _isn't_ a closure or lambda or function pointer type...

EpicVon2468:
> Another option would be to add a variant_name method to the type, which returns a static str, and then just do:
> 
> ```rust
> #[derive(Error)]
> #[error("An invalid method was called by variant {0}")]
> struct InvalidVariant(&'static str);
> 
> impl From<AnyTypeEnum<'_>> for InvalidVariant {
>     fn from(ate) -> Self {
>         InvalidVariant(ate.variant_name())
>     }
> }
> ```
> 
> This nice thing about this approach is that we could eventually use it for other types like BasicTypeEnum as well. Although this lacks method info..

~We've gone full circle to what I had before 😭~

ErisianArchitect:
I just don't think that a string is the right call. If there's only a limited number of values that it could be, it would be better off as an enum.

EpicVon2468:
I won't take sides on which would be the better implementation, but if we do use enums I'd suggest marking them as `#[non_exhaustive]` so the API can be updated without breaking compatibility

EpicVon2468:
Yeah the lack of `'static` is causing issues with `anyhow` as well.
image
image
Because `InvalidVariantError` is not `'static`, it doesn't like it.

ErisianArchitect:
You almost gave me a heart attack, I thought anyhow had become a dependency in inkwell. Lol

EpicVon2468:
Hahahah no, I use it in my project (and I know other people like to use it (I think?)), and given the fact it (and most other `Error` related things (including `std` / `core`)) depend on the `'static` bound, this definitely needs a fix

ErisianArchitect:
Not that there's anything wrong with `anyhow`, it's just that we already have `thiserror` as a dependency, so if `anyhow` were added, it would just signal that someone had given up on writing decent errors and opted to basically just pass the trouble onto users. I'm glad that's not the case.

EpicVon2468:
Sorry to have made this and then left it alone a while, it's currently midyear exam time for me.

In just over a week or two I should be free to try and patch this up fully, although I could definitely work on fixes for the error type when I have some gaps in my schedule

ErisianArchitect:
I wouldn't worry about it, Dan takes his time.

PR #701, opened 2026-08-06
fix: update minimal supported version (12) everywhere

marxin:
Follow-up of d0a07a79625f5ac705e7daab4d9c8eb81016300e.


TheDan64 (approved):
Good catch, thank you!

PR #711, opened 2026-08-21
fix(LLVM 23): explicitly truncate values in const_int

### LINKED ISSUE #709: Change in the implicit truncation of `ConstantInt` in LLVM 23.1
After the following change (https://github.com/llvm/llvm-project/pull/171456), LLVM does not truncate contants and so:
```
    let i32_type = context.i32_type();

    let min_from_sign_extended_u64 = i32_type.const_int(i32::MIN as u64, false);
    let min_from_zero_extended_u64 = i32_type.const_int((i32::MIN as u32) as u64, false);

    assert_eq!(min_from_sign_extended_u64, min_from_zero_extended_u64);
```

will start failing:
```
    thread 'test_values::test_const_int_truncates_to_type_width' (3933854) panicked at tests/all/test_values.rs:931:5:
    assertion `left == right` failed
      left: IntValue { int_value: Value { name: "", address: 0x7ff98c003ed0, is_const: true, is_null: false, is_undef: false, llvm_value: "i32 -2147483648", llvm_type: "i32" } }
     right: IntValue { int_value: Value { name: "", address: 0x7ff98c003f00, is_const: true, is_null: false, is_undef: false, llvm_value: "i32 -2147483648", llvm_type: "i32" } }
```

We can theoretically report an error in `const_int` if the provided bit pattern `value: u64` contains a bit that is unrepresentative in the desired type.

Thoughts on that?

marxin:
Fixes: #709

### DIFF
--- a/src/types/int_type.rs
+++ b/src/types/int_type.rs
@@ -86,6 +86,7 @@ impl<'ctx> IntType<'ctx> {
     }
 
     /// Creates an `IntValue` representing a constant value of this `IntType`. It will be automatically assigned this `IntType`'s `Context`.
+    /// Note the bits in `value` beyond the width of this integer type are discarded.
     ///
     /// # Example
     /// ```no_run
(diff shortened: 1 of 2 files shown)


ErisianArchitect:
I do think it would be better if we used marker types instead. But I suppose that could be a PR later on. I don't mind implementing that, I'm just busy working on Orc2 at the moment.

TheDan64:
I also like marker types in theory, but people tend to want to store values in structs sort of generically (this is any int), and marker types make that harder. I suppose you could have a marker type for "any" but that's really outside the scope of this PR. There's also the question of custom integer types and how they'd be handled

TheDan64 (approved):

PR #719, opened 2026-09-16
fix: check the builder position in build_int_nsw_add

NotAFlightRisk:
## Description

The docs on `Builder` say every `build_*` method can return `UnsetPosition` if you haven't positioned it yet. `build_int_nsw_add` never checks, so on a fresh builder with two function params:

```rust
builder.build_int_add(a, b, "add")         // Err(UnsetPosition)
builder.build_int_nsw_add(a, b, "nsw_add") // Ok(%nsw_add = add nsw i32 %0, %1)
builder.build_int_nuw_add(a, b, "nuw_add") // Err(UnsetPosition)
```

That instruction has no parent block and isnt in the module, so it just goes missing with no error. Give it two constants and you get a folded `i32 2` back instead.

#436 added the check to all the builder methods and this one looks to have just been missed. It's the same three lines `build_int_nuw_add` has right below it.

## Related Issue

There isn't one, it seemed small enough to go straight to a PR.

## How This Has Been Tested

Added a test for an unpositioned `build_int_nsw_add` - fails on master, passes with the fix. Only on LLVM 21.1 though, linked by hand against the system `libLLVM.so` as I've no dev install, and in a seperate test file because the main `all` binary wouldn't link that way. `cargo clippy --tests -- -D warnings` and `cargo fmt --check` are clean.

## Checklist


TheDan64 (approved):
Good catch, thanks!

